PR #97, opened 2020-07-22
Create the structured data capture library

jingtang10:
This module includes

- QuestionnaireFragment that displays the Questionnaire and collects the QuestionnaireResponse from the user
- QuestionnaireViewModel that works as the fragment's data backend

The QuestionnaireFragment class acts as the public API


fhirengine/src/main/java/com/google/fhirengine/ui/QuestionnaireFragment.kt
```diff
@@ -0,0 +1,115 @@
...
+                    viewModel.map[linkId]?.answer = listOf(answer)
+                }
+            }
+            Questionnaire.QuestionnaireItemType.STRING -> {
+                val questionText = TextView(context)
+                questionText.text = questionnaireItemComponent.text
+                viewGroup.addView(questionText)
+                val input = EditText(context)
+                viewGroup.addView(input)
+                input.addTextChangedListener(object : TextWatcher{
```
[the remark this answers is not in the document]
jingtang10: I ended up using doAfterTextChanged.

jingtang10 (commented):
thanks florina!

i've decided to package this into a new module -- which can eventually be shipped as a separate library.

datacapture/build.gradle
```diff
@@ -0,0 +1,58 @@
...
+android {
+    compileSdkVersion 29
+    buildToolsVersion "30.0.2"
+
+    defaultConfig {
+        minSdkVersion 21
+        targetSdkVersion 29
+        versionCode 1
+        versionName "1.0"
+
```
yigit, line 15: not for this CL but eventually we should move these to a common configuration file.

datacapture/build.gradle
```diff
@@ -0,0 +1,58 @@
...
+}
+
+dependencies {
+    api ("ca.uhn.hapi.fhir:hapi-fhir-structures-r4:$hapiFhirStructuresR4Version") {
+        exclude module:'junit'
+    }
+    implementation 'androidx.lifecycle:lifecycle-livedata-ktx:2.2.0'
+    implementation 'androidx.lifecycle:lifecycle-viewmodel-ktx:2.2.0'
+    implementation "org.jetbrains.kotlin:kotlin-stdlib:$kotlinVersion"
+    implementation 'androidx.core:core-ktx:1.2.0'
```
yigit, line 48: same for dependencies here (again, a followup is fine)

datacapture/src/main/AndroidManifest.xml
```diff
@@ -0,0 +1,5 @@
+<?xml version="1.0" encoding="utf-8"?>
```
yigit, line 1: copyright

datacapture/src/main/java/com/google/android/fhir/datacapture/QuestionnaireFragment.kt
```diff
@@ -0,0 +1,122 @@
...
+import android.widget.LinearLayout
+import android.widget.TextView
+import androidx.core.widget.doAfterTextChanged
+import androidx.fragment.app.Fragment
+import androidx.lifecycle.ViewModelProvider
+import java.util.Locale
+import org.hl7.fhir.r4.model.Questionnaire
+import org.hl7.fhir.r4.model.QuestionnaireResponse
+
+class QuestionnaireFragment(private val questionnaire: Questionnaire) : Fragment() {
```
yigit, line 36: this will be a problem in restoring state but maybe you want to tackle it later?

datacapture/src/main/java/com/google/android/fhir/datacapture/QuestionnaireFragment.kt
```diff
@@ -0,0 +1,122 @@
...
+    private lateinit var viewModel: QuestionnaireViewModel
+
+    internal var onQuestionnaireSubmittedListener: OnQuestionnaireSubmittedListener? = null
+
+    override fun onCreate(savedInstanceState: Bundle?) {
+        super.onCreate(savedInstanceState)
+        viewModel = ViewModelProvider(this, QuestionnaireViewModelFactory(questionnaire))
```
yigit, line 43: you can use `by viewmodels` and just make it a field.

https://developer.android.com/topic/libraries/architecture/viewmodel

datacapture/src/main/java/com/google/android/fhir/datacapture/QuestionnaireFragment.kt
```diff
@@ -0,0 +1,122 @@
...
+      inflater: LayoutInflater,
+      container: ViewGroup?,
+      savedInstanceState: Bundle?
+    ): View? {
+        val view = inflater.inflate(R.layout.questionnaire_fragment, container, false)
+        view.findViewById<TextView>(R.id.title).text = viewModel.questionnaire.title
+        val item = view.findViewById<LinearLayout>(R.id.item)
+
+        viewModel.questionnaire.item.forEach {
+            generateAndAttachQuestion(it, item)
```
yigit, line 57: we should use a recyclerview here as the questionnaire might get long.

datacapture/src/main/java/com/google/android/fhir/datacapture/QuestionnaireFragment.kt
```diff
@@ -0,0 +1,122 @@
...
+            onQuestionnaireSubmittedListener?.onSubmitted(viewModel.questionnaireResponse)
+        }
+        return view
+    }
+
+    private fun generateAndAttachQuestion(
+      questionnaireItemComponent: Questionnaire.QuestionnaireItemComponent,
+      viewGroup: ViewGroup
+    ) {
+        Locale.getDefault().language
```
yigit, line 70: ?

datacapture/src/main/java/com/google/android/fhir/datacapture/QuestionnaireFragment.kt
```diff
@@ -0,0 +1,122 @@
...
+            Questionnaire.QuestionnaireItemType.GROUP -> {
+                val linearLayout = LinearLayout(context)
+                linearLayout.layoutParams =
+                    LinearLayout.LayoutParams(LinearLayout.LayoutParams.MATCH_PARENT,
+                        LinearLayout.LayoutParams.WRAP_CONTENT)
+                linearLayout.orientation = LinearLayout.VERTICAL
+                val title = TextView(context)
+                title.text = questionnaireItemComponent.text
+                linearLayout.addView(title)
+                questionnaireItemComponent.item.forEach {
```
yigit, line 100: is probably best to flatten this list. (will play better w/ recyclerview as well)

datacapture/src/main/res/layout/questionnaire_fragment.xml
```diff
@@ -0,0 +1,25 @@
+<?xml version="1.0" encoding="utf-8"?>
```
yigit, line 1: copyright

datacapture/src/test/java/com/google/android/fhir/datacapture/QuestionnaireViewModelTest.kt
```diff
@@ -0,0 +1,142 @@
...
+import org.hl7.fhir.r4.model.QuestionnaireResponse
+import org.hl7.fhir.r4.model.Resource
+import org.hl7.fhir.r4.model.StringType
+import org.junit.Test
+import org.junit.runner.RunWith
+import org.robolectric.RobolectricTestRunner
+import org.robolectric.annotation.Config
+
+/** Unit tests for {@link FhirIndexerImpl}. */
+@RunWith(RobolectricTestRunner::class)
```
yigit, line 33: i would recommend using integration tests here.
you can use fragment scenerio

example/src/main/java/com/google/fhirengine/example/QuestionnaireActivity.kt
```diff
@@ -0,0 +1,129 @@
...
+                  ]
+                }
+              ]
+            }
+        """.trimIndent())
+
+        // modifications to the questionnaire
+        questionnaire.title = "My questionnaire"
+
+        val fragment = QuestionnaireFragment(questionnaire)
```
yigit, line 116: we should use the result API here.

https://developer.android.com/reference/kotlin/androidx/fragment/app/FragmentResultListener

datacapture/build.gradle
jingtang10: created https://github.com/google/android-fhir/pull/122 -- please take a look at that one too!

datacapture/src/main/java/com/google/android/fhir/datacapture/QuestionnaireFragment.kt
jingtang10: yep let's handle that in a separate PR.

datacapture/src/main/java/com/google/android/fhir/datacapture/QuestionnaireFragment.kt
jingtang10: done. thx!

fhirengine/build.gradle
```diff
@@ -71,6 +71,9 @@ dependencies {
     implementation "com.google.guava:guava:$guavaVersion"
     implementation "org.jetbrains.kotlin:kotlin-stdlib-jdk7:$kotlinVersion"
     implementation "org.opencds.cqf:cql-engine-fhir:$cqlEngineVersion"
+    implementation 'androidx.legacy:legacy-support-v4:1.0.0'
```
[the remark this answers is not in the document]
yigit: no we shouldn't need live data here, worst case we have flow

jingtang10 (commented):
thanks yigit for the review - ptal!

regarding integration tests vs unit tests, i actually think that test file itself is best kept as unit tests. it does not have any dependencies on UI or android components, it's purely about the logic to create a questionnaire response from a questionnaire and keep the same structure.

i will add more integration tests.

jingtang10:
@yigit sorry this PR has bloated quite a bit. If possible at all I'd love to be able to merge this asap as this will unblock future work as well for @pankajkatoch. Just wanted to make sure the core structure is good. thanks again.

datacapture/src/test/java/com/google/android/fhir/datacapture/QuestionnaireViewModelTest.kt
jingtang10: replied in the comment above. i think the test cases here have no dependency on fragment so keeping in unit tests is fine?

datacapture/src/main/java/com/google/android/fhir/datacapture/QuestionnaireViewModel.kt
```diff
@@ -0,0 +1,116 @@
...
+
+import androidx.lifecycle.ViewModel
+import androidx.lifecycle.ViewModelProvider
+import org.hl7.fhir.r4.model.BooleanType
+import org.hl7.fhir.r4.model.DateType
+import org.hl7.fhir.r4.model.Questionnaire
+import org.hl7.fhir.r4.model.QuestionnaireResponse
+import org.hl7.fhir.r4.model.StringType
+
+class QuestionnaireViewModel(val questionnaire: Questionnaire) : ViewModel() {
```
yigit, line 27: there is some encapsulation issue here.

right now, you are giving the view model to the adapter (which we shouldn't).

instead, ideally, view model could provide an immutable list that is observed and each time data changes, it changes which will notify the adapter to update itself.

now it will be obviously less efficient but if we ever end up in a situation where the questionnaire might change etc, it will help. we are also responsible to call notifyChange on RV when items are changed which we are not doing right now. It might cause RV to cache invalid data if data changes for reasons outside the UI (not sure if we have that case).

Given that this PR is already huge, i'm OK if we do the cleanup in a followup.

datacapture/src/main/java/com/google/android/fhir/datacapture/QuestionnaireItemAdapter.kt
```diff
@@ -0,0 +1,80 @@
...
+import com.google.android.fhir.datacapture.views.QuestionnaireItemCheckBoxViewHolderFactory
+import com.google.android.fhir.datacapture.views.QuestionnaireItemDatePickerViewHolderFactory
+import com.google.android.fhir.datacapture.views.QuestionnaireItemEditTextViewHolderFactory
+import com.google.android.fhir.datacapture.views.QuestionnaireItemGroupViewHolderFactory
+import com.google.android.fhir.datacapture.views.QuestionnaireItemViewHolder
+import org.hl7.fhir.r4.model.Questionnaire
+import org.hl7.fhir.r4.model.Questionnaire.QuestionnaireItemComponent
+
+class QuestionnaireItemAdapter(
+  val viewModel: QuestionnaireViewModel
```
yigit, line 31: instead of passing the view model, it is better to just receive a callback for modifications for better de-coupling.

datacapture/src/main/java/com/google/android/fhir/datacapture/views/QuestionnaireItemCheckBoxViewHolderFactory.kt
```diff
@@ -0,0 +1,49 @@
...
+private class QuestionnaireItemCheckBoxViewHolder(
+  itemView: View,
+  viewModel: QuestionnaireViewModel
+) :
+    QuestionnaireItemViewHolder(itemView, viewModel) {
+    private val checkBox = itemView.findViewById<CheckBox>(R.id.check_box)
+
+    override fun bind(questionnaireItemComponent: Questionnaire.QuestionnaireItemComponent) {
+        checkBox.text = questionnaireItemComponent.text
+        checkBox.setOnClickListener {
```
yigit, line 45: for all of these, they should add the listener when created and in on bind, simply retrieve the item (can be recorded when bind is called) and then call the callback.

datacapture/src/main/java/com/google/android/fhir/datacapture/views/QuestionnaireItemDatePickerViewHolderFactory.kt
```diff
@@ -0,0 +1,87 @@
...
+    private val input = itemView.findViewById<TextView>(R.id.input)
+    private val button = itemView.findViewById<TextView>(R.id.button)
+
+    override fun bind(questionnaireItemComponent: Questionnaire.QuestionnaireItemComponent) {
+        textView.text = questionnaireItemComponent.text
+        button.setOnClickListener {
+            // TODO: find a more robust way to do this.
+            val context = itemView.context as AppCompatActivity
+            DatePickerFragment().show(context.supportFragmentManager, DatePickerFragment.TAG)
+            context.supportFragmentManager.setFragmentResultListener(
```
yigit, line 56: we should take this out of the view model.
view model might disappear for various reasons.
ideally, such changes should be reflected just on the view model which can notify the change via submit list.

this CL is monster so might be a followup.

datacapture/src/test/java/com/google/android/fhir/datacapture/QuestionnaireItemAdapterTest.kt
```diff
@@ -0,0 +1,79 @@
...
+    @Test
+    fun getItemViewType_stringItemType_shouldReturnEditTextViewHolderType() {
+        questionnaireItemAdapter.submitList(listOf(Questionnaire.QuestionnaireItemComponent()
+            .apply { type = Questionnaire.QuestionnaireItemType.STRING }))
+        assertThat(questionnaireItemAdapter.getItemViewType(0)).isEqualTo(
+            QuestionnaireItemViewHolderType.EDIT_TEXT.value
+        )
+    }
+
+    //    TODO: add this test once upgraded to JUnit5
```
yigit, line 70: i would say just remove if it will stay commented.

yigit (commented):
honestly i just skimmed through the PR because it is way too big to review properly :/

would be good to divide it up (e.g. merge skeleton, then merge type by type etc).

but also, i don't want to block the progress right before holidays so feel free to merge.

fhirengine/build.gradle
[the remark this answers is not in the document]
jingtang10: this is removed

jingtang10:
> honestly i just skimmed through the PR because it is way too big to review properly :/
> 
> would be good to divide it up (e.g. merge skeleton, then merge type by type etc).
> 
> but also, i don't want to block the progress right before holidays so feel free to merge.

Thanks Yigit for the review I have separated the gallery app into a different branch now... Will do the same for the individual types.

With particularly big changes, would it make sense to merge them (with review) into a feature branch and then send a final review from feature branch to master branch? I know we have missed that boat for this PR now.

datacapture/src/main/java/com/google/android/fhir/datacapture/QuestionnaireFragment.kt
```diff
@@ -0,0 +1,80 @@
...
+        adapter.submitList(flatten(viewModel.questionnaire.item))
+
+        view.findViewById<Button>(R.id.submit).setOnClickListener {
+            val serializedResponse = FhirContext.forR4().newJsonParser()
+                .encodeResourceToString(viewModel.questionnaireResponse)
+            setFragmentResult(
+                QUESTIONNAIRE_RESPONSE_REQUEST_KEY,
+                bundleOf(QUESTIONNAIRE_RESPONSE_BUNDLE_KEY to serializedResponse)
+            )
+        }
```
kashyapj2793, line 58: > It is recommended to only inflate the layout in this method and move logic that operates on the returned View to onViewCreated(View, Bundle). 

We should only inflate the layout in onCreateView and move the rest of the code to onViewCreated
[Reference](https://developer.android.com/reference/androidx/fragment/app/Fragment#onCreateView(android.view.LayoutInflater,%20android.view.ViewGroup,%20android.os.Bundle))

datacapture/src/main/java/com/google/android/fhir/datacapture/QuestionnaireItemViewHolderType.kt
```diff
@@ -0,0 +1,37 @@
...
+ * rendered.
+ */
+enum class QuestionnaireItemViewHolderType(val value: Int) {
+    GROUP(1),
+    CHECK_BOX(2),
+    DATE_PICKER(3),
+    EDIT_TEXT(4);
+
+    companion object {
+        private val VALUES = values()
```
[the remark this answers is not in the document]
jingtang10: why not cache this so we don't need to recompute it every time?

ok I've changed it to use the index. I think it can be safe with QuestionnaireItemViewHolderTypeTest#fromInt_shouldReturnQuestionnaireViewHolderType.

datacapture/src/main/java/com/google/android/fhir/datacapture/QuestionnaireItemAdapter.kt
jingtang10: it can't just be a callback because the item adapter doesn't actually know the type of answer to send back. i created an interface called QuestionnaireResponseReceiver. Is that ok?

jingtang10 (commented):
Thanks @yigit and Kashyap for your reviews. I think I addressed all comments now.

@yigit and I discussed the use of recycler view yesterday.

- we will not use list adapter any more. we will use recycler view adapter instead. this is mainly because we don't really need it...
- the view model will produce a list which is passed to the adapter. the adapter will no longer hold reference to the view model (which is incorrect)

Please take another look -- would love to merge this asap if poss.

deepankarb (approved):
Left a couple of nits. Approved!

datacapture/src/main/java/com/google/android/fhir/datacapture/views/DatePickerDialogFragment.kt
```diff
@@ -0,0 +1,59 @@
...
+package com.google.android.fhir.datacapture.views
+
+import android.app.DatePickerDialog
+import android.app.Dialog
+import android.os.Bundle
+import android.widget.DatePicker
+import androidx.core.os.bundleOf
+import androidx.fragment.app.DialogFragment
+import androidx.fragment.app.setFragmentResult
+import java.util.Calendar
```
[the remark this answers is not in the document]
jingtang10: thanks done. had to add suppress warning though - but is ok as long as user of library uses desugaring.

datacapture/proguard-rules.pro
```diff
@@ -0,0 +1,21 @@
+# Add project specific ProGuard rules here.
```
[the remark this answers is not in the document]
jingtang10: done.

jingtang10 (commented):
thanks!

PR #105, opened 2020-08-18
Add APIs for local changes

deepankarb:
* Get all local updates
* Delete all local changes by resource (id, type)


deepankarb:
Implemented in #146

PR #113, opened 2020-11-26
Convert some Java classes to Kotlin (and add basic unit tests for resource utils)

jingtang10:


fhirengine/src/main/java/com/google/fhirengine/ResourceAlreadyExistsException.kt
```diff
@@ -14,24 +14,11 @@
...
-  public String getType() {
-    return type;
-  }
-
-  public String getId() {
-    return id;
-  }
-}
+class ResourceAlreadyExistsException(
+  type: String,
```
yigit, line 21: these need to be `val`'s. (except for cause)

fhirengine/src/main/java/com/google/fhirengine/ResourceNotFoundException.kt
```diff
@@ -14,24 +14,11 @@
...
-  public String getType() {
-    return type;
-  }
-
-  public String getId() {
-    return id;
-  }
-}
+class ResourceNotFoundException(
+  type: String,
```
yigit, line 21: ditto

fhirengine/src/main/java/com/google/fhirengine/resource/Resources.kt
```diff
@@ -0,0 +1,50 @@
...
+ */
+
+package com.google.fhirengine.resource
+
+import java.lang.reflect.InvocationTargetException
+import org.hl7.fhir.r4.model.Resource
+import org.hl7.fhir.r4.model.ResourceType
+
+/** The HAPI Fhir package prefix for R4 resources. */
+val R4_RESOURCE_PACKAGE_PREFIX = "org.hl7.fhir.r4.model."
```
yigit, line 24: i see the previous java class was public but i think these are all internal functionality right?
if so, we should mark these `internal` (everything in this file)

fhirengine/src/main/java/com/google/fhirengine/resource/Resources.kt
```diff
@@ -0,0 +1,50 @@
...
+        throw IllegalArgumentException("Cannot resolve resource type for " + clazz.getName(), e)
+    } catch (e: InstantiationException) {
+        throw IllegalArgumentException("Cannot resolve resource type for " + clazz.getName(), e)
+    } catch (e: InvocationTargetException) {
+        throw IllegalArgumentException("Cannot resolve resource type for " + clazz.getName(), e)
+    }
+}
+
+/** Returns the {@link Class} object for the resource type. */
+fun <R : Resource> getResourceClass(resourceType: String): Class<R> {
```
yigit, line 42: we may want to move this to KClass instead but does not need to be in this PR

yigit (changes requested):

fhirengine/src/main/java/com/google/fhirengine/ResourceAlreadyExistsException.kt
jingtang10: why? we don't need them after the constructor right?

fhirengine/src/main/java/com/google/fhirengine/ResourceNotFoundException.kt
jingtang10: ditto

fhirengine/src/main/java/com/google/fhirengine/resource/Resources.kt
jingtang10: yeah thanks for the catch!

fhirengine/src/main/java/com/google/fhirengine/resource/Resources.kt
jingtang10: ack. i'll do that in a following pr.

jingtang10:
rebased. PTAL! thx! @yigit

fhirengine/src/androidTest/java/com/google/fhirengine/db/impl/DatabaseImplTest.kt
```diff
@@ -158,4 +158,22 @@ class DatabaseImplTest {
             TEST_PATIENT_2.setGender(Enumerations.AdministrativeGender.MALE)
         }
     }
+
+    private companion object {
```
deepankarb, line 162: Duplicate code? There's another `companion` right above.

fhirengine/src/androidTest/java/com/google/fhirengine/db/impl/DatabaseImplTest.java
```diff
@@ -0,0 +1,137 @@
+/*
```
deepankarb, line 1: This should be removed.

deepankarb (changes requested):
Empty file named 'Task' can be removed. Left a couple of comments. Thanks!

Also please take a look at spotless failure in CI: https://github.com/google/android-fhir/pull/113/checks?check_run_id=1487953151#step:10:27

fhirengine/src/main/java/com/google/fhirengine/ResourceAlreadyExistsException.kt
yigit: there are public getters/setters for it in the java version. i thought you wanted to keep them as properties ? with his, they cannot be read back anymore.

jingtang10: yeah you're right. if we don't expose these the class is a bit pointless.

yigit (approved):

PR #114, opened 2020-11-27
Convert DatabaseImplTest to Kotlin

jingtang10:
They do not need to be run as integration tests.


fhirengine/src/test/java/com/google/fhirengine/db/impl/DatabaseImplTest.kt
```diff
@@ -0,0 +1,152 @@
...
+            TEST_PATIENT_1_ID)
+        patient.setGender(Enumerations.AdministrativeGender.FEMALE)
+        database.update(patient)
+        testingUtils.assertResourceEquals(patient,
+            database.select(Patient::class.java,
+                TEST_PATIENT_1_ID))
+    }
+
+    @Test
+    fun update_existingResource_shouldInsertResource() {
```
deepankarb, line 94: Looks like we're updating a non-existent resource `TEST_PATIENT_2` here.
I think the method name is mixed up with `update_nonexistentResource_shouldUpdateResource()` above. They should be swapped.

fhirengine/src/test/java/com/google/fhirengine/db/impl/DatabaseImplTest.kt
```diff
@@ -0,0 +1,152 @@
...
+    private val services =
+        FhirServices.builder(dataSource, ApplicationProvider.getApplicationContext())
+            .inMemory()
+            .build()
+    private val testingUtils = TestingUtils(services.parser)
+    private val database = services.database
+
+    @Before
+    fun setUp() {
+        database.insert(
```
deepankarb, line 52: Methods wrapped/split weirdly throught the file.

deepankarb (approved):
LGTM, Approved, with a couple of nits. Thanks!

fhirengine/src/test/java/com/google/fhirengine/db/impl/DatabaseImplTest.kt
jingtang10: yeah fixed it thanks:)

jingtang10:
> LGTM, Approved, with a couple of nits. I think they should be run as integration tests since sqlite behaviour varies from device to device.

Thanks for the review 👍 

Interesting. What kind of behaviour differences will integration tests be able to check? Can you give me some examples or a link to some resources?

I worry about our integration tests in general - they are not doing much and are constantly failing.

But specifically to this test, I think the behaviours are fairly well defined and understood. Unless we're really catching some weird corner cases (in which case we need to make sure we have device coverage to actually test those differences), I think unit tests would be appropriate.

@yigit @pankajkatoch @florina-muntenescu who might have more input.

yigit:
our official recommendation is to run these as integration tests.
First part is that because different versions of android ship with different versions of sqlite, their capabilities might differ and we might miss those with robolectric tests.
Second, robolectric's sqlite implementation does not match android 1-1. For instance: https://github.com/robolectric/robolectric/blob/master/shadows/framework/src/main/java/org/robolectric/shadows/ShadowSQLiteConnection.java#L97

Hence, we should keep these as integration tests. I know launching an emulator is annoying but besides that, these tests should be very stable and fast since they don't have any UI components.

deepankarb:
Thanks for the details @yigit !

Apart from fragmentation across OS versions, there's also fragmentation across devices/form factors, e.g. see [SQLite syntax error but only on Samsung Galaxy Tab
](https://stackoverflow.com/questions/20905692/sqlite-syntax-error-but-only-on-samsung-galaxy-tab)

I'll fix the CI errors and target 100% green builds.

deepankarb (approved):

jingtang10:
Thanks @yigit and @deepankarb for explanation. It sounds like the robolectric tests wouldn't even be correct in the first place :( but do you think we should aim to add more devices to the integration tests over time for coverage of the inconsistency issue you talked about?

@deepankarb thanks if you can take a look at CI again that'd be great!

@yigit I've added some comments in the test file. Can you please take a look? Please comment especially if there's a link to the official recommendation you spoke of.

deepankarb (approved):

deepankarb:
Good to merge!

jingtang10:
thanks deepankar!

Issue #109, opened 2020-09-30
Target 1.8 for kotlin and enable androidx.lifecycle ktx module


pankajkatoch:
Use [`androidx.lifecycle.observe`](https://developer.android.com/reference/kotlin/androidx/lifecycle/package-summary#(androidx.lifecycle.LiveData).observe(androidx.lifecycle.LifecycleOwner,%20kotlin.Function1)) instead of java observe.

pankajkatoch:
LiveData.observe() has itself been deprecated. [Deprecate LiveData.observe Kotlin extension](https://android-review.googlesource.com/c/platform/frameworks/support/+/1413815)

With move to Kotlin 1.4, Kotlin itself is able to convert the default observe() to lambada syntax.

PR #117, opened 2020-12-01
Upgrade to Kotlin Version 1.4

pankajkatoch:
With move to Kotlin 1.4, Kotlin itself is able to convert the default observe() to lambada syntax.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -1,6 +1,6 @@
 // Top-level build file where you can add configuration options common to all sub-projects/modules.
 buildscript {
-    ext.kotlinVersion = '1.3.72'
+    ext.kotlinVersion = '1.4.20'
     ext.spotlessVersion = '3.27.1'
 
     repositories {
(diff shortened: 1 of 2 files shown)


jingtang10 (approved):

jingtang10:
Please create an issue and link here.

deepankarb (changes requested):
[CI](https://github.com/google/android-fhir/runs/1483118146#step:11:365) is failing since Kotlin 1.4 [throws](https://kotlinlang.org/docs/reference/whatsnew14.html#unified-exception-type-for-null-checks) `NullPointerException` instead of `IllegalStateException`. Could you fix the affected test case by changing it to expect a `NullPointerException`?

pankajkatoch:
This PR is for #118 which is dependency for #109 
Also updated the failing Indexer unit test to account for Unified exception type for null checks in Kotlin 1.4

Issue #73, opened 2020-07-01
Integration test out of memory for API 29


jingtang10:
In recent github action runs for PRs, tests for API 29 consistently failed due to out of memory issue. e.g:

com.google.fhirengine.db.impl.DatabaseImplTest > select_invalidResourceType_shouldThrowIllegalArgumentException[test(AVD) - 10] FAILED 
java.lang.OutOfMemoryError: OutOfMemoryError thrown while trying to throw an exception; no stack trace available

See trace: https://github.com/google/android-fhir/runs/824594488

Note this doesn't happen for API 24 or 27, nor does this happen locally.

deepankarb:
Still seeing OOM failures on some branches.

jingtang10:
A more recent run: https://github.com/google/android-fhir/runs/2054399134?check_suite_focus=true

This blocked code submission over the weekend. We should treat this as high-priority.

jingtang10:
Related issue: https://github.com/ReactiveCircus/android-emulator-runner/issues/131

jingtang10:
This has not been happening - let's close for now and reopen if recurs.

PR #119, opened 2020-12-03
API 29 CI OOM fix

deepankarb:
Attempting to fix `./gradlew connectedCheck` OOM errors in Github Actions by reducing gradle and gradle test JVM max heap. This should make more RAM available for emulator.

Gradle JVM heap reduced from 5 GB to 3 GB.
Gradle test JVM heap reduced from default to 256 MB.

### DIFF
--- a/.github/ci-gradle.properties
+++ b/.github/ci-gradle.properties
@@ -16,5 +16,5 @@
 
  org.gradle.daemon=false
  org.gradle.parallel=true
- org.gradle.jvmargs=-Xmx5120m
+ org.gradle.jvmargs=-Xmx3072m
  org.gradle.workers.max=2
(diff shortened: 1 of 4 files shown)


jingtang10:
Is there a place where this is documented? I imagine there might be a stack overflow post? If so can you link it? Thanks.

Btw i linked the issue here.

jingtang10:
Having thought about this -- can you separate the 3 different changes into separate PRs with CI runs? I'd love to understand exactly which one of these changes is fixing the issue!

jingtang10:
is this still needed?

deepankarb:
We're seeing fairly less `OutOfMemory` errors now. #73 has been better fixed by #193.

jingtang10:
We are seeing some more out of memory errors now. @deepankarb do you think any of these fixes will be useful?

deepankarb:
I think they're worth a try. I'll run each option + current master as control 10 times to quantify if they reduce oom errors.

PR #121, opened 2020-12-08
Convert FhirEngineImplTest to Kotlin

jingtang10:


fhirengine/src/test/java/com/google/fhirengine/impl/FhirEngineImplTest.kt
```diff
@@ -0,0 +1,142 @@
...
+    private val fhirEngine = services.fhirEngine
+    private val testingUtils = TestingUtils(services.parser)
+
+    @Before
+    fun setUp() {
+        fhirEngine.save(TEST_PATIENT_1)
+    }
+
+    @Test
+    @Throws(Exception::class)
```
yigit, line 56: not needed (for all, you don't need to report thrown exceptions)

yigit (approved):

PR #122, opened 2020-12-11
Create common gradle config to organize dependencies

jingtang10:
- Creates a common.gradle file in the project root directory that contains configurations applied to each module
- Bumps up desugaring library version from 1.0.4 to 1.0.9
- Removes java spotless setting which is not needed any more
- Move maven central repository dependency to project build.gradle file


common.gradle
```diff
@@ -0,0 +1,41 @@
...
+    kotlinOptions {
+        jvmTarget = JavaVersion.VERSION_1_8.toString()
+    }
+}
+
+dependencies {
+    implementation "org.jetbrains.kotlin:kotlin-stdlib-jdk7:$kotlinVersion"
+    implementation "org.jetbrains.kotlinx:kotlinx-coroutines-android:$coroutinesVersion"
+    implementation "org.jetbrains.kotlinx:kotlinx-coroutines-core:$coroutinesVersion"
+    coreLibraryDesugaring 'com.android.tools:desugar_jdk_libs:1.0.9'
```
deepankarb, line 40: Can the version here be moved to top-level `build.gradle` like the versions above?

deepankarb (approved):
Left 1 not. LGTM!

Thanks!

common.gradle
jingtang10: done

jingtang10:
@yigit commented:

it is better to just move versions / artifact coordinates to ext
then still add their depepdencies manually
so that you can see what a module depends on

here is an example:
https://github.com/android/architecture-components-samples/blob/main/PagingSample/build.gradle
https://github.com/android/architecture-components-samples/blob/main/PagingSample/versions.gradle

it has two things, the versions and maven repositories
then other module's build gradle files look like this:
https://github.com/android/architecture-components-samples/blob/main/PagingSample/app/build.gradle

deps.gradle
```diff
@@ -0,0 +1,93 @@
...
+versions.atsl_junit = '1.1.1'
+versions.constraint_layout = '1.1.3'
+versions.coroutines = '1.3.4'
+versions.cql_engine = '1.3.14-SNAPSHOT'
+versions.desugar = '1.0.9'
+versions.guava = '28.2-android'
+versions.hapi_r4 = '4.2.0'
+versions.junit = '4.13'
+versions.kotlin = '1.4.20'
+versions.legacy_support = '1.0.0'
```
yigit, line 26: where do we use this?
we shuld probably get rid of it if possible

yigit (approved):

deps.gradle
jingtang10: hm... i think it should have been caught by code review. with min sdk version 21 i don't think we need this.

thanks for catching!

PR #127, opened 2020-12-20
[DRAFT] prototype to provide custom view implementation

jingtang10:


jingtang10:
Closing because the sdc APIs have moved on quite a bit. But if/when we pick this one up again, the general principle will be the same: the developer is able to provide a custom implementation to the sdc lib so that the sdc lib will use the custom view to render a particular item in the questionnaire and add it to the recycler view.

Issue #132, opened 2020-12-28
datacapture : Bug in binding UI element to Questionnaire Item


kashyapj2793:
https://user-images.githubusercontent.com/48797364/103201704-8d14f080-4916-11eb-8338-1e1cf7cc68db.mp4

Sample questionnaire modified by duplicating items. linkid of each item is unique

jingtang10:
Thanks Kashyap for creating this issue. Can you please push a branch with your code?

I think there's a real problem with the use of activityViewModels in QuestionnaireFragment.kt. I tested that passing different questionnaires to different questionnaire fragments will reuse the same view model which is definitely incorrect. We need to somehow change the scope of view models to fragment...

kashyapj2793:
@jingtang10 Please check this branch [https://github.com/google/android-fhir/tree/kj/sdc-binding-bug](url)

jingtang10:
@kashyapj2793, I created PR #134 and tested with your change which I think is working. To verify that on your end, please rebase your branch onto the jt/sdc-set-values branch.

The problem you encountered is very interesting. It's because recycler view reuses (and it's great to see that in action because it improves performance) the questionnaire item views. So if you have ticked a checkbox and scroll it outside of the viewport, recycler view might reuse that checkbox to render another question if that question is to be scrolled into the viewport. Because we had not implemented loading the answers properly from the view model, so when that happened, the adapter didn't set the state of the checkbox correctly. So what you ended up with was a checkbox that had been ticked for a question being reused for another question without being cleared.

kashyapj2793:
I have verified that PR #134 fixes this issue

jingtang10:
Thanks Kashyap -- let's keep this open until that PR is merged into master.

kashyapj2793:
I have reset the tip of this branch to the first commit without the fix from the jt/sdc-set-values branch. This is so that until the PR is merged to master, the issue can be reproduced using this branch.

jingtang10:
Sorry I didn't mean to close this.

Kashyap -- can you please verify this is indeed fixed for you (rebasing your branch onto master).

kashyapj2793:
No problem jing! Yes I have verified that this is fixed!

PR #133, opened 2020-12-28
Change the scope of questionnaire fragment's view model from activity to fragment

jingtang10:


kashyapj2793:
This seems like the right way to use ViewModel

> Caution: A ViewModel must never reference a view, Lifecycle, or any class that may hold a reference to the activity context.

[Reference](https://developer.android.com/topic/libraries/architecture/viewmodel#implement)

jingtang10:
upon closer inspection i think this is not linked with #132. (but i've no idea now to disassociate that issue on github...)

yigit (approved):

PR #134, opened 2020-12-28
Update the UI using the answers in the view model

jingtang10:
We had not implemented this because we assumed that it's ok to render the UI without providing an initial value until we implement the initial value to the SDC specification.

However, this already caused issue because the recycler view reuses the questionnaire item views, and if we don't try to set the answer the view might still carry what the user had input for a different question.

### DIFF
--- a/datacapture/src/main/java/com/google/android/fhir/datacapture/views/QuestionnaireItemCheckBoxViewHolderFactory.kt
+++ b/datacapture/src/main/java/com/google/android/fhir/datacapture/views/QuestionnaireItemCheckBoxViewHolderFactory.kt
@@ -33,24 +33,18 @@ object QuestionnaireItemCheckBoxViewHolderFactory : QuestionnaireItemViewHolderF
       override fun init(itemView: View) {
         checkBox = itemView.findViewById(R.id.check_box)
         checkBox.setOnClickListener {
-          questionnaireItemViewItem.questionnaireResponseItemComponent.answer = listOf(
+          questionnaireItemViewItem.singleAnswerOrNull =
             QuestionnaireResponse.QuestionnaireResponseItemAnswerComponent().apply {
               value = BooleanType(checkBox.isChecked)
             }
-          )
         }
       }
 
       override fun bind(questionnaireItemViewItem: QuestionnaireItemViewItem) {
         this.questionnaireItemViewItem = questionnaireItemViewItem
         checkBox.text = questionnaireItemViewItem.questionnaireItemComponent.text
-        questionnaireItemViewItem.questionnaireResponseItemComponent.answer.also {
-          if (it.size == 1 && it[0].hasValueBooleanType()) {
-            checkBox.isChecked = it[0].valueBooleanType.value
-          } else {
-            checkBox.isChecked = false
-          }
-        }
+        checkBox.isChecked =
+          questionnaireItemViewItem.singleAnswerOrNull?.valueBooleanType?.booleanValue() ?: false
       }
     }
 }
--- a/datacapture/src/main/java/com/google/android/fhir/datacapture/views/QuestionnaireItemCheckBoxViewHolderFactory.kt
+++ b/datacapture/src/main/java/com/google/android/fhir/datacapture/views/QuestionnaireItemCheckBoxViewHolderFactory.kt
@@ -44,6 +44,13 @@ object QuestionnaireItemCheckBoxViewHolderFactory : QuestionnaireItemViewHolderF
       override fun bind(questionnaireItemViewItem: QuestionnaireItemViewItem) {
         this.questionnaireItemViewItem = questionnaireItemViewItem
         checkBox.text = questionnaireItemViewItem.questionnaireItemComponent.text
+        questionnaireItemViewItem.questionnaireResponseItemComponent.answer.also {
+          if (it.size == 1 && it[0].hasValueBooleanType()) {
+            checkBox.isChecked = it[0].valueBooleanType.value
+          } else {
+            checkBox.isChecked = false
+          }
+        }
       }
     }
 }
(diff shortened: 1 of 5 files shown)


datacapture/src/main/java/com/google/android/fhir/datacapture/views/QuestionnaireItemCheckBoxViewHolderFactory.kt
```diff
@@ -44,6 +44,13 @@ object QuestionnaireItemCheckBoxViewHolderFactory : QuestionnaireItemViewHolderF
       override fun bind(questionnaireItemViewItem: QuestionnaireItemViewItem) {
         this.questionnaireItemViewItem = questionnaireItemViewItem
         checkBox.text = questionnaireItemViewItem.questionnaireItemComponent.text
+        questionnaireItemViewItem.questionnaireResponseItemComponent.answer.also {
```
yigit, line 47: let or run

datacapture/src/main/java/com/google/android/fhir/datacapture/views/QuestionnaireItemCheckBoxViewHolderFactory.kt
```diff
@@ -44,6 +44,13 @@ object QuestionnaireItemCheckBoxViewHolderFactory : QuestionnaireItemViewHolderF
...
+          if (it.size == 1 && it[0].hasValueBooleanType()) {
```
yigit, line 48: checkbox.isChecked = answer.size == 1 && answer.first().valueBooleanType?.value == true

datacapture/src/main/java/com/google/android/fhir/datacapture/views/QuestionnaireItemDatePickerViewHolderFactory.kt
```diff
@@ -72,9 +76,27 @@ object QuestionnaireItemDatePickerViewHolderFactory : QuestionnaireItemViewHolde
         }
       }
 
+      @SuppressLint("NewApi") // java.time APIs can be used due to desugaring
       override fun bind(questionnaireItemViewItem: QuestionnaireItemViewItem) {
         this.questionnaireItemViewItem = questionnaireItemViewItem
         textView.text = questionnaireItemViewItem.questionnaireItemComponent.text
+        questionnaireItemViewItem.questionnaireResponseItemComponent.answer.also {
```
yigit, line 83: ditto. let or run

datacapture/src/main/java/com/google/android/fhir/datacapture/views/QuestionnaireItemDatePickerViewHolderFactory.kt
```diff
@@ -72,9 +76,27 @@ object QuestionnaireItemDatePickerViewHolderFactory : QuestionnaireItemViewHolde
...
+          if (it.size == 1 && it[0].hasValueDateType()) {
```
yigit, line 84: ditto. if can return the value

datacapture/src/main/java/com/google/android/fhir/datacapture/views/QuestionnaireItemDatePickerViewHolderFactory.kt
```diff
@@ -33,32 +34,35 @@ object QuestionnaireItemDatePickerViewHolderFactory : QuestionnaireItemViewHolde
...
+                  dayOfMonth
+                ).format(LOCAL_DATE_FORMATTER)
 
-                // Prefer the ICU library on Android N or above.
-                input.text = if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.N) {
-                  android.icu.text.DateFormat.getDateInstance().format(date)
-                } else {
-                  java.text.DateFormat.getDateInstance().format(date)
-                }
                 questionnaireItemViewItem.questionnaireResponseItemComponent.answer =
```
yigit, line 66:
```
val QuestionaireItemViewItem.answer
    get() = this.questionnaireResponseItemComponent.answer
    set(..) = ...
```
questionnaireItemViewItem,answer = foo

jingtang10:
thanks yigit :) ptal!

jingtang10 (commented):
ptal

datacapture/src/main/java/com/google/android/fhir/datacapture/views/QuestionnaireItemViewItem.kt
```diff
@@ -27,4 +28,10 @@ import org.hl7.fhir.r4.model.QuestionnaireResponse.QuestionnaireResponseItemComp
 data class QuestionnaireItemViewItem(
   val questionnaireItemComponent: QuestionnaireItemComponent,
   val questionnaireResponseItemComponent: QuestionnaireResponseItemComponent
-)
+) {
+  var singleAnswerOrNull
+    get() = questionnaireResponseItemComponent.answer.singleOrNull()
```
yigit, line 33: not sure if you meant this but singleOrNull returns null if there are more than 1 items:
https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.collections/single-or-null.html

jingtang10: I think basically there's something really wrong if we have multiple answers here. I'm happy with this rather lenient method rather than throwing exception or returning first.

yigit: :+1:

yigit: might be good to add a comment here to specify it is intentional.

yigit (approved):

kashyapj2793 (approved):

PR #136, opened 2021-01-06
Upgrade gradle to 6.7.1 and gradle plugin to 4.2.0-beta02

jingtang10:


deepankarb (approved):

deepankarb:
Wow I approved this. In the future we might want to stick to stable versions of AGP because beta versions aren't forward compatible.

Issue #131, opened 2020-12-28
Submit button not visible when Questionnaire occupies more space than single screen


kashyapj2793:
Currently testing how the datacapture lib behaves when items in the sample questionnaire are duplicated. When the questionnaire view goes over one screen the submit but does not appear.

Refer to video in issue #132

kashyapj2793:
@jingtang10 Can you please review the latest changes in the kj/sdc-binding-bug branch. I have added the submit button as part of the recycler view. It will be the last member in the view. It solves this problem

jingtang10:
@fredhersch can you please comment on what the UI should be in the case where the questionnaire is long and need scrolling? should the submit button always be at the bottom of the screen, or only appears when the user scrolls to the bottom of the screen?

fredhersch:
I suggest we position the save button in the toolbar in the top right. See this sample here

![Register patient](https://user-images.githubusercontent.com/34733/103646208-45383e00-4f94-11eb-8e27-37ed02999de3.jpg)

kashyapj2793:
I have positioned the button on the top right as suggested by Fred. I Changed the questionnaire_fragment linearLayout to a constraintLayout to do this. One thing to note is that since this is a fragment which is a part of the datacapturegallery activity the submit button is just below the app label but is in line with the Questionnaire title
![Submit_button_top_right](https://user-images.githubusercontent.com/48797364/103742150-5f1f6280-5020-11eb-83eb-9c07f7e80c0d.jpeg)

jingtang10:
@kashyapj2793 with the change made to the main activity now, will you be able to rebase this and move the submit button to the app bar?

kashyapj2793:
Sure Ill rebase and work on this :)

jingtang10:
thanks :) something like this may work: https://material.io/components/app-bars-top/android#regular-top-app-bar?

kashyapj2793:
![Screenshot_20210121-172017.jpg](https://user-images.githubusercontent.com/48797364/105347506-04cff580-5c0d-11eb-9ee8-7178cefe42ca.jpg) @fredhersch @jingtang10

fredhersch:
@kashyapj2793 This looks good to me. Is it possible to show the submit button with different styling (just border outline for example) to see how it sits on the toolbar? Thanks!

kashyapj2793:
[...]

PR #139, opened 2021-01-08
Submit button placement

kashyapj2793:
This PR addresses issue #131 . The questionnaire_fragment.xml used a linear layout which resulted in the submit button vanishing when the questionnaire is long. As advised in issue #131, the submit button should be pinned to the top right corner. I have used a constraintLayout to get this done.


jingtang10:
I think if we follow Fred's suggestion in #131 we'd have to consider not rendering the submit button as part of the library at all. So, the library is in charge of the questionnaire items and collecting answers, but the application would render a button and if the user clicks that button the application will ask the library (the questionnaire fragment) what has been recorded by the user.

I think if we do keep the button (and perhaps that should be an option) in the library, then it perhaps should be another view holder type in the recycler view. Same thing with the header (questionnaire title).

PR #146, opened 2021-01-13
Add basic rev. mgmt. support

deepankarb:
* Added Dao for writing local changes to db
* Diffing and squashing UPDATE type changes to yield a single UPDATE for sync


datacapturegallery/build.gradle
```diff
@@ -34,6 +34,10 @@ android {
         sourceCompatibility JavaVersion.VERSION_1_8
         targetCompatibility JavaVersion.VERSION_1_8
     }
+    packagingOptions {
```
deepankarb, line 37: Excluding to avoid duplication. See: https://stackoverflow.com/questions/35417464/why-are-gradle-build-packagingoptions-needed

fhirengine/src/main/java/com/google/fhirengine/db/Database.kt
```diff
@@ -139,4 +140,16 @@ interface Database {
     ): List<R>
 
     fun <R : Resource> search(query: Query): List<R>
+
+    /**
+     * Retrieves all [LocalChange]s for all [Resource]s. The [LocalChange]s are in a form which
+     * can be synced. Multiple changes for a single [Resource] are squashed into a single
+     * [LocalChange].
+     */
```
jingtang10, line 148:
```suggestion
    /**
     * Retrieves all [LocalChange]s for all [Resource]s, which can be used to update the remote FHIR server. Each [resource] will have at most one [LocalChange] (multiple changes are squashed).
     */
```

fhirengine/src/main/java/com/google/fhirengine/db/impl/LocalChange.kt
```diff
@@ -0,0 +1,77 @@
...
+        ]
+)
+data class LocalChange(
+    @PrimaryKey(autoGenerate = true)
+    val id: Long,
+    val resourceType: String,
+    val resourceId: String,
+    val timestamp: String = "",
+    val type: Type,
+    val diff: String
```
jingtang10, line 66: maybe call this content or payload? diff seems only suitable for update.

fhirengine/src/main/java/com/google/fhirengine/db/impl/dao/LocalChangeDao.kt
```diff
@@ -0,0 +1,237 @@
...
+        val updater = objectMapper.readerForUpdating(first)
+        val merged = updater.readValue(secondPatch, JsonPatch::class.java)
+        return merged.toString()
+    }
+
+    private fun mergeLocalChanges(first: LocalChange, second: LocalChange): LocalChange {
+        val type: Type
+        val diff: String
+        when (second.type) {
+            Type.UPDATE -> if (first.type.equals(Type.UPDATE)) {
```
jingtang10, line 178: use when rather than if

fhirengine/src/main/java/com/google/fhirengine/db/impl/dao/LocalChangeDao.kt
```diff
@@ -0,0 +1,237 @@
...
+                diff = applyPatch(first.diff, second.diff)
+            } else {
+                throw IllegalArgumentException("Don't know how to merge $first and $second.")
+            }
+            Type.DELETE -> {
+                type = Type.DELETE
+                diff = ""
+            }
+            else ->
+                throw IllegalArgumentException("Don't know how to merge $first and $second.")
```
jingtang10, line 192:
```suggestion
                throw IllegalArgumentException("Cannot merge local changes with type ${first.type} and ${second.type}.")
```

fhirengine/src/main/java/com/google/fhirengine/db/impl/dao/LocalChangeDao.kt
```diff
@@ -0,0 +1,237 @@
...
+                resourceId = resourceId,
+                timestamp = timestamp,
+                type = Type.DELETE,
+                diff = ""
+            )
+        )
+    }
+
+    /**
+     * Update a [Resource] with a JSON patch (RFC 6902).
```
jingtang10, line 131: I thinks this doesn't have to be a resourc at all? The documentation can be more generic. for example Update a json object with a json patch...

fhirengine/src/main/java/com/google/fhirengine/db/impl/dao/LocalChangeDao.kt
```diff
@@ -0,0 +1,237 @@
...
+     */
+    private fun applyPatch(resourceString: String, patchString: String): String {
+        val objectMapper = ObjectMapper()
+        val resourceJson = objectMapper.readValue(resourceString, JsonNode::class.java)
+        val patchJson = objectMapper.readValue(patchString, JsonPatch::class.java)
+        return patchJson.apply(resourceJson).toString()
+    }
+
+    /**
+     * JSON patch diff b/w two [Resource]s.
```
jingtang10, line 141: nit: don't use abbreviations b/w.

fhirengine/src/main/java/com/google/fhirengine/db/impl/dao/LocalChangeDao.kt
```diff
@@ -0,0 +1,237 @@
...
+                resourceType = second.resourceType,
+                type = type,
+                diff = diff
+        )
+    }
+
+    /**
+     * Squash the changes by merging them two at a time.
+     */
+    fun squash(localChanges: List<LocalChange>): LocalChange =
```
jingtang10, line 206: I think this function needs to be 1) static (in a companion object) 2) in a separate class. it doesn't seem to belong to local change dao. a utility class seems more appropriate.

you might move the other functions this function uses too.

jingtang10 (commented):
thanks deepankar! haven't looked at the tests.

fhirengine/src/main/java/com/google/fhirengine/Util.kt
```diff
@@ -0,0 +1,35 @@
...
+ * Utility function to add timestamps to database rows, e.g. last local update timestamp.
+ */
+@SuppressLint("NewApi")
+fun Date.toTimeZoneString(): String {
+    val simpleDateFormat = DateTimeFormatter.ofPattern(
+            "yyyy-MM-dd'T'HH:mm:ss.SSSXXX",
+            Locale.US
+    ).withZone(ZoneId.systemDefault())
+    return simpleDateFormat.format(this.toInstant())
+}
```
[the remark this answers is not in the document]
deepankarb: It is indeed and the idea was to refactor `ResourceSynchronizer.kt` after this PR gets merged :)

deepankarb:
Updated with fixes based on comments. @jingtang10 PTAL.

fhirengine/src/main/java/com/google/fhirengine/db/impl/dao/LocalChangeUtils.kt
```diff
@@ -0,0 +1,122 @@
+/*
```
jingtang10, line 1: did you add tests for this static class? if not, please create an issue as a follow up :)

i think one of the benefits of having this static class is that the tests can be written very independently.

fhirengine/src/main/java/com/google/fhirengine/db/impl/dao/LocalChangeUtils.kt
```diff
@@ -0,0 +1,122 @@
...
+        )
+        val targetJson = objectMapper.readValue(
+            parser.encodeResourceToString(target),
+            JsonNode::class.java
+        )
+        val jsonDiff = JsonDiff.asJson(sourceJson, targetJson)
+        if (jsonDiff.size() == 0)
+            Log.w(
+            "ResourceDao",
+            "Trying to UPDATE resource ${target.resourceType}/${target.id} with no changes"
```
jingtang10, line 118: this error message seems wrong

fhirengine/src/main/java/com/google/fhirengine/db/impl/dao/LocalChangeUtils.kt
```diff
@@ -0,0 +1,122 @@
...
+        val objectMapper = ObjectMapper()
+        val sourceJson = objectMapper.readValue(
+            parser.encodeResourceToString(source),
+            JsonNode::class.java
+        )
+        val targetJson = objectMapper.readValue(
+            parser.encodeResourceToString(target),
+            JsonNode::class.java
+        )
+        val jsonDiff = JsonDiff.asJson(sourceJson, targetJson)
```
jingtang10, line 114: I actually think this might look better and more readable if you inline source and target jsons?

```suggestion
        val jsonDiff = JsonDiff.asJson(
            objectMapper.readValue(
                parser.encodeResourceToString(source),
                JsonNode::class.java
            ),
            objectMapper.readValue(
                parser.encodeResourceToString(target),
                JsonNode::class.java
            )
        )
```

fhirengine/src/main/java/com/google/fhirengine/db/impl/DatabaseImpl.kt
```diff
@@ -65,17 +67,27 @@ internal class DatabaseImpl(
...
+    @Transaction
     override fun <R : Resource> insert(resource: R) {
         resourceDao.insert(resource)
+        localChangeDao.addInsert(resource)
     }
 
+    @Transaction
     override fun <R : Resource> insertAll(resources: List<R>) {
         resourceDao.insertAll(resources)
+        // TODO: impl insertAll in LocalChangeDao
```
jingtang10, line 83: this seems to have been implemented?

fhirengine/src/main/java/com/google/fhirengine/db/impl/dao/LocalChangeUtils.kt
deepankarb: Created #187

fhirengine/src/main/java/com/google/fhirengine/db/impl/DatabaseImpl.kt
deepankarb: There's discrepancy between `ResourceDao`, which replaces a duplicate row upon insert, and `LocalChangeDao` which throws upon duplicate row insertion. We need to decide and update tests accordingly.

jingtang10 (approved):
@yigit can you also take a look please? thanks!

fhirengine/src/androidTest/java/com/google/fhirengine/db/impl/DatabaseImplTest.kt
```diff
@@ -99,42 +99,35 @@ class DatabaseImplTest {
...
         val resourceNotFoundException =
-            assertThrows(
-                ResourceNotFoundInDbException::class.java,
-                { database.select(Patient::class.java, "nonexistent_patient") }
-            )
-        assertEquals(
-            "Resource not found with type ${ResourceType.Patient.name} and id nonexistent_patient!",
-            resourceNotFoundException.message
-        )
+            assertThrows(ResourceNotFoundInDbException::class.java) {
```
yigit, line 126: nit: indentation seems off here though i'm surprised spotless didn't catch it.

fhirengine/src/main/java/com/google/fhirengine/Util.kt
```diff
@@ -0,0 +1,35 @@
...
+import java.time.ZoneId
+import java.time.format.DateTimeFormatter
+import java.util.Date
+import java.util.Locale
+
+/**
+ * Utility function to format a [Date] object using the system's default locale.
+ */
+@SuppressLint("NewApi")
+fun Date.toTimeZoneString(): String {
```
yigit, line 29: `internal` unless we want this to be a public API.

fhirengine/src/main/java/com/google/fhirengine/db/impl/DatabaseImpl.kt
```diff
@@ -101,12 +113,14 @@ internal class DatabaseImpl(
         insertAll(resources)
     }
 
+    @Transaction
     override fun <R : Resource> delete(clazz: Class<R>, id: String) {
         val type = getResourceType(clazz)
         resourceDao.deleteResource(
```
yigit, line 119: we should return int from deleteResorce dao which will return # of deleted rows. this way, we can avoid adding it to localChangeDao if it didn't exist in the first place.

fhirengine/src/main/java/com/google/fhirengine/db/impl/DatabaseImpl.kt
```diff
@@ -161,6 +175,20 @@ internal class DatabaseImpl(
         resourceDao.getResources(query.getSupportSQLiteQuery())
             .map { iParser.parseResource(it) as R }
 
+    override fun getAllLocalChanges(): List<LocalChange> {
+        return localChangeDao.getAllLocalChanges()
+                .groupBy { it.resourceId + it.resourceType }
```
yigit, line 180: better to use a pair here:
`it.resourceId to it.resourceType`

fhirengine/src/main/java/com/google/fhirengine/db/impl/dao/LocalChangeDao.kt
```diff
@@ -0,0 +1,151 @@
...
+ * e.g. an INSERT or UPDATE. The UPDATES (diffs) are stored as RFC 6902 JSON patches.
+ * When a resource needs to be synced, all corresponding LocalChanges are 'squashed' to create a
+ * a single LocalChange to sync with the server.
+ */
+@Dao
+internal abstract class LocalChangeDao {
+
+    lateinit var iParser: IParser
+
+    @Insert(onConflict = OnConflictStrategy.ABORT)
```
yigit, line 43: we don't have anything unique in thsi table besides the auto-increment key so I guess we don't need abort?

fhirengine/src/main/java/com/google/fhirengine/db/impl/dao/LocalChangeDao.kt
```diff
@@ -0,0 +1,151 @@
...
+    open fun addInsertAll(resources: List<Resource>) {
+        resources.forEach { resource ->
+            addInsert(resource)
+        }
+    }
+
+    fun addInsert(resource: Resource) {
+        val resourceId = resource.id
+        val resourceType = resource.resourceType
+        val localChanges = getLocalChanges(
```
yigit, line 56: since you only need the last change type, youi don't need to load all changes.

you can write a query that gets the type of last change:
something like:
```
@Query("select type from localChanges where resourceId = :resourceId AND resourceType = :resourceType ORDER BY ID ASC LIMIT 1)
protected fun getLastChangeType(resourceId:String, resourceType:ResourceType):LocalCHange.Type?
```

yigit: same for other methods below.

fhirengine/src/main/java/com/google/fhirengine/db/Database.kt
```diff
@@ -139,4 +140,16 @@ interface Database {
...
+    fun getAllLocalChanges(): List<LocalChange>
+
+    /**
+     * Remove the [LocalChange]s with given ids. Call this after a successful sync.
+     */
+    fun <R : Resource> deleteUpdates(clazz: Class<R>, id: String)
```
yigit, line 154: this cannot work like this because there might be more changes in the database between the time developer queried pending changes and by the time they finish uploading.

we need to given them some sorth of token (or ids of updates) and expect them to give those ids back to avoid any race condition

yigit (changes requested):

fhirengine/src/main/java/com/google/fhirengine/db/Database.kt
yigit:
```
val res1 = fhir.getResource(...)
val updated = update(res1) // some changes
fhir.update(res1, updated)
fhir.getLocalChanges() // this will give me the update1
fhir.update(updated, updated2)
fhir.deleteUpdates(res1) // this will also delete update2
```
so instead:

```
(token, data) = fhir.getLocalChanges(...)
...
fhir.deleteLocalCHanges(token) //token can have the ids, so we can delete the rigth content
```

fhirengine/src/main/java/com/google/fhirengine/db/impl/dao/LocalChangeDao.kt
```diff
@@ -0,0 +1,151 @@
...
+        )
+    }
+
+    @Query(
+        """
+        SELECT *
+        FROM LocalChange
+        WHERE LocalChange.resourceId = (:resourceId)
+        AND LocalChange.resourceType  = (:resourceType)
+        ORDER BY LocalChange.timestamp ASC"""
```
yigit, line 129: better to order by id as it auto-increments.

fhirengine/src/main/java/com/google/fhirengine/db/impl/dao/LocalChangeDao.kt
```diff
@@ -0,0 +1,151 @@
...
+    )
+    abstract fun getLocalChanges(resourceId: String, resourceType: String): List<LocalChange>
+
+    @Query(
+        """
+        SELECT *
+        FROM LocalChange
+        ORDER BY LocalChange.timestamp ASC"""
```
yigit, line 137: ditto

fhirengine/src/main/java/com/google/fhirengine/db/Database.kt
jingtang10: can that token just be the id of the local change?

jingtang10:
Discussed with @yigit and concluded that the new APIs for local changes should all be internal -- they are used as part of our lib for synchronization, not for developers to do anything.

fhirengine/src/main/java/com/google/fhirengine/db/Database.kt
deepankarb: @yigit suggested a wrapper around List<Id>. Something like `internal data class Token(val ids: List<Int>)`

fhirengine/src/androidTest/java/com/google/fhirengine/db/impl/DatabaseImplTest.kt
deepankarb: I thought split assignment and expressions need to be indented on the first new line.

fhirengine/src/main/java/com/google/fhirengine/db/Database.kt
deepankarb: Implemented.

jingtang10:
@googlebot I consent.

jingtang10 (approved):
Looks good! 🎉

`LocalChangeUtils.kt` is a really critical class -- can you add more documentation in the class  in general please? like add some examples in the doc each method.

yigit (approved):

PR #162, opened 2021-01-21
Fix for refreshing patient list.

pankajkatoch:
PatientListActivity observes a MutableLiveData reference, and on change submits the list to the PatientItemRecyclerViewAdapter. The RecyclerViewAdapter only compares and diffs when the list reference being submitted changes, otherwise it doesn't take any action.

Submitting a copy of the list to circumvent the RecyclerViewAdapter optimisation.

Please suggest if there's a better way to handle this.

### DIFF
--- a/example/src/main/java/com/google/fhirengine/example/PatientListActivity.kt
+++ b/example/src/main/java/com/google/fhirengine/example/PatientListActivity.kt
@@ -72,7 +72,7 @@ class PatientListActivity : AppCompatActivity() {
                 // the MutableLiveData (list of PatientItems) has changed.
                 // submit a copy of the list, otherwise adapter still thinks it's same unchanged
                 // list.
-                adapter.submitList(it?.toMutableList())
+                adapter.submitList(it)
             })
 
         patientListViewModel.getObservations().observe(this,
--- a/example/src/main/java/com/google/fhirengine/example/PatientListActivity.kt
+++ b/example/src/main/java/com/google/fhirengine/example/PatientListActivity.kt
@@ -69,7 +69,10 @@ class PatientListActivity : AppCompatActivity() {
         patientListViewModel.getSearchedPatients().observe(this,
             Observer<List<PatientListViewModel.PatientItem>> {
                 Log.d("PatientListActivity", "Submitting ${it.count()} patient records")
-                adapter.submitList(it)
+                // the MutableLiveData (list of PatientItems) has changed.
+                // submit a copy of the list, otherwise adapter still thinks it's same unchanged
+                // list.
+                adapter.submitList(it?.toMutableList())
             })
 
         patientListViewModel.getObservations().observe(this,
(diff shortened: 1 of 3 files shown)


pankajkatoch:
Updating the PR with these changes:

- Use private MutableLiveData, and but return LiveData to be observed.
- To ensure that LiveData.value reference changes, Return a copy of the list while returning patient items.

@yigit @jingtang10 Please review.

example/src/main/java/com/google/fhirengine/example/data/SamplePatients.kt
```diff
@@ -65,7 +65,8 @@ class SamplePatients {
             createPatientItem(index + 1, fhirPatient)
         }?.let { patients.addAll(it) }
 
-        return patients
+        // Return a cloned List
+        return patients.toMutableList()
```
jingtang10, line 69: does `patients` even need to be a val in this class? can't it just be a local variable?

pankajkatoch: Thanks! No they don't need to be at the class level at all. Moved to local and simplified.

example/src/main/java/com/google/fhirengine/example/PatientListActivity.kt
```diff
@@ -69,6 +69,9 @@ class PatientListActivity : AppCompatActivity() {
         patientListViewModel.getSearchedPatients().observe(this,
             Observer<List<PatientListViewModel.PatientItem>> {
                 Log.d("PatientListActivity", "Submitting ${it.count()} patient records")
+                // the MutableLiveData (list of PatientItems) has changed.
+                // submit a copy of the list, otherwise adapter still thinks it's same unchanged
+                // list.
```
jingtang10, line 74:
```suggestion
                // The MutableLiveData (list of PatientItems) has changed.
                // Submit a copy of the list, otherwise adapter still thinks it is the same unchanged
                // list.
```

jingtang10 (approved):
@yigit can you also take a final look please? thanks!

example/src/main/java/com/google/fhirengine/example/PatientListActivity.kt
pankajkatoch: I'll be careful about capitalization and avoiding apostrophes in the comments.
However the old comments are no longer valid, and need to be removed entirely. Removed them.

example/src/main/java/com/google/fhirengine/example/data/SamplePatients.kt
```diff
@@ -32,9 +32,6 @@ private const val MAX_RESOURCE_COUNT = 20
  * PatientListViewModel.
  */
 class SamplePatients {
-    private val patients: MutableList<PatientListViewModel.PatientItem> = mutableListOf()
-    private val observations: MutableList<PatientListViewModel.ObservationItem> = ArrayList()
-
     // The resource bundle with Patient objects.
     private var fhirBundle: Bundle? = null
```
yigit, line 36: is this used?

yigit: this almost feels like a utility class (could be an object) that has methods to convert a json string into patients.

PR #166, opened 2021-01-22
Add patient registration form to the sdc gallery

jingtang10:


pld (commented):
Very cool, have we considered how this would support multiple languages?

jingtang10:
> Very cool, have we considered how this would support multiple languages?

@pld So the basic idea is to use the [translation extension](http://hl7.org/fhir/languages.html##ext) (see 7.15.3). If a string in the questionnarie resource has the translation extension that has a language code that matches the system's language code, then we will use the value from that translation extension.

For example: we might have a questionnaire:
```
{
  "id": "client-registration-sample",
  "item": [{
    "linkId": "1",
    "text": "The first question",
    "extension": [{ 
      "url" : "http://hl7.org/fhir/StructureDefinition/translation",
      "extension" : [{
        "url" : "lang", 
        "valueCode" : "es"
      },{
        "url" : "content",
        "valueString" : "La primera pregunta"
      }]
    }]
  }]
}
```

The sdc library will call the following API to figure out what's the system language:

```
Locale.getDefault().toLanguageTag()
```

If the system language matches the one defined in one of the translation extensions, we will use that to render the UI widget, otherwise, we will use the default string. In other words, if the system has Spanish as default language (language code `es`), the library will create the UI using the string "La primera pregunta". Otherwise, it will just render "The first question".

The translation extension defined in FHIR allows any text (anything of the string type) to contain multiple languages. The library needs to respect the extension and select the correct language to display that matches the system settings. For example, if the system language of the device is French, the library should attempt to read the French translations in the translation extensions to render the questionnaire.

The translation extension uses a value set that conforms to IETF BCP 47 so theoretically these codes should match. In reality, however, more lenient matching might be needed. For example, a translation might be provided if the 2-letter language code matches even though the country code is different. This is future work.

joiskash (approved):

PR #183, opened 2021-01-25
Hide group header if the text is empty

jingtang10:
This is useful when the group isn't used as a header but for logical grouping purposes. In those cases we don't want to end up with a big gap (with height of one text line height).

### DIFF
--- a/datacapture/src/main/res/layout/questionnaire_item_group_header_view.xml
+++ b/datacapture/src/main/res/layout/questionnaire_item_group_header_view.xml
@@ -1,4 +1,5 @@
-<?xml version="1.0" encoding="utf-8"?><!--
+<?xml version="1.0" encoding="utf-8"?>
+<!--
  Copyright 2020 Google LLC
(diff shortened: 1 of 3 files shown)


datacapture/src/main/res/layout/questionnaire_item_group_header_view.xml
```diff
@@ -1,5 +1,4 @@
-<?xml version="1.0" encoding="utf-8"?>
-<!--
+<?xml version="1.0" encoding="utf-8"?><!--
```
yigit, line 1: nit

yigit (approved):

datacapture/src/main/res/layout/questionnaire_item_group_header_view.xml
jingtang10: done

jingtang10 (commented):
thanks yigit!

Issue #153, opened 2021-01-15
Add dropdown list support in the sdc library


jingtang10:
This should make use of the extension item control: http://hl7.org/fhir/uv/sdc/2019May/rendering.html#control-appearance

kashyapj2793:
Suppose there is not item-control extension, would this be the default widget for choice questions?

jingtang10:
I think it makes sense for this to be default as opposed to radio button. We can also define a heuristic that says if the number of items is less than 3 or 4 we use radio button instead. But we don't have to do that now. Default to this is fine. FYI @fredhersch .

jingtang10:
Perhaps use this as an example as well: https://www.hl7.org/fhir/questionnaire-example-bluebook.html

It's a good exmaple becuase it doesn't use valuesets for the choices questions.

PR #188, opened 2021-01-27
Add support for drop down

joiskash:
#153

### DIFF
--- a/datacapture/src/main/java/com/google/android/fhir/datacapture/QuestionnaireItemViewHolderType.kt
+++ b/datacapture/src/main/java/com/google/android/fhir/datacapture/QuestionnaireItemViewHolderType.kt
@@ -37,13 +37,7 @@ enum class QuestionnaireItemViewHolderType(val value: Int) {
     DROP_DOWN(7);
 
     companion object {
-        const val EXTENSION_ITEM_CONTROL_URL =
-            "http://hl7.org/fhir/StructureDefinition/questionnaire-itemControl"
         private val VALUES = values()
         fun fromInt(value: Int) = VALUES[value]
-        fun fromString(code: String?) = when (code) {
-            "drop-down" -> VALUES[7]
-            else -> null
-        }
     }
 }
--- a/datacapture/src/main/java/com/google/android/fhir/datacapture/QuestionnaireItemViewHolderType.kt
+++ b/datacapture/src/main/java/com/google/android/fhir/datacapture/QuestionnaireItemViewHolderType.kt
@@ -33,10 +33,17 @@ enum class QuestionnaireItemViewHolderType(val value: Int) {
     EDIT_TEXT_SINGLE_LINE(3),
     EDIT_TEXT_MULTI_LINE(4),
     EDIT_TEXT_INTEGER(5),
-    EDIT_TEXT_DECIMAL(6);
+    EDIT_TEXT_DECIMAL(6),
+    DROP_DOWN(7);
 
     companion object {
+        const val EXTENSION_ITEM_CONTROL_URL =
+            "http://hl7.org/fhir/StructureDefinition/questionnaire-itemControl"
         private val VALUES = values()
         fun fromInt(value: Int) = VALUES[value]
+        fun fromString(code: String?) = when (code) {
+            "drop-down" -> VALUES[7]
+            else -> null
+        }
     }
 }
--- a/datacapture/src/main/java/com/google/android/fhir/datacapture/views/QuestionnaireItemDropDownViewHolderFactory.kt
+++ b/datacapture/src/main/java/com/google/android/fhir/datacapture/views/QuestionnaireItemDropDownViewHolderFactory.kt
@@ -22,11 +22,10 @@ import android.widget.AdapterView
 import android.widget.ArrayAdapter
 import android.widget.AutoCompleteTextView
 import com.google.android.fhir.datacapture.R
+import com.google.android.fhir.datacapture.getResponseAnswer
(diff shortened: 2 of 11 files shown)


datacapture/src/main/java/com/google/android/fhir/datacapture/QuestionnaireItemViewHolderType.kt
```diff
@@ -33,10 +33,17 @@ enum class QuestionnaireItemViewHolderType(val value: Int) {
     EDIT_TEXT_SINGLE_LINE(3),
     EDIT_TEXT_MULTI_LINE(4),
     EDIT_TEXT_INTEGER(5),
-    EDIT_TEXT_DECIMAL(6);
+    EDIT_TEXT_DECIMAL(6),
+    DROP_DOWN(7);
 
     companion object {
+        const val EXTENSION_ITEM_CONTROL_URL =
+            "http://hl7.org/fhir/StructureDefinition/questionnaire-itemControl"
```
jingtang10, line 41: i'd keep this logic outside of this class. either as part of the adapter or create some kind of converter.

datacapture/src/main/java/com/google/android/fhir/datacapture/views/QuestionnaireItemDropDownViewHolderFactory.kt
```diff
@@ -0,0 +1,132 @@
...
+                                .build()
+                        )
+                    } else if (it.value.hasStringValue()) {
+                        answerOptionString.add(it.value.stringValue.value.toString())
+                        QuestionnaireResponse.Item.Answer.ValueX.newBuilder()
+                            .setStringValue(it.value.stringValue)
+                            .build()
+                    }
+                }
+            }
```
jingtang10, line 130: I think 1) this function can be made into an extension function to AnswerOption. 2) this function should be in a separate class so that this can be reused (e.g. for radio button group view type).

what about a new class called MoreAnswerOptions.kt and then inside have something like
```
fun AnswerOption.getString(): String {
    if(this.value.hasCoding) {
        ...
    } else if (...) {
        ...
    }
    ...
}
```

datacapture/src/main/java/com/google/android/fhir/datacapture/views/QuestionnaireItemDropDownViewHolderFactory.kt
```diff
@@ -0,0 +1,132 @@
...
+object QuestionnaireItemDropDownViewHolderFactory : QuestionnaireItemViewHolderFactory(
+    R.layout.questionnaire_item_drop_down_view
+) {
+    override fun getQuestionnaireItemViewHolderDelegate() =
+        object : QuestionnaireItemViewHolderDelegate {
+            private lateinit var textInputLayout: TextInputLayout
+            private lateinit var questionnaireItemViewItem: QuestionnaireItemViewItem
+            private lateinit var context: Context
+            private var answerOptionString = arrayListOf<String>()
+            private var answerOptionValueX = arrayListOf<QuestionnaireResponse.Item.Answer.ValueX>()
```
jingtang10, line 40: do these have to be here? can `QuestionnaireResponse.Item.Answer.ValueX` be generated by looking up the order of potential answers in `QuestionnaireItemViewItem` only after an option has been chosen ?

datacapture/src/main/java/com/google/android/fhir/datacapture/views/QuestionnaireItemViewItem.kt
```diff
@@ -44,4 +46,18 @@ data class QuestionnaireItemViewItem(
                 questionnaireResponseItemBuilder.addAnswer(it)
             }
         }
+
+    val questionnaireItemControlType
+        get() = getItemControlFromExtensions(questionnaireItem.extensionList)
+
+    // Returns item control or null
+    private fun getItemControlFromExtensions(extensions: MutableList<Extension>):
```
jingtang10, line 54: can this be folded into the val above? why create another private val?

I think the return type should be the the value of the extension itself (string?). Again this class (like the enum class) shouldn't be making decisions on which view type to use. That decision needs to reside in the adapter.

jingtang10 (commented):
thanks @joiskash for the fast work. some intial comments.

datacapture/src/main/java/com/google/android/fhir/datacapture/views/QuestionnaireItemViewItem.kt
joiskash: Since we are comparing the Extension.value.code.value with "drop-down" I am directly returning the string . Ideally we should pass the [valueset ](http://hl7.org/fhir/R4/valueset-questionnaire-item-control.html) to the library. So that this is not hardcoded

datacapture/src/main/java/com/google/android/fhir/datacapture/views/QuestionnaireItemDropDownViewHolderFactory.kt
joiskash: done

datacapture/src/main/java/com/google/android/fhir/datacapture/QuestionnaireItemViewHolderType.kt
joiskash: moved it to the adapter

jingtang10:
hey @joiskash please let me know when this is ready for review (and convert it from a draft to a PR please). thanks :)

joiskash:
Refer to #209 Closing this PR . It is in a dirty state.

PR #205, opened 2021-02-01
Enable core lib desugaring on fhirengine

deepankarb:
Set `coreLibraryDesugaringEnabled true` in `fhirengine.gradle` `compileOptions`

### DIFF
--- a/fhirengine/build.gradle
+++ b/fhirengine/build.gradle
@@ -34,6 +34,9 @@ android {
         }
     }
     compileOptions {
+        // Flag to enable support for the new language APIs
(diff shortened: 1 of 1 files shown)


jingtang10:
I think you need to add the new line in the dependency block too:

```
coreLibraryDesugaring deps.desugar
```

deepankarb:
> I think you need to add the new line in the dependency block too:
> 
> ```
> coreLibraryDesugaring deps.desugar
> ```

@jingtang10 updated PTAL.

jingtang10 (approved):
thanks deepankar!!!

PR #213, opened 2021-02-02
Use ListAdapter instead of RecyclerView.Adapter

jingtang10:


deepankarb (approved):

datacapture/src/main/java/com/google/android/fhir/datacapture/QuestionnaireItemAdapter.kt
```diff
@@ -95,6 +95,20 @@ internal class QuestionnaireItemAdapter(
...
+    ): Boolean {
+        return oldItem == newItem
+    }
 
-    override fun getItemCount() = questionnaireItemViewItemList.size
+    override fun areContentsTheSame(
+        oldItem: QuestionnaireItemViewItem,
+        newItem: QuestionnaireItemViewItem
+    ): Boolean {
+        return oldItem.questionnaireItem.linkId == newItem.questionnaireItem.linkId
```
yigit, line 112: this is the opposite.
contents the same is item ==, items the same is id ==

jingtang10: done.

jingtang10:
thanks both!

PR #220, opened 2021-02-04
Show/hide questions based on the enablement evaluation

### LINKED ISSUE #165: Support enableWhen in the sdc library
http://build.fhir.org/ig/HL7/sdc/behavior.html#other-control

> This controls which questions, groups and even display items should be displayed based on answers to other questions within the response. An item is 'enabled' based on whether the referenced questions have one (or more) of the answers as specified by the enableWhen element. If multiple enableWhen elements are present on an item, the interpretation is governed by the enableBehavior element. Elements that are not enabled should either be hidden from the user or at least grayed-out and non-selectable and non-editable. Any behavioral constraints associated with non-enabled elements (such as required or minOccurs are ignored and no answers are stored for non-enabled content.
> 
> For more sophisticated enableWhen behavior, the enableWhenExpression element can be used instead.

jingtang10:
close #165


datacapture/src/main/java/com/google/android/fhir/datacapture/QuestionnaireViewModel.kt
```diff
@@ -32,7 +33,27 @@ class QuestionnaireViewModel(state: SavedStateHandle) : ViewModel() {
     internal val questionnaireResponseBuilder = QuestionnaireResponse.newBuilder()
 
     /** The list of [QuestionnaireItemViewItem] to be used for the [RecyclerView]. */
```
yigit, line 35:
```
private val  modificationCount = MutableStateFlow<Int>(0)

private val notifyChangeCallback = {
    modificationCount.value += 1
}
val questionaires: StateFlow<List<QuestionnaireItemViewItem>> = modificationCount.map {
   getQuestionnaireItemViewItemList()
}.stateIn(viewModelScope)

private getQuestionnaireItemViewItemList() {....}
```

```
// in fragment
onCreate() {
    lifecycle.launchWhenCreated {
        viewModel.questionaires.collect { adapter.submitList(it) }
   }
}
```

then we could make `QuestionnaireItemViewItem` could have a notifyChange method and view model can set itself as the listener and automatically update itself.

in the adapter code:

```
bind {
   onChangeListener {
      item.notifyChange()
   }
}
```

jingtang10:
Updated the PR with @yigit 's suggestion. Tested locally and works.

datacapture/src/main/java/com/google/android/fhir/datacapture/QuestionnaireFragment.kt
```diff
@@ -50,8 +60,7 @@ class QuestionnaireFragment : Fragment() {
 
     override fun onViewCreated(view: View, savedInstanceState: Bundle?) {
         val recyclerView = view.findViewById<RecyclerView>(R.id.recycler_view)
-        val adapter = QuestionnaireItemAdapter()
-        adapter.submitList(viewModel.questionnaireItemViewItemList)
+        adapter = QuestionnaireItemAdapter()
```
yigit, line 63: also,keeping a reference to adapter in the fragment is wrong for the same reason (it will leak the recyclerview because recyclerview is observing the adapter, hence leak the whole fragment view).

if you start collection here (the right place anyways) you won't need a reference to the adapter in the fragment.

datacapture/src/main/java/com/google/android/fhir/datacapture/QuestionnaireViewModel.kt
```diff
@@ -31,8 +37,38 @@ class QuestionnaireViewModel(state: SavedStateHandle) : ViewModel() {
...
 
+    private val  modificationCount = MutableStateFlow(0)
+
+    private val notifyChangeCallback = {
+        modificationCount.value += 1
+    }
+    val questionnaires: Flow<List<QuestionnaireItemViewItem>> = modificationCount.map {
+        getQuestionnaireItemViewItemList()
+    }
+        //.stateIn(viewModelScope)
```
yigit, line 48: did this break somehow?
if not, it is good to have but not super curucial (it is the difference between recomputing the question list on fragment recreate or not)

yigit (commented):
lgtm. small changes

datacapture/src/main/java/com/google/android/fhir/datacapture/QuestionnaireViewModel.kt
jingtang10: I couldn't find the stateIn API... even after changing the coroutine version to 1.4.2.... am I missing some dependency?

datacapture/src/main/java/com/google/android/fhir/datacapture/QuestionnaireFragment.kt
```diff
@@ -24,11 +24,21 @@ import androidx.appcompat.view.ContextThemeWrapper
...
 import androidx.recyclerview.widget.RecyclerView
+import kotlinx.coroutines.flow.collect
 
 class QuestionnaireFragment : Fragment() {
     private val viewModel: QuestionnaireViewModel by viewModels()
+    private lateinit var adapter: QuestionnaireItemAdapter
+
+    override fun onCreate(savedInstanceState: Bundle?) {
+        super.onCreate(savedInstanceState)
+        viewLifecycleOwner.lifecycleScope.launchWhenCreated {
```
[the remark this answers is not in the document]
jingtang10: done

datacapture/src/main/java/com/google/android/fhir/datacapture/QuestionnaireFragment.kt
jingtang10: done

joiskash (approved):

Issue #164, opened 2021-01-22
Support minValue maxValue in the sdc library


jingtang10:
See http://build.fhir.org/ig/HL7/sdc/behavior.html

> minValue	question (where type='date', 'dateTime', 'time', 'decimal', 'integer')
> This extension is used for questions with simple quantitative types. The extension allows establishing the lower bound for an answer. For example, ensuring that birth dates are >= 1990-01-01 or an integer is greater than or equal to 0. (Note that the integer and decimal types both allow negative numbers if not constrained.) This extension can be used together with the cqf-calculatedValue extension to establish a dynamic limit. For example, ensuring that a date of birth will result in a maximum age by setting minValue with an expression of 'today() - 18 years'.
> 
> The minValue SHALL be less than or equal to the maxValue, if both are specified. If min and/or max are calculated values, care should be taken to ensure that they cannot ever conflict. I.e. It should never be possible for calculated max to be less than calculated min. If this does occur, systems SHALL prohibit a value from being specified for the element for so long as the conflict exists. If the question is required, this may mean that it is not possible to mark the QuestionnaireResponse as "completed".
> 
> maxValuebquestion (where type='date', 'dateTime', 'time', 'decimal', 'integer')
> This extension is used for questions with simple quantitative types. The extension allows establishing the upper bound for an answer. For example, ensuring that birth dates are >= 1990-01-01 or an integer is greater than or equal to 0. (Note that the integer and decimal types both allow negative numbers if not constrained.) This extension can be used together with the cqf-calculatedValue extension to establish a dynamic limit. For example, ensuring that a date of birth will result in a minimum age by setting maxValue with an expression of 'today() - 18 years'.
> 
> The minValue SHALL be less than or equal to the maxValue, if both are specified. If min and/or max are calculated values, care should be taken to ensure that they cannot ever conflict. I.e. It should never be possible for calculated max to be less than calculated min. If this does occur, systems SHALL prohibit a value from being specified for the element for so long as the conflict exists. If the question is required, this may mean that it is not possible to mark the QuestionnaireResponse as "completed".

cqf-calculatedValue support is out of scope

jingtang10:
@kunjan8794 are you interested in taking this?

kunjan8794:
Surely, I'll take this up. Thanks 👍🏻

jingtang10:
Awesome :) There are a few different data types to support (see comment above) but you can certainly start with one type (e.g. integer) and send a pr.

PR #224, opened 2021-02-08
Support for minValue & maxValue extension for the data type integer

kunjan8794:
@jingtang10 @joiskash Please review the pr.

closes https://github.com/google/android-fhir/issues/164


kunjan8794:
@googlebot I signed it!

kunjan8794:
@googlebot I signed it!

kunjan8794:
@googlebot I signed it!

jingtang10:
Thanks for the PR @kunjan8794. My first comment is that I think the validation logic should not be in the view holders. This is because there migth be multiple view types for the same data type (input box for integer but also slider for integer) but the validation logic for the same data type is always the same (there should be only one validator for integer type).

Can you please create a new validation package under datacapture and add the following interface (or something similar)?

```
interface Validator {
    fun validate(questionnaireItem: Questionnaire.Item, questionnaireResponseItem: QuestionnaireResponse.Item): ValidationResult
}
```

and the result class:

```
data class ValidationResult(val pass: Boolean, val errorMsgs: List<String>)
```

jingtang10 (changes requested):
Thanks for the PR @kunjan8794. My first comment is that I think the validation logic should not be in the view holders. This is because there migth be multiple view types for the same data type (input box for integer but also slider for integer) but the validation logic for the same data type is always the same (there should be only one validator for integer type).

Can you please create a new validation package under datacapture and add the following interface (or something similar)?

```
interface Validator {
    fun validate(questionnaireItem: Questionnaire.Item, questionnaireResponseItem: QuestionnaireResponse.Item): ValidationResult
}
```

and the result class:

```
data class ValidationResult(val pass: Boolean, val errorMsgs: List<String>)
```

kunjan8794:
@jingtang10 sure, I will do the changes as per the suggestion

kunjan8794:
@jingtang10 Please review it

datacapture/src/main/java/com/google/android/fhir/datacapture/validation/QuestionnaireItemIntegerValidator.kt
```diff
@@ -0,0 +1,62 @@
...
+                    }
+
+                    QuestionnaireItemValidator.EXTENSION_MIN_VALUE -> {
+                        if (validateMinValue(extension, answer)) {
+                            val messages = listOf("Error Value!")
+                            questionnaireResponseItemBuilder.clearAnswer()
+                            return ValidationResult(false, messages)
+                        }
+                    }
+                }
```
jingtang10, line 39: This is difficult to extend. You'll end up in a situation where you have way too many branches of `when` very quickly.

A couple of things to consider.

Create extensions to `Questionnaire.Item`. Have something like
```
internal fun Questionnaire.Item.maxValueExtension
    get() {
        ...
    }
```
this will simplify this logic.

Secondly, consider having more role-specific validators. i.e. max value validator, min value validator. So in your validator all you have to do is to apply all these validators to a question/answer and then aggregate the results.

datacapture/src/main/java/com/google/android/fhir/datacapture/views/QuestionnaireItemEditTextViewHolderFactory.kt
```diff
@@ -63,4 +64,10 @@ abstract class QuestionnaireItemEditTextViewHolderDelegate(
      * answer to the question (may be input by the user or previously recorded).
      */
     abstract fun getText(answer: QuestionnaireResponse.Item.Answer.Builder?): String
+
+    /**
+     * Validates the input data if there is any behavioral extension in the Questionnaire and
+     * renders the View accordingly
+     */
+    abstract fun validate(questionnaireItemViewItem: QuestionnaireItemViewItem, textInputLayout: TextInputLayout)
```
jingtang10, line 72: This shouldn't be abstract. I think this should be actually calling your validator (which callls sub validators such as min/max value validators).

datacapture/src/main/java/com/google/android/fhir/datacapture/validation/QuestionnaireItemIntegerValidator.kt
```diff
@@ -0,0 +1,62 @@
+package com.google.android.fhir.datacapture.validation
+
+import com.google.fhir.r4.core.Extension
+import com.google.fhir.r4.core.Questionnaire
+import com.google.fhir.r4.core.QuestionnaireResponse
+
+class QuestionnaireItemIntegerValidator : QuestionnaireItemValidator {
```
jingtang10, line 7: I'm not sure I agree with this abstraction. I think the logic of validation quickly becomes unmanageable across all data types...

Have you considered creating a validator for the max value extesions and just checks if the answer doesn't exceed the max value? and creating a min value validator that just checks the answer isn't less than the min value?

And then in your overall validator you just have this logic where you call all these validators, and aggregate the results by composing a validator result from all the error messages from the sub validators.

jingtang10 (changes requested):

datacapture/src/main/java/com/google/android/fhir/datacapture/validation/QuestionnaireItemIntegerValidator.kt
kunjan8794: My assumption here was, there will be different validators for different data types. So the interface QuestionnaireItemValidator can have the abstract validator that can be implemented by implementor classes of different data types such as integer, decimal, or even a date

datacapture/src/main/java/com/google/android/fhir/datacapture/views/QuestionnaireItemEditTextViewHolderFactory.kt
kunjan8794: @jingtang10  I believe we have 'getText' and 'getValue' implemented separately for the individual data types, so in that case, shouldn't the method 'validate' be implemented separately as well? Because maybe we will need to handle the validation logic of the answer as per the data type? 

Please point out if I am missing something major here

datacapture/src/main/java/com/google/android/fhir/datacapture/validation/QuestionnaireItemIntegerValidator.kt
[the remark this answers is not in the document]
kunjan8794: Sure, changing this

datacapture/src/main/java/com/google/android/fhir/datacapture/validation/QuestionnaireItemIntegerValidator.kt
```diff
@@ -0,0 +1,62 @@
...
+    companion object {}
```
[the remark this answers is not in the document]
kunjan8794: Done

datacapture/src/main/java/com/google/android/fhir/datacapture/validation/QuestionnaireItemIntegerValidator.kt
```diff
@@ -0,0 +1,62 @@
...
+
+    override fun validate(questionnaireItem: Questionnaire.Item, questionnaireResponseItemBuilder: QuestionnaireResponse.Item.Builder): ValidationResult {
+        if (questionnaireResponseItemBuilder.answerBuilderList.isNotEmpty()) {
+
```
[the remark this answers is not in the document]
kunjan8794: done

datacapture/src/main/java/com/google/android/fhir/datacapture/validation/QuestionnaireItemIntegerValidator.kt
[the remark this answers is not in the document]
kunjan8794: Got it, done

kunjan8794:
@jingtang10 please review the PR (hopefully finally :P)

jingtang10 (changes requested):

kunjan8794:
@jingtang10 Hi Jing, please review the recent code changes

datacapture/src/main/java/com/google/android/fhir/datacapture/validation/MaxValueValidator.kt
```diff
@@ -0,0 +1,33 @@
...
+    private fun maxValueIntegerValidator(extension: Extension, questionnaireResponseItemBuilder: QuestionnaireResponse.Item.Builder): QuestionnaireItemValidator.ValidationResult {
+        val response = questionnaireResponseItemBuilder.getAnswerBuilder(0).value
+        if (extension.value.hasInteger() && response.hasInteger()) {
+            val answer = questionnaireResponseItemBuilder.getAnswerBuilder(0).value.integer.value
+            if (answer > extension.value.integer.value) {
+                return QuestionnaireItemValidator.ValidationResult(false, validationMessageGenerator(extension))
+            }
+        }
+        return QuestionnaireItemValidator.ValidationResult(true, emptyList())
+    }
```
jingtang10, line 28: I think there needs to be a `when` block for the different types because we will be adding support for `Decimal`, `Date`, `DateTime` etc.

datacapture/src/main/java/com/google/android/fhir/datacapture/views/QuestionnaireItemEditTextViewHolderFactory.kt
```diff
@@ -46,9 +47,29 @@ abstract class QuestionnaireItemEditTextViewHolderDelegate(
...
+        var validationMessages: String? = null
+        validationResults.forEach {
+            if (!it.isValid) {
+                if (validationMessages == null)
+                    validationMessages = it.messages[0]
+                else
+                    validationMessages = validationMessages.plus(it.messages[0]).plus(messagesSeparator)
+                questionnaireItemViewItem.questionnaireResponseItemBuilder.clearAnswer()
+            }
+        }
```
jingtang10, line 66: I think you can use something like this: https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.collections/join-to-string.html

jingtang10 (changes requested):
Thanks @kunjan8794 ! Really good progress -- can you also add some tests please? Thanks!

kunjan8794:
@googlebot I consent.

jingtang10 (changes requested):
Thanks I think the interface is much better.

Please rebase/merge to pull in the latest change on master. You'll need to adapt to the new hapi structures lib too.

jingtang10:
@googlebot I consent.

datacapture/src/test/java/com/google/android/fhir/datacapture/validation/ConstraintValidatorTest.kt
```diff
@@ -0,0 +1,156 @@
...
+import org.hl7.fhir.r4.model.QuestionnaireResponse
+import org.hl7.fhir.r4.model.QuestionnaireResponse.QuestionnaireResponseItemAnswerComponent
+import org.junit.Test
+import org.junit.runner.RunWith
+import org.robolectric.RobolectricTestRunner
+import org.robolectric.annotation.Config
+
+@RunWith(RobolectricTestRunner::class)
+@Config(sdk = [Build.VERSION_CODES.P])
+class ConstraintValidatorTest {
```
jingtang10, line 33: in the interest of avoiding exponential permutations of tests, i think we can reorganise this file in the following way:

1. have a test class that just tests the `QuestionnaireResponseItemValidator`. what it needs to test is that it aggregates the result of child validators.
2. have a test class that tests the max validator
3. have a test class that tests the min validator

jingtang10 (changes requested):
Thanks @kunjan8794!

datacapture/src/main/java/com/google/android/fhir/datacapture/views/QuestionnaireItemEditTextViewHolderFactory.kt
```diff
@@ -46,9 +47,26 @@ internal abstract class QuestionnaireItemEditTextViewHolderDelegate(
...
+          questionnaireItemViewItem.questionnaireResponseItem
+        )
+      )
     }
   }
 
+  private fun applyValidationResults(
+    validationResults: List<QuestionnaireResponseItemValidator.ValidationResult>
+  ) {
+    val messagesSeparator = '\n'
```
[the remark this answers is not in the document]
kunjan8794: Inline, separated by a comma? Maybe? Or a space? @jingtang10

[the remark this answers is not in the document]
jingtang10: no i mean inline the variable `messageSeparator`. Also can you try using `System.getProperty("line.separator")` instead?

datacapture/src/main/java/com/google/android/fhir/datacapture/validation/MaxValueValidator.kt
```diff
@@ -0,0 +1,61 @@
...
+ * limitations under the License.
+ */
+
+package com.google.android.fhir.datacapture.validation
+
+import org.hl7.fhir.r4.model.Extension
+import org.hl7.fhir.r4.model.Questionnaire
+import org.hl7.fhir.r4.model.QuestionnaireResponse
+
+object MaxValueValidator : ConstraintValidator {
```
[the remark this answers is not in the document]
kunjan8794: @jingtang10 in that case can we just have one MinMaxValueValidator class that would validate both Min and Max values in case the QuestionnaireItem has any of the extensions?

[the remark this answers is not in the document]
jingtang10: I think you should have

`ValueConstraintValidator` which `MaxValueConstraintValidator` and `MinValueConstraintValidator` extend.

The `ValueConstraintValidator` would basically compare the answer with the given value using a predicate and generate the error message. The min and max implementations would overload that predicate (one of them would be `>` and the other `<`) and generate different error message.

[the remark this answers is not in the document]
kunjan8794: This is what should be the template for the validators, am I right?

class MinValueValidator : ValueConstraintValidator(), ConstraintValidator { }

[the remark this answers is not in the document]
kunjan8794: Hi @jingtang10 , I have created a class ValueConstraintValidator.kt & it has a companion method -->

fun valueConstraintValidator(extension: Extension, questionnaireResponseItemBuilder: QuestionnaireResponse.QuestionnaireResponseItemComponent, operator: String): Boolean

Here, there is comparison based on the operator being passed from the MaxValueValidator and MinValueValidator.

Will this work? I am committing this, in case you want to have a look.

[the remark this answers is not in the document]
jingtang10: Maybe something like this:

```
class ValueConstraintValidator(
  val url: String, 
  val predicate: (extension) -> Boolean,
  val messageGenerator: (extension) -> String
) : ConstraintValidator {
  override fun validate(...) {
    if (hasExtensionUrl(url)) {
      if (predicate(extension.value, questionnaireResponseItem.value)) {
        return (false, messageGenerator())
      } else {
        return (true, ...)
      }
    }
  } 
}

object MaxValueConstraintValidator = ValueConstraintValidator(
  ...,
  { a, b
    a > b
  },
  ...
)

...
```
?

jingtang10 (changes requested):

datacapture/src/main/java/com/google/android/fhir/datacapture/validation/MaxValueValidator.kt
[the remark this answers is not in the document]
kunjan8794: @jingtang10 i am just wondering how will we handle the different data types in the predicate function? We need to handle float & date as well.

datacapture/src/test/java/com/google/android/fhir/datacapture/validation/QuestionnaireResponseItemValidatorTest.kt
```diff
@@ -0,0 +1,101 @@
+package com.google.android.fhir.datacapture.validation
+
+import android.os.Build
+import com.google.common.truth.Truth
+import com.google.common.truth.Truth.*
```
s-ayush2903, line 5: @kunjan8794 Can you please remove this wildcard import, maybe it is the cause for spotless failure

kunjan8794: @s-ayush2903 thank you Ayush! That was simple :)

datacapture/src/main/java/com/google/android/fhir/datacapture/validation/MaxValueValidator.kt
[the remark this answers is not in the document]
kunjan8794: @jingtang10 can you please review the recent code push? I have committed based on your comments, but thats for the integer type for now

datacapture/src/test/java/com/google/android/fhir/datacapture/validation/QuestionnaireResponseItemValidatorTest.kt
s-ayush2903: Oh apologies for inconvenience :sweat_smile:

datacapture/src/main/java/com/google/android/fhir/datacapture/validation/ValueConstraintValidator.kt
```diff
@@ -0,0 +1,62 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.google.android.fhir.datacapture.validation
+
+import org.hl7.fhir.r4.model.Questionnaire
+import org.hl7.fhir.r4.model.QuestionnaireResponse
+
+open class ValueConstraintValidator(
```
vincent-karuri, line 22: Is this class intended to only validate Integer constraints? In that case, I wonder if `IntegerValueConstraintValidator` is a more suitable name?

datacapture/src/main/java/com/google/android/fhir/datacapture/validation/MaxValueValidator.kt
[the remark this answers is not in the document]
jingtang10: So let's keep this PR to integer type. For the future, we should create some utilities similar to in this class: https://github.com/google/android-fhir/blob/master/datacapture/src/main/java/com/google/android/fhir/datacapture/MoreQuestionnaireItemTypes.kt.

For now please do the following

- [ ] Create an issue to add support for other types
- [ ] Assign yourself
- [ ] Add a comment to link to that issue in the code

datacapture/src/main/java/com/google/android/fhir/datacapture/validation/ConstraintValidator.kt
```diff
@@ -0,0 +1,37 @@
...
+ */
+
+package com.google.android.fhir.datacapture.validation
+
+import org.hl7.fhir.r4.model.Questionnaire
+import org.hl7.fhir.r4.model.QuestionnaireResponse
+
+interface ConstraintValidator {
+  /**
+   * Validates the \`answer\`(s) in [questionnaireResponseItem] satify any constraints of the
```
jingtang10, line 24: fix the escape:

```suggestion
   * Validates the `answer`(s) in [questionnaireResponseItem] satify any constraints of the
```

datacapture/src/main/java/com/google/android/fhir/datacapture/validation/QuestionnaireResponseItemValidator.kt
```diff
@@ -0,0 +1,55 @@
...
+object QuestionnaireResponseItemValidator {
+
+  private val validators = mutableListOf(MaxValueValidator, MinValueValidator)
+
+  /**
+   * Validates [questionnaireResponseItemBuilder] contains valid answer(s) to [questionnaireItem].
+   */
+  fun validate(
+    questionnaireItem: Questionnaire.QuestionnaireItemComponent,
+    questionnaireResponseItemBuilder: QuestionnaireResponse.QuestionnaireResponseItemComponent
```
jingtang10, line 31:
```suggestion
    questionnaireResponseItem: QuestionnaireResponse.QuestionnaireResponseItemComponent
```

datacapture/src/main/java/com/google/android/fhir/datacapture/validation/ConstraintValidator.kt
```diff
@@ -0,0 +1,37 @@
...
+   * [questionnaireItem] according to the [structured data capture implementation guide]
+   * (http://build.fhir.org/ig/HL7/sdc/behavior.html). This does not validate the
+   * `questionnaireResponseItem` and its child items are structurally consistent with the
+   * `questionnaireItem` and its child items.
```
jingtang10, line 28:
```suggestion
   * [questionnaireResponseItem] and its child items are structurally consistent with the
   * [questionnaireItem] and its child items.
```

datacapture/src/main/java/com/google/android/fhir/datacapture/validation/ValueConstraintValidator.kt
jingtang10: @kunjan8794 please add a todo and link to an issue here. But I don't think we need to rename this class if we have a follow-up PR after this PR.

datacapture/src/main/java/com/google/android/fhir/datacapture/validation/ValueConstraintValidator.kt
```diff
@@ -0,0 +1,62 @@
...
+            )
+          }
+        }
+      }
+    }
+    return ConstraintValidator.ConstraintValidationResult(true, null)
+  }
+}
+
+object MaxValueValidator :
```
jingtang10, line 50: this is great, thanks :)

let's move these out of this file -- we'll probably end up with a bunch of these validators, they probably deserve to be in their own files.

datacapture/src/main/java/com/google/android/fhir/datacapture/views/QuestionnaireItemEditTextViewHolderFactory.kt
```diff
@@ -48,9 +49,25 @@ internal abstract class QuestionnaireItemEditTextViewHolderDelegate(
...
+        )
+      )
     }
   }
 
+  private fun applyValidationResult(
+    validationResult: QuestionnaireResponseItemValidator.ValidationResult
+  ) {
+    val validationMessage =
+      validationResult.validationMessages.filter { it != null }.joinToString {
```
jingtang10, line 65:
```suggestion
      validationResult.validationMessages.filterNotNull().joinToString {
```

actually why would you end up with `null`s in the list?

datacapture/src/test/java/com/google/android/fhir/datacapture/validation/ConstraintValidatorTest.kt
jingtang10: still to do in this PR:
- [ ] have a test class that just tests the QuestionnaireResponseItemValidator. what it needs to test is that it aggregates the result of child validators.
- [ ] have a test class that tests the max validator
- [ ] have a test class that tests the min validator

datacapture/src/test/java/com/google/android/fhir/datacapture/validation/QuestionnaireResponseItemValidatorTest.kt
```diff
@@ -0,0 +1,92 @@
...
+    val extensionUrlMinValue = "http://hl7.org/fhir/StructureDefinition/minValue"
+
+    /** Scenario 1 - answerValue is greater than maxValue */
+    var minValue = 250
+    var maxValue = 200000
+    var answerValue = 200001
+    val validateMaxValueViolation =
+      validateAScenario(answerValue, maxValue, minValue, extensionUrlMaxValue, extensionUrlMinValue)
+    assertThat(validateMaxValueViolation.isValid).isFalse()
+    assertThat(validateMaxValueViolation.validationMessages.size == 1).isTrue()
```
jingtang10, line 46: in general in unit tests we like our test cases to be as isolated as possible. you should be able to read the test failures and immediately get a sense of why the test is failing.

if you mix all these different cases in the same `@Test`, when you get a test failure you still need to come back to the code and debug to figure out what is failing.

so separate these into separate cases and give them descriptive names.

datacapture/src/main/java/com/google/android/fhir/datacapture/validation/QuestionnaireResponseItemValidator.kt
```diff
@@ -0,0 +1,55 @@
...
+      return ValidationResult(true, mutableListOf())
+    }
+    val validationResults = mutableListOf<ConstraintValidator.ConstraintValidationResult>()
+    validators.forEach {
+      validationResults.add(it.validate(questionnaireItem, questionnaireResponseItemBuilder))
+    }
+    return packConstraintValidationResults(validationResults)
+  }
+
+  data class ValidationResult(var isValid: Boolean, val validationMessages: MutableList<String?>)
```
jingtang10, line 43:
```suggestion
  data class ValidationResult(var isValid: Boolean, val validationMessages: List<String?>)
```

datacapture/src/main/java/com/google/android/fhir/datacapture/validation/QuestionnaireResponseItemValidator.kt
```diff
@@ -0,0 +1,55 @@
...
+  private fun packConstraintValidationResults(
+    validationResults: List<ConstraintValidator.ConstraintValidationResult>
+  ): ValidationResult {
+    val validationResult = ValidationResult(true, mutableListOf())
+    validationResults.forEach {
+      validationResult.isValid = validationResult.isValid && it.isValid
+      if (!it.message.isNullOrEmpty()) validationResult.validationMessages.add(it.message)
+    }
+    return validationResult
+  }
```
jingtang10, line 54:
```suggestion
= ValidationResult(
      validationResults.all { it.isValid },
      validationResults.mapNotNull { it.message }.toList()
    )
```

jingtang10 (approved):
Thanks - I think the interface is much better and extensible now.

jingtang10 (changes requested):
unapprove. sorry misclicked. please address the comments.

datacapture/src/main/java/com/google/android/fhir/datacapture/validation/ConstraintValidator.kt
```diff
@@ -0,0 +1,37 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.google.android.fhir.datacapture.validation
+
+import org.hl7.fhir.r4.model.Questionnaire
+import org.hl7.fhir.r4.model.QuestionnaireResponse
+
+interface ConstraintValidator {
```
jingtang10, line 22: internal

datacapture/src/main/java/com/google/android/fhir/datacapture/validation/QuestionnaireResponseItemValidator.kt
```diff
@@ -0,0 +1,55 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.google.android.fhir.datacapture.validation
+
+import org.hl7.fhir.r4.model.Questionnaire
+import org.hl7.fhir.r4.model.QuestionnaireResponse
+
+object QuestionnaireResponseItemValidator {
```
jingtang10, line 22: internal

datacapture/src/main/java/com/google/android/fhir/datacapture/validation/QuestionnaireResponseItemValidator.kt
jingtang10, line 43: move this out of the class and keep it in the same file. and make it internal.

datacapture/src/main/java/com/google/android/fhir/datacapture/validation/ValueConstraintValidator.kt
jingtang10, line 22: internal

datacapture/src/main/java/com/google/android/fhir/datacapture/validation/ValueConstraintValidator.kt
jingtang10, line 50: internal

datacapture/src/main/java/com/google/android/fhir/datacapture/validation/ValueConstraintValidator.kt
```diff
@@ -0,0 +1,62 @@
...
+  ValueConstraintValidator(
+    url = "http://hl7.org/fhir/StructureDefinition/maxValue",
+    predicate = { a: Int, b: Int -> a > b },
+    { allowedValue: String -> "Maximum value allowed is:$allowedValue" }
+  )
+
+object MinValueValidator :
```
jingtang10, line 57: internal

jingtang10 (changes requested):

datacapture/src/main/java/com/google/android/fhir/datacapture/views/QuestionnaireItemEditTextViewHolderFactory.kt
kunjan8794: @jingtang10 I think yeah, now we will not end up with null values here.

datacapture/src/main/java/com/google/android/fhir/datacapture/validation/QuestionnaireResponseItemValidator.kt
kunjan8794: @jingtang10 I am packing the not null values of ConstraintValidationResult into ValidationResult in the method "packConstraintValidationResults" and hence we needed a mutable list. Or else it wont allow me to use the method "add"

datacapture/src/main/java/com/google/android/fhir/datacapture/validation/ConstraintValidator.kt
kunjan8794: @jingtang10 ConstraintValidator is being implemented by ValueConstraintValidator and hence it gives me an error if I make ConstraintValidator internal.

datacapture/src/test/java/com/google/android/fhir/datacapture/validation/QuestionnaireResponseItemValidatorTest.kt
kunjan8794: @jingtang10 done

datacapture/src/test/java/com/google/android/fhir/datacapture/validation/QuestionnaireResponseItemValidatorTest.kt
```diff
@@ -0,0 +1,113 @@
...
+    extensionMinValue.setValue(IntegerType(minValue))
+    val extensions =
+      mutableListOf<Extension>().apply {
+        add(extensionMaxValue)
+        add(extensionMinValue)
+      }
+    val validateAggregationFromChildValidators = applyScenario(extensions, answerValue)
+    assertThat(validateAggregationFromChildValidators.isValid).isFalse()
+    assertThat(validateAggregationFromChildValidators.validationMessages.size == 1).isTrue()
+  }
```
jingtang10, line 78: these two are actuall the same case. what you need is a case where both child validators fail. so you'll need a min validator with a high value e.g. 200 and a max validator with a low value e.g. 100 and an answer with a value in between e.g. 150 so both will fail.

datacapture/src/test/java/com/google/android/fhir/datacapture/validation/MaxValueConstraintValidatorTest.kt
```diff
@@ -0,0 +1,76 @@
...
+    val maxValue = 200000
+    val answerValue = 199999
+    val extensionUrl = "http://hl7.org/fhir/StructureDefinition/maxValue"
+    val maxValueValidatorScenarioTwo =
+      maxValueValidatorScenarioTester(answerValue, maxValue, extensionUrl)
+    Truth.assertThat(maxValueValidatorScenarioTwo.isValid).isTrue()
+    Truth.assertThat(maxValueValidatorScenarioTwo.message.isNullOrBlank()).isTrue()
+  }
+
+  private fun maxValueValidatorScenarioTester(
```
jingtang10, line 59: i think if you write idiomatic kotlin by using apply (see comment below) you don't really need this. just inline this function.

datacapture/src/test/java/com/google/android/fhir/datacapture/validation/MaxValueConstraintValidatorTest.kt
```diff
@@ -0,0 +1,76 @@
+/*
```
jingtang10, line 1: most comments in this file should be applied to the other two test files also

datacapture/src/main/java/com/google/android/fhir/datacapture/validation/ValueConstraintValidator.kt
```diff
@@ -0,0 +1,48 @@
...
+ */
+
+package com.google.android.fhir.datacapture.validation
+
+import org.hl7.fhir.r4.model.Questionnaire
+import org.hl7.fhir.r4.model.QuestionnaireResponse
+
+internal open class ValueConstraintValidator(
+  val url: String,
+  val predicate: (Int, Int) -> Boolean,
```
jingtang10, line 24: hm sorry i missed this in the last review - I think the predicate should be type agnostic. If you keep it this way, in the next PR you'll need to write a separate validator for each data type for max and min. that's too much.

so you should pass in the Type (which is the value in the extension) and the QuestionnaireResponseItemAnswerComponent which is the value in the answer. and in the predicate in the max and min validators you should have a when statement that compares each type.

```suggestion
  val predicate: (Type, QuestionnaireResponseItemAnswerComponent) -> Boolean,
```

datacapture/src/main/java/com/google/android/fhir/datacapture/validation/ValueConstraintValidator.kt
```diff
@@ -0,0 +1,48 @@
...
+  val messageGenerator: (allowedValue: String) -> String
```
jingtang10, line 25:
```suggestion
  val messageGenerator: (Type, QuestionnaireResponseItemAnswerComponent) -> String
```

jingtang10 (changes requested):
Thanks @kunjan8794!

datacapture/src/main/java/com/google/android/fhir/datacapture/validation/QuestionnaireResponseItemValidator.kt
```diff
@@ -0,0 +1,50 @@
...
+  /** Validates [questionnaireResponseItem] contains valid answer(s) to [questionnaireItem]. */
+  fun validate(
+    questionnaireItem: Questionnaire.QuestionnaireItemComponent,
+    questionnaireResponseItem: QuestionnaireResponse.QuestionnaireResponseItemComponent
+  ): ValidationResult {
+    if (questionnaireResponseItem.answer.isEmpty()) {
+      return ValidationResult(true, mutableListOf())
+    }
+    val validationResults = mutableListOf<ConstraintValidator.ConstraintValidationResult>()
+    validators.forEach {
```
vincent-karuri, line 35: @jingtang10 quick question, if I read this correctly, the logic here will always iterate over all possible validators on a field even when they're not specified for that field? In that case, I wonder if it would be better to only get the validation extensions defined by the field upfront instead of relying on the list?

jingtang10: yeah definitely. a map would be more efficient. but that can be a follow-up.

datacapture/src/main/java/com/google/android/fhir/datacapture/validation/ValueConstraintValidator.kt
kunjan8794: @jingtang10 Using "Extension" instead of "Type"

datacapture/src/main/java/com/google/android/fhir/datacapture/validation/ValueConstraintValidator.kt
kunjan8794: @jingtang10 Using "Extension" instead of "Type"

datacapture/src/test/java/com/google/android/fhir/datacapture/validation/MaxValueConstraintValidatorTest.kt
```diff
@@ -0,0 +1,86 @@
...
+        addAnswer(
+          QuestionnaireResponse.QuestionnaireResponseItemAnswerComponent().apply {
+            value = IntegerType(200001)
+          }
+        )
+      }
+    val validationResult =
+      MaxValueConstraintValidator.validate(questionnaireItem, questionnaireResponseItem)
+    Truth.assertThat(validationResult.isValid).isFalse()
+    Truth.assertThat(validationResult.message.equals("Maximum value allowed is:200000"))
```
jingtang10, line 57:
```suggestion
    Truth.assertThat(validationResult.message).isEuqalTo("Maximum value allowed is:200000")
```

datacapture/src/main/java/com/google/android/fhir/datacapture/validation/MaxValueConstraintValidator.kt
```diff
@@ -0,0 +1,24 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.google.android.fhir.datacapture.validation
+
+internal object MaxValueConstraintValidator :
+  ValueConstraintValidator(
+    url = "http://hl7.org/fhir/StructureDefinition/maxValue",
```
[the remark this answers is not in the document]
jingtang10: reminder to do this

datacapture/src/main/java/com/google/android/fhir/datacapture/validation/MinValueConstraintValidator.kt
```diff
@@ -0,0 +1,24 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.google.android.fhir.datacapture.validation
+
+internal object MinValueConstraintValidator :
+  ValueConstraintValidator(
+    url = "http://hl7.org/fhir/StructureDefinition/minValue",
```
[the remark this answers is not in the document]
jingtang10: reminder to do this

datacapture/src/test/java/com/google/android/fhir/datacapture/validation/MaxValueConstraintValidatorTest.kt
```diff
@@ -0,0 +1,86 @@
...
+import org.junit.Test
+import org.junit.runner.RunWith
+import org.robolectric.RobolectricTestRunner
+import org.robolectric.annotation.Config
+
+@RunWith(RobolectricTestRunner::class)
+@Config(sdk = [Build.VERSION_CODES.P])
+class MaxValueConstraintValidatorTest {
+
+  // Scenario 1 - answerValue is greater than maxValue
```
jingtang10, line 34: remove comments like this

datacapture/src/test/java/com/google/android/fhir/datacapture/validation/MaxValueConstraintValidatorTest.kt
```diff
@@ -0,0 +1,86 @@
...
+        )
+      }
+    val validationResult =
+      MaxValueConstraintValidator.validate(questionnaireItem, questionnaireResponseItem)
+    Truth.assertThat(validationResult.isValid).isFalse()
+    Truth.assertThat(validationResult.message.equals("Maximum value allowed is:200000"))
+      .isTrue()
+  }
+
+  // Scenario 2 - answerValue is less than maxValue
```
jingtang10, line 61: remove

datacapture/src/test/java/com/google/android/fhir/datacapture/validation/MinValueConstraintValidatorTest.kt
```diff
@@ -0,0 +1,84 @@
...
+import org.junit.Test
+import org.junit.runner.RunWith
+import org.robolectric.RobolectricTestRunner
+import org.robolectric.annotation.Config
+
+@RunWith(RobolectricTestRunner::class)
+@Config(sdk = [Build.VERSION_CODES.P])
+class MinValueConstraintValidatorTest {
+
+  //Scenario 1 - answerValue is less than minValue
```
jingtang10, line 35: remove

datacapture/src/test/java/com/google/android/fhir/datacapture/validation/MinValueConstraintValidatorTest.kt
```diff
@@ -0,0 +1,84 @@
...
+          }
+        )
+      }
+    val validationResult =
+      MinValueConstraintValidator.validate(questionnaireItem, questionnaireResponseItem)
+    assertThat(validationResult.isValid).isFalse()
+    assertThat(validationResult.message.equals("Minimum value allowed is:10")).isTrue()
+  }
+
+  // Scenario 2 - answerValue is greater than minValue
```
jingtang10, line 60: remove

datacapture/src/test/java/com/google/android/fhir/datacapture/validation/MaxValueConstraintValidatorTest.kt
```diff
@@ -0,0 +1,86 @@
...
+      QuestionnaireResponse.QuestionnaireResponseItemComponent().apply {
+        addAnswer(
+          QuestionnaireResponse.QuestionnaireResponseItemAnswerComponent().apply {
+            value = IntegerType(200001)
+          }
+        )
+      }
+    val validationResult =
+      MaxValueConstraintValidator.validate(questionnaireItem, questionnaireResponseItem)
+    Truth.assertThat(validationResult.isValid).isFalse()
```
jingtang10, line 56: import assertThat

datacapture/src/test/java/com/google/android/fhir/datacapture/validation/QuestionnaireResponseItemValidatorTest.kt
```diff
@@ -0,0 +1,95 @@
...
+    val extensionUrlMinValue = "http://hl7.org/fhir/StructureDefinition/minValue"
+    val extensionMaxValue = Extension()
+    val extensionMinValue = Extension()
+    val minValue = 250
+    val maxValue = 200000
+    val answerValue = 251
+    extensionMaxValue.url = extensionUrlMaxValue
+    extensionMaxValue.setValue(IntegerType(maxValue))
+    extensionMinValue.url = extensionUrlMinValue
+    extensionMinValue.setValue(IntegerType(minValue))
```
jingtang10, line 46: do the same thing as you did in the other test files

jingtang10 (changes requested):

datacapture/src/test/java/com/google/android/fhir/datacapture/validation/QuestionnaireResponseItemValidatorTest.kt
```diff
@@ -0,0 +1,95 @@
...
+        addAnswer(
+          QuestionnaireResponse.QuestionnaireResponseItemAnswerComponent().apply {
+            value = IntegerType(550)
+          }
+        )
+      }
+    val validateAggregationFromChildValidators =
+      QuestionnaireResponseItemValidator.validate(questionnaireItem, questionnaireResponseItem)
+    assertThat(validateAggregationFromChildValidators.isValid).isFalse()
+    assertThat(validateAggregationFromChildValidators.validationMessages.size == 2).isTrue()
```
jingtang10, line 93:
```suggestion
    assertThat(validateAggregationFromChildValidators.validationMessages.size).isEqualTo(2)
```

datacapture/src/test/java/com/google/android/fhir/datacapture/validation/QuestionnaireResponseItemValidatorTest.kt
```diff
@@ -0,0 +1,95 @@
...
+        addAnswer(
+          QuestionnaireResponse.QuestionnaireResponseItemAnswerComponent().apply {
+            value = IntegerType(275)
+          }
+        )
+      }
+    val validateAggregationFromChildValidators =
+      QuestionnaireResponseItemValidator.validate(questionnaireItem, questionnaireResponseItem)
+    assertThat(validateAggregationFromChildValidators.isValid).isTrue()
+    assertThat(validateAggregationFromChildValidators.validationMessages.isEmpty()).isTrue()
```
jingtang10, line 62:
```suggestion
    assertThat(validateAggregationFromChildValidators.validationMessages).isEmpty()
```

datacapture/src/test/java/com/google/android/fhir/datacapture/validation/MaxValueConstraintValidatorTest.kt
```diff
@@ -0,0 +1,83 @@
...
+      }
+    val questionnaireResponseItem =
+      QuestionnaireResponse.QuestionnaireResponseItemComponent().apply {
+        addAnswer(
+          QuestionnaireResponse.QuestionnaireResponseItemAnswerComponent().apply {
+            value = IntegerType(200001)
+          }
+        )
+      }
+    val validationResult =
```
jingtang10, line 53: new line

let's try to follow this structure:

```
test() {
  set up 
  set up 
  set up 

  the action to be tested

  assertions
  assertions
  assertions
}
```

datacapture/src/test/java/com/google/android/fhir/datacapture/validation/MaxValueConstraintValidatorTest.kt
```diff
@@ -0,0 +1,83 @@
...
+      MaxValueConstraintValidator.validate(questionnaireItem, questionnaireResponseItem)
+    assertThat(validationResult.isValid).isFalse()
```
jingtang10, line 55: new line

datacapture/src/test/java/com/google/android/fhir/datacapture/validation/MaxValueConstraintValidatorTest.kt
```diff
@@ -0,0 +1,83 @@
...
+      }
+    val questionnaireResponseItem =
+      QuestionnaireResponse.QuestionnaireResponseItemComponent().apply {
+        addAnswer(
+          QuestionnaireResponse.QuestionnaireResponseItemAnswerComponent().apply {
+            value = IntegerType(501)
+          }
+        )
+      }
+    val validationResult =
```
jingtang10, line 78: new line

datacapture/src/test/java/com/google/android/fhir/datacapture/validation/MaxValueConstraintValidatorTest.kt
```diff
@@ -0,0 +1,83 @@
...
+      MaxValueConstraintValidator.validate(questionnaireItem, questionnaireResponseItem)
+    assertThat(validationResult.isValid).isTrue()
```
jingtang10, line 80: new line

and same with the other 2 files.

jingtang10 (approved):
Thanks @kunjan8794! 🎉

Please address the final few nits before merging.

datacapture/src/test/java/com/google/android/fhir/datacapture/validation/MinValueConstraintValidatorTest.kt
```diff
@@ -0,0 +1,80 @@
...
+    val questionnaireResponseItem =
+      QuestionnaireResponse.QuestionnaireResponseItemComponent().apply {
+        addAnswer(QuestionnaireResponseItemAnswerComponent().apply { value = IntegerType(501) })
+      }
+
+    val validationResult =
+      MinValueConstraintValidator.validate(questionnaireItem, questionnaireResponseItem)
+
+    assertThat(validationResult.isValid).isTrue()
+    assertThat(validationResult.message.isNullOrBlank()).isTrue()
```
jingtang10, line 78: does this work?
```suggestion
    assertThat(validationResult.message)isEmpty()
```

datacapture/src/test/java/com/google/android/fhir/datacapture/validation/QuestionnaireResponseItemValidatorTest.kt
```diff
@@ -0,0 +1,99 @@
...
+    val questionnaireResponseItem =
+      QuestionnaireResponse.QuestionnaireResponseItemComponent().apply {
+        addAnswer(
+          QuestionnaireResponse.QuestionnaireResponseItemAnswerComponent().apply {
+            value = IntegerType(550)
+          }
+        )
+      }
+
+    val validateAggregationFromChildValidators =
```
jingtang10, line 93:
```suggestion
    val validationResult =
```

datacapture/src/test/java/com/google/android/fhir/datacapture/validation/QuestionnaireResponseItemValidatorTest.kt
```diff
@@ -0,0 +1,99 @@
...
+    val questionnaireResponseItem =
+      QuestionnaireResponse.QuestionnaireResponseItemComponent().apply {
+        addAnswer(
+          QuestionnaireResponse.QuestionnaireResponseItemAnswerComponent().apply {
+            value = IntegerType(275)
+          }
+        )
+      }
+
+    val validateAggregationFromChildValidators =
```
jingtang10, line 60:
```suggestion
    val validationResult =
```

PR #231, opened 2021-02-09
Fix QuestionnaireItemDropDownViewHolderFactory visibility

deepankarb:
'public' subclass `QuestionnaireItemDropDownViewHolderFactory` exposes its 'internal' supertype `QuestionnaireItemViewHolderFactory`

We make it `internal` to hide the supertype.


jingtang10 (approved):
Thanks @deepankarb

PR #232, opened 2021-02-09
Add CI badge

deepankarb:
Added CI badge per https://docs.github.com/en/actions/managing-workflow-runs/adding-a-workflow-status-badge.


jingtang10 (approved):
👍

Issue #236, opened 2021-02-11
Support enablement operator EQUALS and NOT_EQUALS


jingtang10:
https://www.hl7.org/fhir/valueset-questionnaire-enable-operator.html

this code change needs to be in https://github.com/google/android-fhir/blob/master/datacapture/src/main/java/com/google/android/fhir/datacapture/enablement/EnablementEvaluator.kt

jingtang10:
this is currently causing a crash in the covid questionnaire

PR #239, opened 2021-02-14
Add enablement operators EQUALS and NOT_EQUAL_TO

pld:
Closes https://github.com/google/android-fhir/issues/236

Based on [Valueset-questionnaire-enable-operator](https://www.hl7.org/fhir/valueset-questionnaire-enable-operator.html) `EQUALS` is true if any answers contain the `enableWhen` answer. The definition of Not Equals is oddly worded, coded as `NOT_EQUAL_TO` is the negation of `EQUALS`.


datacapture/src/main/java/com/google/android/fhir/datacapture/enablement/EnablementEvaluator.kt
```diff
@@ -105,7 +105,24 @@ internal object EnablementEvaluator {
...
+     * Returns whether the list of `Item` answer values contains the answer in `enableWhen`.
+     *
+     * @param responseItem contains the list of answers to look within.
+     * @param enableWhen the answer to look for.
+     */
+    private fun responseItemContainsAnswer(
+        responseItem: QuestionnaireResponse.Item,
+        enableWhen: Questionnaire.Item.EnableWhen
+    ) =
+        responseItem.answerList.map { it.value.toByteString() }.contains(
```
pld, line 126: Is there a better way to equality? `values` with the same content are not equal

datacapture/src/main/java/com/google/android/fhir/datacapture/enablement/EnablementEvaluator.kt
```diff
@@ -105,7 +105,24 @@ internal object EnablementEvaluator {
         return when (val operator = enableWhen.operator.value) {
             QuestionnaireItemOperatorCode.Value.EXISTS ->
                 (responseItem.answerCount > 0) == enableWhen.answer.boolean.value
+            QuestionnaireItemOperatorCode.Value.EQUALS ->
+                responseItemContainsAnswer(responseItem, enableWhen)
+            QuestionnaireItemOperatorCode.Value.NOT_EQUAL_TO ->
```
pld, line 110: There was no `NOT_EQUALS`  enum

datacapture/src/main/java/com/google/android/fhir/datacapture/enablement/EnablementEvaluator.kt
```diff
@@ -105,7 +105,24 @@ internal object EnablementEvaluator {
         return when (val operator = enableWhen.operator.value) {
             QuestionnaireItemOperatorCode.Value.EXISTS ->
                 (responseItem.answerCount > 0) == enableWhen.answer.boolean.value
+            QuestionnaireItemOperatorCode.Value.EQUALS ->
+                responseItemContainsAnswer(responseItem, enableWhen)
```
jingtang10, line 109: Looking at rest of the opeartors: https://www.hl7.org/fhir/valueset-questionnaire-enable-operator.html it probably would make a lot of sense to provide some kind of predicate function here.

so you'd have in here:
```
responseItem.contains(enableWhen.operator.toPredicate())
```
and then have two extension functions.

the first extension function converts an enable when operator to a predicate
```
private fun Questionnaire.Item.EnableWhen.OperatorCode.toPredicate(): (QuestionnaireResponse.Item.Answer) -> Boolean {
    ...
}
```

and the second extension function takes that predicate to figure out if any answer in the answer list in the response item satisfies that predicate:
```
private fun QuestionnaireResponse.Item.contains(predicate:(QuestionnaireResponse.Item.Answer) -> Boolean) {
    ...
}
```
I think as a result this `when` block will be moved to the `toPredicate` extension function, and the code in this function will be simplified.

datacapture/src/main/java/com/google/android/fhir/datacapture/enablement/EnablementEvaluator.kt
jingtang10: hm I think this doesn't equal because they're different classes (`Questionnaire.Item.EnableWhen.AnswerX` and `QuestionnaireResponse.Item.Answer.ValueX`)... The right way of doing this probably write a utility function to compare each data type (since we don't actually know the type here). I'm happy for that to be a follow-up (or just create an issue)

jingtang10 (changes requested):
Thanks Peter for your contribution 👍  just a couple of high-level comments. Haven't looked through the tests yet.

datacapture/src/main/java/com/google/android/fhir/datacapture/enablement/EnablementEvaluator.kt
```diff
@@ -102,10 +102,39 @@ internal object EnablementEvaluator {
     ): Boolean {
         val responseItem =
             questionnaireResponseItemRetriever(enableWhen.question.value) ?: return true
-        return when (val operator = enableWhen.operator.value) {
-            QuestionnaireItemOperatorCode.Value.EXISTS ->
-                (responseItem.answerCount > 0) == enableWhen.answer.boolean.value
-            else -> throw NotImplementedError("Enable when operator $operator is not implemented.")
-        }
+        return if (QuestionnaireItemOperatorCode.Value.EXISTS == enableWhen.operator.value)
+            (responseItem.answerCount > 0) == enableWhen.answer.boolean.value else
```
pld, line 106: this operator seems like a special case

datacapture/src/main/java/com/google/android/fhir/datacapture/enablement/EnablementEvaluator.kt
```diff
@@ -102,10 +102,39 @@ internal object EnablementEvaluator {
...
+            responseItem.contains(enableWhen.toPredicate())
+    }
+}
+
+/**
+ * Returns a predicate based on the `EnableWhen` `operator` and `Answer` value.
+ */
+private fun Questionnaire.Item.EnableWhen.toPredicate(): (QuestionnaireResponse.Item.Answer)
```
pld, line 114: because the predicate depends on the value of the EnableWhen.Answer, I wrote this extending EnableWhen instead of EnableWhen.operator. This could take the answer as an additional argument, hmm maybe there's a way to clean this up

datacapture/src/main/java/com/google/android/fhir/datacapture/enablement/EnablementEvaluator.kt
```diff
@@ -102,10 +102,39 @@ internal object EnablementEvaluator {
...
+-> Boolean {
+    when (val operator = this.operator.value) {
+        QuestionnaireItemOperatorCode.Value.EQUALS ->
+            return {
+                this.answer.toByteString() == it.value.toByteString()
```
pld, line 119: Kinda suspicious, I'm sure I'm missing a better way with the similarity here to line 123.

I looked a bit at how to write the type aware comparisons, having this return only the operator could help to remove the need for the answer here (but then the answer would need to be passed to contains I think) and the type aware comparisons could go in contains, using the operator.

Or, and probably better,  have a helper here that takes the operator and the answer, then runs the type aware comparisons

datacapture/src/main/java/com/google/android/fhir/datacapture/ValueForType.kt
```diff
@@ -0,0 +1,52 @@
+package com.google.android.fhir.datacapture
```
pld, line 1: is there better name for this file? not sure how best to name structure in kotlin projects

datacapture/src/test/java/com/google/android/fhir/datacapture/enablement/EnablementEvaluatorTest.kt
```diff
@@ -342,4 +342,520 @@ class EnablementEvaluatorTest {
...
+                        .build()
+                } else {
+                    null
+                }
+            }
+        ).isFalse()
+    }
+
+    @Test
+    fun evaluate_expectsAnswer_answerEqualOne_shouldReturnTrue() {
```
jingtang10, line 419: the way i think about these test cases is that: we'll cover each operator in positive and negative cases, and then we'll have some tests that cover cases with multiple answers regardless of operators. having these cases is fine, but i kinda feel they'll really make the number of test cases explode without improving test coverage (since all cases should already have been hit).

as an example: we also have the ANY / OR enable when behaviors, does it mean the number of tests cases will x2? I don't think that'd be manageable.

jingtang10 (changes requested):
Thanks Peter!

datacapture/src/main/java/com/google/android/fhir/datacapture/MoreQuestionnaireItemTypes.kt
```diff
@@ -0,0 +1,67 @@
...
+
+import com.google.fhir.r4.core.Questionnaire
+import com.google.fhir.r4.core.QuestionnaireItemTypeCode
+import com.google.fhir.r4.core.QuestionnaireResponse
+import com.google.fhir.shaded.protobuf.Message
+
+/**
+ * Returns the value of the [Questionnaire.Item.EnableWhen.AnswerX] for the [type].
+ *
+ * Used to retrieve the value to set the field in the extracted FHIR resource.
```
jingtang10, line 27: this line can be removed since it's now a utility function used for various things.

datacapture/src/main/java/com/google/android/fhir/datacapture/MoreQuestionnaireItemTypes.kt
```diff
@@ -0,0 +1,67 @@
...
+    QuestionnaireItemTypeCode.Value.TIME -> this.time
+    QuestionnaireItemTypeCode.Value.STRING, QuestionnaireItemTypeCode.Value.TEXT ->
+        this.stringValue
+    else -> throw IllegalArgumentException("Unsupported value type $value")
+}
+
+/**
+ * Returns the value of the [QuestionnaireResponse.Item.Answer] for the [type].
+ *
+ * Used to retrieve the value to set the field in the extracted FHIR resource.
```
jingtang10, line 49: and this.

datacapture/src/main/java/com/google/android/fhir/datacapture/enablement/EnablementEvaluator.kt
jingtang10: hm. yeah. damn. 😛 

nit: can you please just reformat this a little bit

```
return if (...) {
    ...
}
else {
    ...
}

File: datacapture/src/main/java/com/google/android/fhir/datacapture/enablement/EnablementEvaluator.kt [hunk 11]
```diff
@@ -97,14 +101,49 @@ internal object EnablementEvaluator {
...
+/**
+ * Returns a predicate based on the `EnableWhen` `operator` and `Answer` value.
+ *
+ * @param type used to get value based on [Questionnaire.Item.TypeCode].
+ */
+private fun enableWhenTypeToPredicate(
+    enableWhen: Questionnaire.Item.EnableWhen,
+    type: Questionnaire.Item.TypeCode
+):
+        (QuestionnaireResponse.Item.Answer) -> Boolean {
```
jingtang10, line 138: nit: this can be on the previous line

datacapture/src/main/java/com/google/android/fhir/datacapture/enablement/EnablementEvaluator.kt
```diff
@@ -97,14 +101,49 @@ internal object EnablementEvaluator {
...
+    val enableWhenAnswerValue = enableWhen.answer.getValueForType(type)
+    return {
+        val answerValue = it.getValueForType(type)
+        when (val operator = enableWhen.operator.value) {
```
jingtang10, line 142: i think the structure here should be changed. here a function closure is returned... and the branching of operator happens when the function closure predicate is called. in other words if you need to evaluate multiple answers, each time you check which operator and choose the correct comparison logic. that seems logically redundant and an (admitedly tiny tiny) performance penalty.

jingtang10 (approved):
🎉🎉🎉

approved. just a few nits. thanks again peter!

datacapture/src/main/java/com/google/android/fhir/datacapture/enablement/EnablementEvaluator.kt
pld: cool yea definitely makes sense to keep the closure tight, there's probably room later for a further refactor here to get rid of some redundancy, was looking at the `compareTo` function, but couldn't get to a better place, will push in

jingtang10:
i'm not sure this bot is working...

jingtang10:
it seems to think your email address is pld@users.noreply.github.com for some reason...

pld:
Oh yea that's correct w/the email privacy option on, let me turn it off and rebase my commits to add my email...

pld:
should be set to peter@ona.io now

jingtang10:
from what I can see peter@ona.io is still in the state of `need_sender_cla`... urg i'm so sorry about this faff. can you check again?

pld:
@googlebot I signed it!

PR #249, opened 2021-02-19
create a maven repo from builds

yigit:
This PR adds a workflow to create a maven repo from builds

This simply attaches the build id to the version. We may eventually
make it duplicate the publication and also release with the right
version which we can use to upload to gmaven.


yigit:
triggering presubmit to see if it works :)

jingtang10:
@yigit how can this be tested? I struggled to find documentation online on the workflow.

jingtang10 (changes requested):
can you document (however briefly) how this can be used in either the readme file or in the user's guide on the wiki page? thanks!

https://github.com/google/android-fhir/wiki//User's-Guide

jingtang10 (approved):
Thanks for updating the wiki.

PR #277, opened 2021-02-25
Edit Text Widget, Date Widget, Date + Time Widget All have been orien…

MuhammadSalman-7214:
…ted according to our latest text Question Orientation


jingtang10 (changes requested):
Hi @MuhammadSalman-7214 thanks for this PR!

Looks good to me -- one thing, will you be able to figure out how to set the order of imports in android studio (i think you want alphabetically ordered rather than having java blocks at the end) and document your finding in https://github.com/google/android-fhir/wiki/Developer's-Guide in the set up section?

There're a few changes in this PR that change the order of imports which I think will actually break our spotless check.

jingtang10:
please update once you're done.

MuhammadSalman-7214:
> Hi @MuhammadSalman-7214 thanks for this PR!
> 
> Looks good to me -- one thing, will you be able to figure out how to set the order of imports in android studio (i think you want alphabetically ordered rather than having java blocks at the end) and document your finding in https://github.com/google/android-fhir/wiki/Developer's-Guide in the set up section?
> 
> There're a few changes in this PR that change the order of imports which I think will actually break our spotless check.

As we cannot configure IDE to follow spotlessApply rules, though we can set indentation and single import name, and I have updated that in developer documentation as well

jingtang10 (approved):
Thanks @MuhammadSalman-7214! Approved.

🎉

ps please resolve any conflicts before merging

PR #279, opened 2021-02-25
Clean-up the indexing code

jingtang10:
- Remove the FhirIndexer interface since we don't use Dagger any more
- Simplify the ResourceIndices data class by removing unnecessary `also` calls
- Reorder the indices as defined in FHIR


deepankarb (approved):
Thanks! Much cleaner and simpler now!

PR #294, opened 2021-03-09
Use ktfmt in spotless and switch (back) to google-java-format

jingtang10:
After trying out a few different linter options in spotless (see [documentation](https://github.com/diffplug/spotless/tree/main/plugin-gradle#kotlin)) and using a few different IDE plugins available for kotlin formatting in Android Studio, I've really seen the advantages of using ktfmt but decided to keep also ktlint in the list of linters in spotless (you can use multiple linters).

The documentation [here](https://github.com/facebookincubator/ktfmt) explains quite well why they decided to use google-java-format, not the least because of the [rectangle rule](https://github.com/google/google-java-format/wiki/the-rectangle-rule). It's very deterministic, which means it really takes formatting away as a concern for the developer.

See the linked issue for a bit more background.


joiskash (approved):
This is awesome!

MuhammadSalman-7214 (approved):

jingtang10:
thanks @MuhammadSalman-7214 and @joiskash for taking a look.

To other reviewers, if there's too much to go through, please focus on the non-kotlin files. Namely, the `ci.yml`, `build.gradle` and `deps.gradle` are the only 3 files that are interesting. the kt files have been reformatted automatically. thanks!

jingtang10:
changing java version used in github action which might have an impact on the integration tests

deepankarb (approved):

PR #301, opened 2021-03-10
Replace FQ id with logical id for Resources

### LINKED ISSUE #296: Insert unqualified (logical) ids for resources instead of fully qualified ids
**Describe the bug**
Upon sync, fully qualified IDs are being inserted into the database (in Resource, Index and Local change tables). This leads to crashes when deserialised resources are updated.

**Component**
Core library

**To Reproduce**
1. Sync resources from the HAPI FHIR server (happening automatically in the reference app.
2. Modify any resource property e.g. `birthDate`.
3. Call `FhirEngine.update(resource)`
4. Crash while selecting old resource for update.

**Root cause**

Calling `resource.id` on synced resources read from a `Bundle` yields fully qualified id, e.g. `http://hapi.fhir.org/baseR4/Patient/1915026`. 
Upon serialisation the base URL is removed and only the logical part is stored in the JSON representation of the resource. E.g.:
```
{
  "resourceType": "Patient",
  "id": "1915026",
 ...
``` 
After deserialising this JSON, calling `resource.id` would yield `1915026`. This will not match with the stored, fully qualified id leading to crashes and the database being essentially unusable.

**Expected behavior**
Synced resources should be update-able via `FhirEngine.update(resource)`

**Screenshots**
N/A

 - Device:  Pixel4a 
 - Android version: 11

**Additional context**
None.

deepankarb:
* `Resource.getId()` returns the fully qualified id if underlying fields are populated.
* We replace `Resource.getId()` with `Resource.getId().getIdPart()` to retrieve the logical id instead.

This fixes #296


core/src/main/java/com/google/android/fhir/Util.kt
```diff
@@ -33,3 +34,9 @@ internal fun Date.toTimeZoneString(): String {
     ).withZone(ZoneId.systemDefault())
     return simpleDateFormat.format(this.toInstant())
 }
+
+/**
+ * Returns only the logical ID part of this Resource's ID. For example, if the ID is
+ * "http://example.com/fhir/Patient/123/_history/456", then this method would return "123".
+ */
+internal fun Resource.logicalId(): String = this.idElement.idPart
```
jingtang10, line 42: I would change this to be a property rather than a function :)

```
internal val Resource.logicalId: String
  get() = this.idElement.idPart
```

jingtang10 (changes requested):
thanks @deepankarb :)

core/src/main/java/com/google/android/fhir/Util.kt
deepankarb: I kept thinking about this but was too lazy to look this up. Thanks!

jingtang10: please comment once updated

jingtang10 (changes requested):

jingtang10:
https://github.com/google/android-fhir/pull/294 has been merged. So please rebase your branch onto `master` (or merge `master` into your feature branch if rebase is too difficult) and follow [this section](https://github.com/google/android-fhir/wiki/Developer's-Guide#kotlin-coding-conventions) in the developer's guide to set up your environment to use the new coding conventions. Thanks!

core/src/main/java/com/google/android/fhir/Util.kt
deepankarb: Updated to property.

jingtang10 (approved):
Thanks! 😀👌

PR #303, opened 2021-03-10
Skip adding row if there's no change in resource

### LINKED ISSUE #295: Not insert a local change row when the Resource diff is empty
**Describe the bug**

**Component**
Core library

**To Reproduce**
1. Insert a new `Resource`
2. Do not modify the `Resource` instance
3. Update the `Resource` with `FhirEngine.update(resource)` 

**Expected behavior**
Since we didn't make any change and the JSON diff is an empty array : `[]`, no row should be inserted in the `LocalChangesEntity` table.

**Screenshots**

Any device

**Additional context**
JSON diffs are being passed back as strings to the `LocalChangesDao` hence checks would be kludgy. Returning a `JSONArray` or other object will enable check for emptiness.

deepankarb:
This fixes #295 

* If `update` is called on an unmodified resource, we do not update the Local Change table.


jingtang10 (commented):
OK I'm just worried there are no failing tests. what tests should we add that should have failed as a result of this change?

jingtang10:
> > OK I'm just worried there are no failing tests. what tests should we add that should have failed as a result of this change?
> 
> The squashing algorithm works fine with empty changes. There's no API besides checking the table for row count change. Not sure how to test this reliably. I'll think a bit more about this.

can you please create an issue for this? thanks

jingtang10 (changes requested):
please create an issue for the test

jingtang10 (approved):

jingtang10:
please update the branch

PR #317, opened 2021-03-12
Update the gradle wrapper version to 6.8.3 using the wrapper workflow

jingtang10:
Ran `./gradlew wrapper --gradle-version 6.8.3`


gradlew
```diff
@@ -109,10 +126,11 @@ if $darwin; then
     GRADLE_OPTS="$GRADLE_OPTS \"-Xdock:name=$APP_NAME\" \"-Xdock:icon=$APP_HOME/media/gradle.icns\""
 fi
 
-# For Cygwin, switch paths to Windows format before running java
-if $cygwin ; then
+# For Cygwin or MSYS, switch paths to Windows format before running java
+if [ "$cygwin" = "true" -o "$msys" = "true" ] ; then
     APP_HOME=`cygpath --path --mixed "$APP_HOME"`
```
pankajkatoch, line 131: I'm not well conversant with MingW installations. Do you know if MingW always includes cygpath binary? Else we need to fail gracefully saying cygpath is required.

Ignore comment if cygpath is always available.

pankajkatoch (approved):

gradlew
jingtang10: i'm not so sure, but this change is automatically generated by running `./gradlew wrapper --gradle-version 6.8.3` so hopefully it's good.

PR #324, opened 2021-03-13
fix: Fixes reference app crash on triggering Evaluation

### LINKED ISSUE #319: Reference app crashes when asked for evaluation
**Describe the bug**
Reference App in cql loading screen crashes when asked for evaluation 

**Component**
Reference Application

**To Reproduce**
1. Click on `Load CQL Resources & Patients` from the menu at top right
2. Click evaluate button and witness the crash

The crash occurs irrespective of the arguments supplied, i.e., no matter whether those params are empty or filled

**Expected behavior**
No crash

**Screenshots**
NA

 - Device: Samsung Galaxy M31
 - Android version: 10
 - Build number: RP1A.200720.012.M315FXXU2BUAC

**Additional context**
NA

s-ayush2903:
Close #319 
* Checked and fixed nullability issues in engine implementation
* Moved the files to `core` module which were being used in multiple modules
* Abstracted out the FhirApplication to CoreModule as `CoreApplication`

~PS: The engine doesn't run fine `cqlreference`(throws class-cast exception, saying `android.app.application` cannot be cast to `{packagename}.cqlreference.FhirApplication`) but works smoothly in the `reference` application; attempting a fix~(fixed as per 74da8e1)


s-ayush2903:
@deepankarb @jingtang10 Would it be Ok to create a factory in this PR and use it to create instances of FhirEngine, as currently  we have different `FhirApplication`s(BaseApplication classes that extend the `CoreApplication`(class which is the abstraction of FhirApplication class)) for `reference` and `cqlreference` modules whose companion object method is accessed to create a `fhirEngine`.

s-ayush2903:
Why is spotless check failing, irrespective of the fact that I've same settings as mentioned in [wiki](https://github.com/google/android-fhir/wiki/Developer's-Guide#kotlin-coding-conventions) : (
See:
|Indents for Kt|Ktfmt settings|
|--|---|
|||

jingtang10 (changes requested):
@s-ayush2903 Hi thanks for this PR and apologies for the delay in reviewing. Been trying to work out the right approach and architecture here.

I created #330 where I added a little more details. Basically, we want to separate the CQL functionalities from the `core` module. I think this PR should remove rather than add to the `core` module. Things like `CoreApplication.kt` should be in the new `cqlreference` application instead.

I think the new `cqlreference` module should probably also use hardcoded data rather than data read from the db. But that can be in a separate PR.

I understand this is quite an involved PR. Thanks for taking this on. I'm happy if you want to pick other issues to work on first. But please do comment first.

s-ayush2903:
>  Things like `CoreApplication.kt` should be in the new `cqlreference` application instead.

As of now, I don't have any plausible reason to move `CoreApplication.kt` to `cqlreference` module, as it is being used by both, the `reference` module and the `cqlreference` module(to instantiate fhirEngine), also the files moved to the core module have the same reason as mentioned for the `CoreApplication.kt`; we can do move 'em to `cqlreference` module but they won't be accessible from `reference` module and we'll have to rewrite 'em there(in `reference` module) as well, thereby increasing the code duplicacy.

core/src/main/java/com/google/android/fhir/db/impl/DatabaseImpl.kt
```diff
@@ -34,28 +34,28 @@ import org.hl7.fhir.r4.model.ResourceType
  * The implementation for the persistence layer using Room. See docs for
  * [com.google.android.fhir.db.Database] for the API docs.
  */
-internal class DatabaseImpl(context: Context, private val iParser: IParser, databaseName: String?) :
```
jingtang10, line 37: please revert changes in this file. the database need to be internal and should not be exposed to applications. the sdk's APIs should be exposed by `FhirEngine`.

cqlreference/src/main/java/com/google/android/fhir/cqlreference/cql/CqlEngineUtils.kt
```diff
@@ -0,0 +1,65 @@
...
+    val contextSplit = contextString.split("/").toTypedArray()
+    contextMap[contextSplit[0]] = contextSplit[1]
+    val versionedIdentifier = VersionedIdentifier().withId(libraryVersionId)
+    val expressions: MutableSet<String> = HashSet()
+    expressions.add(expression)
+    val map: MutableMap<VersionedIdentifier, Set<String>> = HashMap()
+    map[versionedIdentifier] = expressions
+    return if (!map.get("system").isNullOrEmpty())
+      Factory().createCqlEngine(context).evaluate(contextMap, null, map)
+    else null
```
jingtang10, line 63: i think the best course of action for now is to remove this code and add an not implemented exception before we rewrite this application.

cqlreference/build.gradle
```diff
@@ -0,0 +1,79 @@
...
+    implementation deps.core
+    implementation deps.appcompat
+    implementation 'androidx.core:core-ktx:1.3.2'
+    implementation deps.kotlin.test
+    implementation deps.lifecycle.viewmodel_ktx
+    implementation deps.material
+    testImplementation deps.junit
+    testImplementation deps.atsl.core
+    implementation deps.constraint_layout
+    implementation project(path: ':core')
```
jingtang10, line 71: remove this dependency for now. we haven't defined an API for the cql reference app to use the core lib correctly yet.

jingtang10 (changes requested):
thanks! i think what i probably haven't made very clear is that we'll need to rewrite the cql reference app (see the linked issue) by depending on a different library (`cql-evaluator` rather than `cql-engine`).

I think the best thing to do is actually just removing the cql activity from the reference app altogether and remove the evaluatecql api. and then create a skeleton app called cqlreference.

I think what's been changed in this PR about the database API is not right -- we don't want to expoose the db to applications.

sorry about this - but I think you're better off just starting afresh with a new PR. i'm happy to do that as well if you just want to take on a new issue?

cqlreference/build.gradle
s-ayush2903: If we remove this, then we'll have to write the FhirEngine interface and its impl both again in the cqlref module, as there is an option of downloading the fhir res in the cqlLoadActivity and for that we'll have to write a functionality to save the Res(for which the working engine is needed). So there exists some way of achieving the same behavior(as current one's) without adding it as a dependency, that'll be interesting.

s-ayush2903:
>  we'll need to rewrite the cql reference app (see the linked issue) by depending on a different library (cql-evaluator rather than cql-engine).

Oh this makes it clear, I was under the impression that we need to separate the engine only. 

> just starting afresh with a new PR


So will I have to create a new PR for this issue then, what's the problem with this one? I can rewrite commit history if that is an issue, rest is on your preference

cqlreference/build.gradle
jingtang10: the direction we'll be taking with the cql ref module is that we will be using hard-code resources for now. so it won't have a dependency on the core module where there's a db. this will be a breaking change and we will need to rewrite the cql ref module (the UI).

s-ayush2903: So it means, I'll have to hardcode it here in this PR. Ok fine :+1:, I was unsure as you mentioned that it could be done in a separate PR

jingtang10: can yo uwork on this issue as a priority instead of grabbing other issues? this is blocking to other things.

i don't mind even if you just leave things in the cql reference module unimplemented. the key thing that needs to happen in the core module is the evaluateCql api needs to be deleted as well as the cql folder.

core/build.gradle
```diff
@@ -91,7 +91,10 @@ dependencies {
     implementation deps.room.runtime
     implementation deps.work.runtime
     implementation deps.json_tools.json_patch
-
+    implementation deps.okhttp_logging_interceptor
+    implementation deps.retrofit.core
+    implementation deps.retrofit.gson
+    implementation deps.retrofit.mock
```
jingtang10, line 97: remove these.

jingtang10 (changes requested):

settings.gradle
```diff
@@ -2,3 +2,4 @@ include ':core'
 include ':datacapture'
 include ':datacapturegallery'
 include ':reference'
+include ':cqlreference'
```
jingtang10, line 5: please move this up so that these modules are sorted alphabetically. thx!

reference/src/main/java/com/google/android/fhir/reference/PatientListViewModel.kt
```diff
@@ -64,7 +64,7 @@ class PatientListViewModel(application: Application, private val fhirEngine: Fhi
         .run()
     Log.d(
       "PatientListViewModel",
-      "${searchResults.count()} search results: " + "${searchResults.joinToString(" ")}"
+      "${searchResults.count()} search results: " + searchResults.joinToString(" ")
```
jingtang10, line 67:
```suggestion
      "${searchResults.count()} search results: ${searchResults.joinToString(" ")}"
```

reference/src/main/java/com/google/android/fhir/reference/FhirApplication.kt
```diff
@@ -22,15 +22,14 @@ import androidx.work.Constraints
 import ca.uhn.fhir.context.FhirContext
 import com.google.android.fhir.FhirEngine
 import com.google.android.fhir.FhirEngineBuilder
-import com.google.android.fhir.reference.api.HapiFhirService.Companion.create
-import com.google.android.fhir.reference.data.FhirPeriodicSyncWorker
-import com.google.android.fhir.reference.data.HapiFhirResourceDataSource
```
jingtang10, line 27: keep these intact please.

jingtang10 (changes requested):

core/build.gradle
s-ayush2903: 7d4e820

settings.gradle
s-ayush2903: 7a220c7

reference/src/main/java/com/google/android/fhir/reference/PatientListViewModel.kt
s-ayush2903: 7a220c7

reference/src/main/java/com/google/android/fhir/reference/FhirApplication.kt
s-ayush2903: 7a220c7

reference/src/main/java/com/google/android/fhir/reference/api/FhirConverter.kt
```diff
@@ -14,7 +14,7 @@
  * limitations under the License.
  */
 
-package com.google.android.fhir.reference.api
+package com.google.android.fhir.api
```
jingtang10, line 17: revert this change

reference/src/main/java/com/google/android/fhir/reference/data/FhirPeriodicSyncWorker.kt
```diff
@@ -19,13 +19,13 @@ package com.google.android.fhir.reference.data
 import android.content.Context
 import androidx.work.WorkerParameters
 import com.google.android.fhir.FhirEngine
-import com.google.android.fhir.reference.FhirApplication
+import com.google.android.fhir.reference.FhirApplication.Companion.fhirEngine
```
jingtang10, line 22: revert this change - seems unnecessary.

jingtang10 (approved):
Thanks @s-ayush2903 for your contribution! Just a couple of small nits.

👍

reference/src/main/java/com/google/android/fhir/reference/data/FhirPeriodicSyncWorker.kt
s-ayush2903: Is it a poor practice to use static imports? d075369

reference/src/main/java/com/google/android/fhir/reference/api/FhirConverter.kt
s-ayush2903: d075369

reference/src/main/java/com/google/android/fhir/reference/data/FhirPeriodicSyncWorker.kt
jingtang10: oops missed this i don't think we have a strong view here actually :) i just thought it would provide a bit more context including the classname in the invocation. but i don't feel strongly.

jingtang10:
Thanks @s-ayush2903 for preservering - this is a structural change and we had some direction changes. Appreciate it.

👍 👍 👍

s-ayush2903:
Thanks to you too for clarifying stuff. With this, maybe an Issue can be created that targets about the todos in the cqlref module. And I would be happy to work on it! Thoughts?

PR #329, opened 2021-03-16
Move sdc library back to HAPI structures

jingtang10:


pld (approved):

joiskash (approved):
Huge change! I did not go through every line but i did pull the branch and tested it on my phone. The only difference I see is that the Questionnaire response json is not "pretty". It doesn't matter. Just an observation between hapi and fhir-protos I guess. Thanks Jing!
Screenshot_20210318-130953

jingtang10:
Yeah that's right @joiskash - HAPI formats the JSON output slightly differently. We can look into JSON formatters if this becomes an issue.

deepankarb (approved):

Issue #336, opened 2021-03-18
Gradle Sync Failure 


Ana2k:
**Describe the bug**
Gradle Sync failure due to beta version

**To Reproduce**
Steps to reproduce the behaviour:
1. Fork the repo
2. Build the app in stable version of Android studio

**Expected behavior**
Succesful Gradle Snc and stable build of app.

**Screenshots**
![Screenshot from 2021-03-18 21-31-32](https://user-images.githubusercontent.com/54909114/111657576-9173ea00-8831-11eb-830a-b0e2e1de0ee4.png)

**Android Studio :**
 - Studio Version [4.1.2]
 
**Additional context**
Upon doing gradle version changes app runs well.

Ana2k:
@jingtang10 I would like to work on this issue. Can you assign it to me please?

epicadk:
You can take a look at [this discussion](https://github.com/google/android-fhir/discussions/312)

Ana2k:
> You can take a look at [this discussion](https://github.com/google/android-fhir/discussions/312)

 @epicadk I had resolved the issue and made a pr already. 
 Thanks for the discussion thread link anyways!

jingtang10:
Please see developer's guide here: https://github.com/google/android-fhir/wiki/Developer's-Guide which recommends using AS 4.2. I'm closing this issue and if your issue isn't resolved after upgrading to 4.2 please feel free to re-open with more details. Thanks!

atm1504:
@jingtang10 I was scrolling through the issue section and found this issue. Even, while building the app I faced the same issue. I went through the guidelines  https://github.com/google/android-fhir/wiki/Developer's-Guide and found that it asks to use 4.2+ version of the android studio which is the beta version currently.

I did as stated. I downloaded the latest beta release which is Android 4.Beta 6. but the issue raised again. To make the app run, I have to update the Gradle version from 4.2.0-beta02 to 4.2.0-beta06

I hope the best thing would be to make the app compatible to run in the latest stable version of android studio and the latest stable Gradle version. As most of us would really be using the latest stable versions.

Please look into this issue and update us. 

Below is the attached ss
![Screenshot from 2021-03-23 16-55-48](https://user-images.githubusercontent.com/43731599/112139353-ac12de00-8bf8-11eb-857e-1541ab9eff62.png)

deepankarb:
[...]

PR #337, opened 2021-03-18
fix: Gradle Sync Failure Fixed

Ana2k:
fixes: #336 

Screenshot from 2021-03-18 23-39-53

Steps:
1. Updated deps.gradle file to previous android version<Plugin 4.1.1>
2. gradle version to 6.5-all.zip updated in wrapper


Ana2k:
@jingtang10 @deepankarb please review this fix. for #336 issue

joiskash:
If you want to use android version 4.2 (which is needed for ktfmt to get detected as a plugin and for unit tests to not randomly crash) you have to make sure you reinstall android studio from this [link](https://developer.android.com/studio/preview). I think  because these are preview builds, you cannot go from android 4.1 to 4.2 without doing a fresh install of android studio. I have faced this issue before. Just make sure you uninstall your current version of android studio before installing 4.2 and above.

PS: downgrading to 4.1 is not a fix.

jingtang10:
Thanks @Ana2k for the PR - please try upgrading to AS 4.2 first. If that doesn't work please reopen the issue and we can discuss a solution.

PR #357, opened 2021-03-20
remove jcenter

epicadk:
Thanks to [Coolmind](https://stackoverflow.com/questions/49573157/android-studio-3-1-could-not-find-org-jetbrains-trove4jtrove4j20160824)


epicadk:
Not exactly sure why the upload is failing. I will look into it.

.idea/jarRepositories.xml
```diff
@@ -47,5 +47,10 @@
       <option name="name" value="maven" />
       <option name="url" value="https://oss.sonatype.org/content/repositories/snapshots" />
     </remote-repository>
+    <remote-repository>
+      <option name="id" value="Gradle Central Plugin Repository" />
+      <option name="name" value="Gradle Central Plugin Repository" />
+      <option name="url" value="https://plugins.gradle.org/m2" />
+    </remote-repository>
```
jingtang10, line 54: do you actually need this?

build.gradle
```diff
@@ -3,7 +3,8 @@ buildscript {
     apply from: 'deps.gradle'
     repositories {
         google()
-        jcenter()
+        mavenCentral()
+        gradlePluginPortal()
```
jingtang10, line 7: have you tested without this? which artifact do you need this for?

build.gradle
```diff
@@ -16,8 +17,8 @@ buildscript {
 allprojects {
     repositories {
         google()
-        jcenter()
         mavenCentral()
+        gradlePluginPortal()
```
jingtang10, line 21: and this?

jingtang10:
Thanks for the fix! A few comments.

jingtang10 (changes requested):
please see comments.

build.gradle
epicadk: It's required for this artifact `org.jetbrains.trove4j:trove4j:20160824`

.idea/jarRepositories.xml
epicadk: Sorry wasn't sure. Removed it.

s-ayush2903:
@epicadk Can you please keep this PR intact for some time, as #364 is almost to finish and if this'll get merged then there'll be a ton of conflicts to resolve as the aforementioned PR has so many changes :pleading_face:

build.gradle
epicadk: same as above.

jingtang10 (approved):
Thanks for this fix! 👍

PR #370, opened 2021-03-22
Change CI API level from 29 to 30

jingtang10:


deepankarb:
Failed with a different error on API 30 (not OOM). I have triggered a re-run.

deepankarb (approved):

Issue #282, opened 2021-03-05
Improve ResourceIndexerTest.kt


jingtang10:
https://github.com/google/android-fhir/blob/master/core/src/test/java/com/google/android/fhir/index/impl/ResourceIndexerTest.kt

- It's really difficult to read the tests because the actual test resources are hidden away in json files in a different folder
- It would be good to really lock down the behavior by having exhaustive unit tests
- It might be useful to have some 'integration tests' where we use some real world resources, and check that everything that should be indexed is indexed

At the moment the test cases seem to be somewhere in between unit and integration tests. in the unit tests i would love for the test resources to be really simple -- as long as we hit the code path we're sorted. in the integration tests it makes sense to have more complex resources.

pld:
Related to the broader theme of code coverage mentioned here is this issue about tracking coverage and enforcing in CI, https://github.com/google/android-fhir/issues/233

epicadk:
Can I work on this? I have added some unit tests in #332 I'll setup the integration tests in a different folder.

epicadk:
I was making a few of my own FHIR resources for testing and I had a question. Should I only add resources marked as normative to the tests or will an [maturity level](https://www.hl7.org/fhir/versions.html#maturity) do?

jingtang10:
@epicadk please feel free to send a pr

PR #387, opened 2021-03-24
Improve ResourceIndexer tests

epicadk:
Fixes: #282

**Description:**

1.  Renamed unit tests to signify the code path they test. 
2.  Use minimally constructed resources to test specific types
3. Add integration tests that use resources parsed from JSON files and verify that all indices are created.
4. Removed setup method so that only the resource being used in the test is parsed as compared to all resources when using the setup method. 

**Type** 
  Testing


jingtang10:
is there any update on this one? thanks!

jingtang10:
please make sure @deepankarb also has a chance to review once done. thanks!

epicadk:
> is there any update on this one? thanks!

Sorry for the delay. Will be done within a day.

epicadk:
Sorry something urgent came up. I've completed adding unit tests. Will finish integration tests soon.

epicadk:
I need to place the integration tests in a different folder. I was thinking I can create a separate directory for unit tests and integration tests under the test directory. Any suggestions?

epicadk:
I've also added some tests in #332 for the date types. I've covered all cases except the integer case of the number index. I can't seem to find a resource that uses a search parameter of that type.  Even the example given on the specification website doesn't work.

core/src/test/java/com/google/android/fhir/ResourceIndexerTest.kt
```diff
@@ -0,0 +1,280 @@
+/*
```
jingtang10, line 1: I don't think it's right to create another file in this package. This should be in the same directory as the other file.

jingtang10 (changes requested):
It might be ok to keep everything in the same file, but you should have two classes of test cases

1. cases that test specific code paths e.g. `string_empty_shouldIndexNothing` or `int_shouldCreateIndex` etc etc. We would use minimally constructed resources to test specific types.
2. cases where we use more complex resources and assert that all indecies that should be created are created.

I think these two classes of tests can probably co-exist in the same file if named sensibly.

core/src/test/java/com/google/android/fhir/ResourceIndexerTest.kt
epicadk: > I don't think it's right to create another file in this package. This should be in the same directory as the other file.

Reverted this change

epicadk:
I've kept the order similar as they occur in the resource indexer. I've also included tests for null and empty values wherever possible.

engine/src/test/java/com/google/android/fhir/index/ResourceIndexerTest.kt
```diff
@@ -26,548 +26,803 @@ import com.google.android.fhir.index.entities.ReferenceIndex
...
-        MolecularSequence::class.java,
-        "/number_test_molecular_sequence.json"
-      )
-    specialTestLocation =
-      testingUtils.readFromFile(Location::class.java, "/location-example-hl7hq.json")
+    testPatient = testingUtils.readFromFile(Patient::class.java, "/date_test_patient.json")
+    testLocation = testingUtils.readFromFile(Location::class.java, "/location-example-hl7hq.json")
   }
 
+  /** Integration tests for ResourceIndexer. */
```
jingtang10, line 87: can you move the 'integration' tests after the 'unit' tests?

engine/src/test/java/com/google/android/fhir/index/ResourceIndexerTest.kt
```diff
@@ -26,548 +26,803 @@ import com.google.android.fhir.index.entities.ReferenceIndex
...
 
+  /** Unit tests for resource indexer */
   @Test
-  fun index_patient_emptyGivenName_shouldNotIndexGivenName() {
-    val resourceIndices = ResourceIndexer.index(TEST_PATIENT_NULL_FIELDS)
-    assertThat(
-        resourceIndices.stringIndices.any { stringIndex ->
-          stringIndex.path.equals("Patient.name.given")
-        }
+  fun index_Patient_shouldIndexLastUpdatedDate() {
```
jingtang10, line 322: the point of these unit test cases is to cover all code paths so i think they need to be re-organized by index type, and please rename them, e.g.:

```
fun index_string_null
fun index_string
...

fun date_period_someCondition
fun date_period_someOtherCondition
...
```

jingtang10 (changes requested):
A couple of high-level comments. Please address! Thanks!

epicadk:
Noticed a bug in the date indexer while testing so I made the corresponding change. The value to should be inclusive (makes querying easier using the SQL between operator.)

jingtang10:
@epicadk did you remove all the other changes in the test class you made?

epicadk:
> @epicadk did you remove all the other changes in the test class you made?

Yup all of the tests are in the same file and I've renamed the tests as requested [here](https://github.com/google/android-fhir/pull/387#discussion_r616014331). I've also updated the pr description to explain the changes I have made.

engine/src/test/java/com/google/android/fhir/index/ResourceIndexerTest.kt
```diff
@@ -26,542 +26,831 @@ import com.google.android.fhir.index.entities.ReferenceIndex
...
+          "Invoice.participant.role",
+          testInvoice.participantFirstRep.role.codingFirstRep.system,
+          testInvoice.participantFirstRep.role.codingFirstRep.code
+        )
+      )
+
+    assertThat(resourceIndices.uriIndices).isEmpty()
+
+    assertThat(resourceIndices.dateIndices)
       .contains(
```
jingtang10, line 657:
```suggestion
      .containsExactly(
```

jingtang10 (changes requested):
@epicadk Really happy with this - this is a very good PR! Thank you so much for this!

Apologies for the delay in reviewing. Please let me know once you're done with the few small comments I made.

epicadk:
@jingtang10 divided the tests in 3 parts.

1.  The creation of resources 
2. Actual code for indexing
3. assertions

jingtang10 (approved):
Great! Thanks!

PR #391, opened 2021-03-24
Use Maven Publish Plugin for the core module

Andati:
**Description:**
Add support to publish core.
You can publish core locally via command: `./gradlew :core:publishToMavenLocal` to include in your project for local testing

**Type**
Builds


jingtang10:
Hi @Andati is this ready for review? thanks!

Andati:
@jingtang10 I see that the master has changed, let me update the pr first

core/build.gradle.kts
```diff
@@ -2,6 +2,33 @@ plugins {
...
+          licenses {
+            license {
+              name.set("The Apache License, Version 2.0")
+              url.set("http://www.apache.org/licenses/LICENSE-2.0.txt")
+            }
+          }
+        }
+      }
+    }
+  }
```
s-ayush2903, line 31: Maybe we can extract it out as a common functionality(like [SpotlessConfig](https://github.com/google/android-fhir/blob/master/buildSrc/src/main/kotlin/SpotlessConfig.kt#L21), and rename file to `BuildConfig.kt` or something more semantic), as the config is nearly same to that of dc; 
https://github.com/google/android-fhir/blob/b337529fcbac32556c580bdd79b4aca43b9feaeb/datacapture/build.gradle.kts#L18-L37

Thoughts?

s-ayush2903: Ah, I referred to extracting the publishing method in  `afterEvaluate{}`, not everything

jingtang10: We can. Not a high priority though.

jingtang10 (approved):
Thanks @Andati! Can you please rename the PR title to "Use Maven Publish Plugin for the core module" to be more exact?

Issue #57, opened 2020-06-02
[Backlog] Investigate using DSL in Kotlin for search


jingtang10:

s-ayush2903:
I would like to work on this issue.

PR #395, opened 2021-03-26
Rewrite the search API in Kotlin DSL

jingtang10:

**Fixes** #57 

**Description**
Use Kotlin DSL for the search API.

Before

```
fhirEngine
  .search()
  .of(Patient::class.java)
  .filter(string(Patient.ADDRESS_CITY, ParamPrefixEnum.EQUAL, "NAIROBI"))
  .sort(stringSort(Patient.GIVEN, true))
  .limit(10)
  .skip(10)
  .run()
```

After
```     
fhirEngine.search<Patient> {
  filter(Patient.ADDRESS_CITY) {
    prefix = ParamPrefixEnum.EQUAL
    value = "NAIROBI"
  }
  sort(Patient.GIVEN, Order.ASCENDING)
  count = 10
  from = 10
}
```

Improvements:
- use Kotlin DSL to achieve the idiomatic Kotlin style
- more concise APIs that maintain type-safety (e.g. no need to specify `string` type in `filter` API any more but `string` type will be inferred from the search parameter provided)
- reified type parameter makes the function call more concise (can use reified generic type to infer type without passing the type parameter explicitly)
- reworked the API naming to move away from SQL terms, e.g. `size` instead of `limit` and `from` instead of `skip`

**Alternative(s) considered**
Existing APIs are signficantly improved by Kotlin DSL for reasons mentioned above.

**Type**
Feature

**Checklist**


core/src/main/java/com/google/android/fhir/search/MoreSearch.kt
```diff
@@ -0,0 +1,90 @@
...
+import org.hl7.fhir.r4.model.Resource
+import org.hl7.fhir.r4.model.ResourceType
+
+suspend fun <R : Resource> Search.run(database: Database): List<R> {
+  return database.search(getQuery())
+}
+
+fun Search.getQuery(): SearchQuery {
+  val filterQuery =
+    (stringFilters.map { it.query(type!!) } + referenceFilter.map { it.query(type!!) }).intersect()
```
s-ayush2903, line 29: We can avoid double bang/potential crash using null safety: 
```kotlin
val filterQuery = type?.let { (stringFilters.map { it.query(type) } + referenceFilter.map { it.query(type) }).intersect() }
```

Not a review just a point you might have missed :sweat_smile:, as this was the reason for reference application to crash- presence of so many double bangs

jingtang10: Thanks - I've made `type` non-nullable now.

pld (approved):
awesome

core/src/main/java/com/google/android/fhir/search/MoreSearch.kt
```diff
@@ -0,0 +1,104 @@
...
+ * limitations under the License.
+ */
+
+package com.google.android.fhir.search
+
+import com.google.android.fhir.db.Database
+import org.hl7.fhir.r4.model.Resource
+import org.hl7.fhir.r4.model.ResourceType
+
+suspend fun <R : Resource> Search.run(database: Database): List<R> {
```
yigit, line 23: lets use `execute` or `executeWith` instead of `run`.

core/src/main/java/com/google/android/fhir/search/MoreSearch.kt
```diff
@@ -0,0 +1,104 @@
...
+
+fun Search.getQuery(): SearchQuery {
+  val filterQuery =
+    (stringFilters.map { it.query(type) } + referenceFilter.map { it.query(type) }).intersect()
+
+  val queryBuilder = StringBuilder()
+  queryBuilder.appendLine(
+    """
+    SELECT a.serializedResource
+    FROM ResourceEntity a${sort?.let {
```
yigit, line 35: better to move this to a variable since it is not very readable as inline

core/src/main/java/com/google/android/fhir/search/MoreSearch.kt
```diff
@@ -0,0 +1,104 @@
...
+  )
+  filterQuery?.also {
+    queryBuilder.appendLine(it.query)
+    queryBuilder.appendLine(")")
+  }
+  sort?.also {
+    queryBuilder.appendLine(
+      """
+      ORDER BY b.index_value ${
+        when (order!!) {
```
yigit, line 51: this should give a proper error if order is null.
similar to sort, it is better to move to a `val` and then use it here.

e.g.

```
val orderStmt = if (sort != null && order != null) {
   ...
} else {
   ""
}
```

core/src/main/java/com/google/android/fhir/search/Search.kt
```diff
@@ -16,43 +16,11 @@
...
-    /** Returns a [SearchSpecifications] object that only includes the first [limit] results. */
-    fun limit(limit: Int): SearchSpecifications
-
-    /** Returns a [SearchSpecifications] object that skips the first [skip] results. */
-    fun skip(skip: Int): SearchSpecifications
-
-    /** Runs a search with the [SearchSpecifications]. */
-    suspend fun <R : Resource> run(): List<R>
-  }
+suspend inline fun <reified R : Resource> FhirEngine.search(init: Search.() -> Unit): List<R> {
```
yigit, line 22: we probably still need an overload that receives the class for java clients.

core/src/main/java/com/google/android/fhir/search/MoreSearch.kt
jingtang10: ah ok - i'm guessing this is to avoid clashing with the kotlin scope function?

jingtang10: done

core/src/main/java/com/google/android/fhir/search/Search.kt
jingtang10: https://github.com/google/android-fhir/issues/403

core/src/main/java/com/google/android/fhir/search/MoreSearch.kt
jingtang10: reorged.

core/src/main/java/com/google/android/fhir/search/MoreSearch.kt
jingtang10: reorged.

jingtang10:
Thanks @yigit I rewrote the query generation code.

brynrhodes:
Very cool, I just have one minor nitpick here, if you are changing the "limit" search term, I'd suggest using "count" as the term more aligned with FHIR [search](http://hl7.org/fhir/search.html#count)

core/src/main/java/com/google/android/fhir/search/MoreSearch.kt
jingtang10: thanks @kevinmost - i've made it a property if that's ok.

core/src/main/java/com/google/android/fhir/search/MoreSearch.kt
```diff
@@ -0,0 +1,116 @@
+/*
```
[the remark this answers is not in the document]
jingtang10: I'm having trouble converting the `.addAll`. Getting a strange error in the IDE...  `Type mismatch. Required MutableList<Any> Found: List<Any>`... couldn't work out why

PR #413, opened 2021-03-31
SpotlessConfig on entire project

### LINKED ISSUE #412: Run Spotless in buildSrc as well
**Describe the Issue**
Files that fall under `buildSrc` folder also should be formatted automatically via spotless

Yems

s-ayush2903:

Fixes #412 

**Description**
Spotless now runs on entire module irrespective of module depth & surface

**Alternative(s) considered**
Extracted out the common spotless functionality to a file and applied in root gradle file

**Type**
Choose one: (| Code health |)

NA

**Checklist**

PS: Branched off of [`jt/kotlin-gradle-spotless`](https://github.com/google/android-fhir/tree/jt/kotlin-gradle-spotless)


s-ayush2903:
Set an upper bound on runtime for pipelines so that we don't end up in situations like [this](https://github.com/google/android-fhir/pull/413/checks?check_run_id=2241107660)?

.idea/jarRepositories.xml
```diff
@@ -25,7 +24,10 @@
     <remote-repository>
       <option name="id" value="jboss.community" />
       <option name="name" value="JBoss Community repository" />
-      <option name="url" value="https://repository.jboss.org/nexus/content/repositories/public/" />
+      <option
+                name="url"
```
jingtang10, line 28: this seems odd - do you know why the indentation is so big?

buildSrc/src/main/kotlin/SpotlessConfig.kt
```diff
@@ -0,0 +1,45 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import org.gradle.api.Project
+import org.gradle.kotlin.dsl.apply
+import org.gradle.kotlin.dsl.configure
+
+fun Project.configureSpotless() {
```
jingtang10, line 21: the only effective change that leads to spotless being applied to the whole project is remove the config from `subprojects` block right?

buildSrc/src/main/kotlin/SpotlessConfig.kt
```diff
@@ -0,0 +1,45 @@
...
+  apply(plugin = "com.diffplug.spotless")
+  configure<com.diffplug.gradle.spotless.SpotlessExtension> {
+    kotlin {
+      target("**/*.kt")
+      targetExclude("**/generated-sources/**")
+      ktlint().userData(mapOf("indent_size" to "2", "continuation_indent_size" to "2"))
+      ktfmt().googleStyle()
+      licenseHeaderFile(
+        "${project.rootProject.projectDir}/license-header.txt",
+        "package|import|class|object|sealed|open|interface|abstract "
```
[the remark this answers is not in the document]
jingtang10: how come this is needed now?

buildSrc/src/main/kotlin/SpotlessConfig.kt
```diff
@@ -0,0 +1,45 @@
...
+import org.gradle.api.Project
+import org.gradle.kotlin.dsl.apply
+import org.gradle.kotlin.dsl.configure
+
+fun Project.configureSpotless() {
+  apply(plugin = Plugins.BuildPlugins.spotless)
+  configure<com.diffplug.gradle.spotless.SpotlessExtension> {
+    kotlin {
+      target("**/*.kt")
+      targetExclude("**/generated-sources/**")
```
jingtang10, line 26: why is this change needed?

buildSrc/src/main/kotlin/SpotlessConfig.kt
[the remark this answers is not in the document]
s-ayush2903: Because spotless needs some kind of declaration in the file to find a line before which to apply license-header

https://github.com/diffplug/spotless/issues/135

buildSrc/src/main/kotlin/SpotlessConfig.kt
s-ayush2903: Yes

buildSrc/src/main/kotlin/SpotlessConfig.kt
s-ayush2903: To not apply spotless on generated build files

buildSrc/src/main/kotlin/SpotlessConfig.kt
s-ayush2903: Yes, should it be moved back to gradle file?

.idea/jarRepositories.xml
s-ayush2903: Yeah, that's weird, I didn't pay any attention to it after applying spotless. Let me see

buildSrc/src/main/kotlin/SpotlessConfig.kt
s-ayush2903: Ah, its needless 8877df3

.idea/jarRepositories.xml
s-ayush2903: Ah, here comes an interesting fact: when I run spotless after moving this code section to some XML file(that isn't in any "dot directory", like `.idea`), it automatically formats it correctly, but doesn't work when the same patch is in a dot directory. Feels like there's some problem with prettier.

jingtang10: Can you please file a bug in prettier codebase?

I feel it's probably still better to keep using prettier (as opposed to excluding this file or directory for prettier) -- given that the issue is tracked and prettier will fix this in the future.

buildSrc/build.gradle.kts
```diff
@@ -2,4 +2,14 @@ import org.gradle.kotlin.dsl.`kotlin-dsl`
...
-repositories { mavenCentral() }
\ No newline at end of file
+repositories {
+  google()
+  gradlePluginPortal()
+  mavenCentral()
+}
+
+dependencies {
+//  implementation(Plugins.kotlinGradlePlugin)
```
jingtang10: remove?

buildSrc/build.gradle.kts
```diff
@@ -2,4 +2,14 @@ import org.gradle.kotlin.dsl.`kotlin-dsl`
...
+  implementation("com.diffplug.spotless:spotless-plugin-gradle:5.11.0")
```
jingtang10, line 12: can you use the dependencies?

buildSrc/src/main/kotlin/SpotlessConfig.kt
jingtang10: no i think this is fine :)

buildSrc/src/main/kotlin/SpotlessConfig.kt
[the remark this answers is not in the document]
jingtang10: can you add a comment here to link to that issue?

jingtang10 (changes requested):
Thanks @s-ayush2903 for this. Just a few small comments.

.idea/jarRepositories.xml
s-ayush2903: https://github.com/prettier/prettier/issues/10657

buildSrc/build.gradle.kts
s-ayush2903: Turns out we can't do it this simply. When I do it, it tells me an unresolved reference and also tells me to make sure if we are using kotlin-gradle-plugin repo in buildSrc. There should be some fancy workaround for it, but that'll require more effort in comparison to that of putting a simple string(the current case). Should I go ahead for that workaround?

buildSrc/src/main/kotlin/SpotlessConfig.kt
jingtang10: if you move this back, can you use the dependencies better in the `build.gradle.kts` file rather than using `implementation("com.diffplug.spotless:spotless-plugin-gradle:5.11.0")`? if that's the case please move this back.

buildSrc/src/main/kotlin/SpotlessConfig.kt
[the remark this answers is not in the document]
jingtang10: reminder to do this

.idea/jarRepositories.xml
jingtang10: hm, so do you think this is a spotless issue? so odd...

s-ayush2903: I'm not sure. The scene is:
1. File is exposed (by spotless) to prettier for applying formatting
2.  And prettier ends up formatting it incorrectly.
Quite difficult(to me) to comment where the problem is

s-ayush2903: The observation to be made here is that prettier does format it but incorrectly. I manually changed formatting of some section of file and then re-ran spotless, again it ended up formatting file the way it was earlier(i.e., incorrectly).

buildSrc/src/main/kotlin/SpotlessConfig.kt
s-ayush2903: Yep, Plugins will be accessible from there. But by doing so, won't we be reducing cohesion?(Just a question out of curiosity, no problem in moving it to gradle file)

s-ayush2903: Nope it won't be possible even then : (
I saw junit's project just to see if they've done something this way, they have some workaround, but even then are unable to achieve any dependency via reference in buildSrc level gradle file, I've checked many projects and all of 'em hardcoded deps in buildSrc level gradle file. Maybe it is so difficult to refer deps by `somefile.object.subobject.dep` in buildSrc level gradle file : (

jingtang10: got it thx.

buildSrc/build.gradle.kts
jingtang10: as i commented above, just move the sptless config back to the top level `build.gradle.kts`

buildSrc/src/main/kotlin/SpotlessConfig.kt
jingtang10: > Yes, should it be moved back to gradle file?

actually just do this. thx!

jingtang10 (changes requested):
thanks @s-ayush2903 please see comments below

buildSrc/src/main/kotlin/SpotlessConfig.kt
s-ayush2903: db7dac7 & no offense, does it mean we can sacrifice cohesion over a single string?

buildSrc/src/main/kotlin/SpotlessConfig.kt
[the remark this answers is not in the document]
s-ayush2903: db7dac7

buildSrc/build.gradle.kts
s-ayush2903: db7dac7

.idea/jarRepositories.xml
jingtang10: i've tested locally and i actually got the formatting to work... so i've no idea what's going on here.

buildSrc/src/main/kotlin/SpotlessConfig.kt
jingtang10: yeah interesting point. i think it's quite nice to have the dependencies properly referenced. and i don't feel having the format config on the top level gradle being a sacrifice of cohesion -- but do you think the opposite is true?

you can still define a function in the top level gradle file (please do that actually).

build.gradle.kts
```diff
@@ -38,6 +40,7 @@ subprojects {
     }
     format("xml") {
       target("**/*.xml")
+      targetExclude("**/build/", ".idea/")
```
jingtang10, line 43: can you try remove this and see if you can format the xml file in .idea

s-ayush2903: I was doing the same thing previously, but added it now due to https://github.com/google/android-fhir/runs/2290241853#step:11:49

jingtang10: have you run `./gradlew sA`?

.idea/jarRepositories.xml
s-ayush2903: This astonishes me, as it neither works on my studio nor @_CI

buildSrc/src/main/kotlin/SpotlessConfig.kt
s-ayush2903: I think it would've been better to have that functionality separate, as it was just a config spec and gradle file is about applying configs to module(s)(not to define and use there), and over this maybe we could have that specific dep as a raw string with some special comment 

Well, that was my point of view on it, I definitely don't know as much you do, so it may sound stupid to you :sweat_smile:

s-ayush2903:
I'm unsure if you're aware of https://github.com/diffplug/spotless/issues/831 or not but you should know that it exists now.

build.gradle.kts
s-ayush2903: Ah no, might've missed it.

s-ayush2903:
> if re-including the two directories works that'd be great.

07cf553 & Btw, Any specific reason for exposing the build and (git)ignored files to formatting or we just want to maintain the inertia?

buildSrc/src/main/kotlin/SpotlessConfig.kt
s-ayush2903: Btw 07cf553

jingtang10:
Yeah you're right there's no need to format files nobody edits.

jingtang10:
Yeah I think putting the spotless config in the kotlin folder is a reasonable thing :) if you feel that's better please do that and let me know when you're ready for this to be merged.

.idea/jarRepositories.xml
```diff
@@ -1,4 +1,4 @@
-<?xml version="1.0" encoding="UTF-8"?>
+<?xml version="1.0" encoding="UTF-8" ?>
```
s-ayush2903, line 1: What should be the destiny of this file? This entire file is incorrectly formatted, should we just ignore it or reformat it correctly(reformatting to desired spec can be done via [prettier playground](https://prettier.io/playground) easily, if needed) and put it in the shelf. Btw it is now ignored(via spotless) and (hopefully) will never come up again for formatting issues.

jingtang10: just format it in android studio i think that's enough

jingtang10: or prettier playground

s-ayush2903: 0ea1030  & [playground](https://prettier.io/playground/#N4Igxg9gdgLgprEAuEAeADgJwgKzmGAAgDc5MBnAS2gF4AdEAFgYD46pDDVIBbdaBESgBDHnHogASnB4R40-lRgRMlOOQDC0AGaUA5gFdMwmNSit2nTqkwy5cALS3FlZZgCebDlesR0p6EIRMQlKABMGEmEAGwNxBjBBY2jIgHovH19-MyDReJBguEjiGLiJAFlhUg4NJJjCZwglFXc0jMzUPwCOQokjFJAo2PyACxgYdHIkVNTGgEYAOh4qhAWVPVTl6oAmNssrVFm7eCc4FzdPfetbWRPG5o92n07swN6GcOLS-JwAIybyAteDwDFBXK1Bukrgcujl3gU8l9hhIAFIAIQBhC0PBBYJg7gaZyarhae28z1h0GhmRpuRCDH6DGptJ8JWRDDGEymM3uJI8Cz+ALWmA2UDgAA8DORUpBYIIjucVGppegDL9opQwKkmeSWYQobrDjd7KdFY9oTZjo5eRcnjDXj08qEIoM2WUGGjKLBjO4UbU5ZgyTSXt06flCkj3SBPd7hL7-fBA5C7VlQ-DGa7vhJOZNpqkcIkAwtfl6YD6gRAeNrkxajrdrUSHpdDca7o2+c3g5THfSQJ9M+yQABxCAQPTRIo13WpuFOhgRgdRkdjidBjrdsN9TADIZRnPc1JhaILPSj8dwCtVo+pYRQMLYcKbFZQbbV-V2o1W03E20W1sNs1O3XB1Nw+F1d3ySpqgUCA1wpED4QXCCKmfGC4PtNM5xADNkI5cZcx5IklmfBZhHQYQwBGC91ifHY3wNZ46xNG0Wg-f9vybFMuA3eF+1wkAtgQdCZzeLCkLdfJBPMKcuwdZk9VA7Dtx1BSaQk7N8IPIVyGgEx3HQaiRRlaB4FgBUfyVdRUnIERJhGORyBU1SGIOJi20AtivxY81p24hCsL49SGCHYwjzgLE6miQgAAVYj0L1CBgpthL8zDe3ErNgtCicItjKLYoMeKOCSjsUpDWdexwoKQH3PN0Dir1AT0bLDI2Hhdhk1z2O8oDDl4fgxVgLxDiwXB8BgLwQAAGhAbtyGQUBhEwbAAHdoqWhB5pQGIVrjeaZt+YwwAAazgGAAGVyLAL09GQMs4hmmRfjgMIwhegAZW9DGEPQ4AAMRUZZxhu5AQGEAxlGmmqYB4aIAHURlcdQrrgc7NtcShiHBUGwHIfa+ygcgyBgaLjD0ZZkG0GIiZmnByHFNEjtOi68ner04Ep6m4Fp+nzpuicAEUDHsTnohpkByIoMhQd+YRnuiKGsFLOHwhgEZkAADgABhm0aibh4x0FBrB1DIUgoYAR2F+BSb8LawfIBwxRel6odsK3KFsUmfoppAqbF7mQCJnhKDuzAHqD-m4CFkW-a5maYDllWwjV5BtgT4xKA1KA9GxX2QHUABWKGpTgAAVOWtv98XiDiABJO9BHOsBVH8ABBO9zvxCdRaJgBfPugA)

jingtang10 (approved):

jingtang10:
Many thanks for your perseverance with the number of comments! Great work! 🎉🎉🎉

s-ayush2903:
Thanks, more the comments, more I get to learn : )

Issue #372, opened 2021-03-22
Implement initial value in questionnaire


jingtang10:
Implement the initial value as specified here:
https://www.hl7.org/fhir/questionnaire-definitions.html#Questionnaire.item.initial

MuhammadSalman-7214:
@jingtang10 I have some questions:
1. what and when we want to show initial value
2. We have initial values as list in the code so how do we want to show specific value
3. For which widgets we want to have initial values

can you please tell the answers? thanks

jingtang10:
Hi @MuhammadSalman-7214 

> @jingtang10 I have some questions:
> 
> 1. what and when we want to show initial value

The initial values should be displayed if there isn't already an answer in the questionnaire response.

From the spec:

```
One or more values that should be pre-populated in the answer when initially rendering the questionnaire for user input.

The user is allowed to change the value and override the default (unless marked as read-only). If the user doesn't change the value, then this initial value will be persisted when the QuestionnaireResponse is initially created. Note that initial values can influence results. The data type of initial[x] must agree with the item.type, and only repeating items can have more then one initial value.
```

> 2. We have initial values as list in the code so how do we want to show specific value

You may have multiple initial values for repeated questions. But I would not worry about that for now. Focus on single answer questions.

> 3. For which widgets we want to have initial values

Probably all of them. But most importantly text inputs, checkboxes, radio buttons, drop down lists.

> 
> can you please tell the answers? thanks

Thanks!

MuhammadSalman-7214:
@jingtang10 initial values list is empty for checkboxes, radio buttons, drop downs, and edit texts as well except one edit text and that edit text too contains initial value as IntegerType[0] which I am not understanding, so how do I set the initial values for all these widgets? kindly advise. Thanks

jingtang10:
If the initial value list is empty that means there isn't an initial value set for the question. So you don't need to do anything.

MuhammadSalman-7214:
so that means we don't have to set initial value for checkboxes, radio buttons, and drop downs? because we don't have any one widget in the application among these that contain initial values.

jingtang10:
https://www.hl7.org/fhir/questionnaire-profile-example-ussg-fht.json.html has a question with intial value.

Can you also ask on the fhir chat for more examples?

I'm not sure what you mean by we don't have to set initial values. It's an optional field, so as I said you don't need to do anything if it's not there. Does this make sense?

Thanks.

MuhammadSalman-7214:
[...]

PR #393, opened 2021-03-26
Initial Values Pre Filled Functionality Added

MuhammadSalman-7214:
**IMPORTANT: All PRs must be linked to an issue (except for extremely trivial and straighforward changes).**

**Fixes** https://github.com/google/android-fhir/issues/372

**Description**
For Check boxes, Radio buttons, Drop downs and Edit texts, if there is any initial value to set, then that value would already be pre filled after this PR

**Alternative(s) considered**
No

**Type**
Feature
Screenshot_1616762185

**Screenshots (if applicable)**

**Checklist**
- [x] I have read and acknowledge the [Code of conduct](https://github.com/google/android-fhir/blob/master/docs/CODE_OF_CONDUCT.md)
- [x] I have read [How to Contribute](https://github.com/google/android-fhir/blob/master/docs/contributing.md)
- [x] I have read the [Developer's guide](https://github.com/google/android-fhir/wiki/Developer's-Guide)
- [x] I have signed the Google [Individual CLA](https://cla.developers.google.com/about/google-individual), or I am covered by my company's [Corporate CLA](https://cla.developers.google.com/about/google-corporate )
- [x] I have discussed my proposed solution with code owners in the linked issue(s) and we have agreed upon the general approach
- [x] I have run `./gradlew spotlessApply` and `./gradlew spotlessCheck` to check my code follows the style guide of this project
- [x] I have run `./gradlew check` and `./gradlew connectedCheck` to test my changes locally
- [x] I have built and run the reference app(s) to verify my change fixes the issue and/or does not break the reference app(s)

Issue #216, opened 2021-02-04
Handle Questionnaire with item repeats to handle questions which can have more than one answer


joiskash:
[Questionnaire.item.repeats](https://www.hl7.org/fhir/questionnaire-definitions.html#Questionnaire.item.repeats) is used to indicate that a particular question / group of questions can have multiple answers. 

> When rendering the questionnaire, it is up to the rendering software whether to render the question text for each answer repetition (i.e. "repeat the question") or to simply allow entry/selection of multiple answers for the question (repeat the answers).
This issue is linked to the following:
- [ ] https://github.com/google/android-fhir/issues/606

epicadk:
can I work on this?

jingtang10:
I believe @joiskash was gonna take this.

joiskash:
Yup that's right :)

joiskash:
There are two main requirements as defined in the [FHIR spec](https://www.hl7.org/fhir/questionnaire-definitions.html#Questionnaire.item.repeats)

1. Multiple answers for questions 
2. Multiple sets of answers for groups.

> If repeats is missing : Items are generally assumed not to repeat unless explicitly specified. Systems SHOULD always populate this value

> If a question is marked as repeats=true, then multiple answers can be provided for the question in the corresponding QuestionnaireResponse. When rendering the questionnaire, it is up to the rendering software whether to render the question text for each answer repetition (i.e. "repeat the question") or to simply allow entry/selection of multiple answers for the question (repeat the answers). Which is most appropriate visually may depend on the type of answer as well as whether there are nested items.

@jingtang10 @fredhersch just wanted your views on the comment above. 

que-6 | Rule | Required and repeat aren't permitted for display items | type!='display' or (required.empty() and repeats.empty())

jingtang10:
I didn't know that a group can be repeated too. Interesting. It makes sense. I guess that could be handled in a separate PR or even in a separate issue if you prefer.

About the text you quoted, I think it depends on type of answer and whether nested items exist. For example, multiple choices questions with checkboxes probably don't need to repeat the question. But questions with nested items might need to have the question repeated to make it clear. We can probably depriorise this case (repeating the question) for now and focus on repeated questions without nested items.

Would this be enough as a starting point? 

@fredhersch we might benefit from some mocks / UX support for this one.

joiskash:
ok got it. Yea that makes sense and is a good starting point for this PR. Thanks @jingtang10

joiskash:
[...]

PR #417, opened 2021-04-01
Ms/feature initial values

MuhammadSalman-7214:

Fixes https://github.com/google/android-fhir/issues/372 

**Description**
Adding on the functionality of https://github.com/google/android-fhir/pull/393 in which we were setting the initial values without any checks and violations, this PR caters all rules in place for the initial values and throws exception if there is any rule break for initial values

**Alternative(s) considered**
Yes considered the alternative approach of changing the factories, but the approach of Viewmodel (current) is better as discussed?

**Type**
Choose one: (Feature)

**Checklist**


MuhammadSalman-7214:
@jingtang10 and @joiskash please check the updated checks now. Also updating the test cases

datacapture/src/main/java/com/google/android/fhir/datacapture/QuestionnaireViewModel.kt
```diff
@@ -176,13 +176,25 @@ private fun Questionnaire.QuestionnaireItemComponent.createQuestionnaireResponse
...
       answer =
         mutableListOf(
           QuestionnaireResponse.QuestionnaireResponseItemAnswerComponent().apply {
             value = this@createQuestionnaireResponseItem.initial[0].value
           }
         )
+    } else if (initial.isNotEmpty() &&
+        this@createQuestionnaireResponseItem.type == Questionnaire.QuestionnaireItemType.GROUP
+    ) {
+      throw java.lang.IllegalArgumentException("GROUP type can't have initial value")
```
joiskash, line 192: Change the error statement to  "Questionnaire item ${questionnaireItemListIterator.next().linkId} breaks the rule (type!='group' and type!='display') or initial.empty()". Same below. And also add a condition for type display

joiskash (changes requested):

datacapture/src/main/java/com/google/android/fhir/datacapture/QuestionnaireViewModel.kt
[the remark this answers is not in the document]
MuhammadSalman-7214: Done

MuhammadSalman-7214: Done

jingtang10:
please update your PR description to link to the issue properly - thanks!

MuhammadSalman-7214:
> please update your PR description to link to the issue properly - thanks!

Already done that, thanks

datacapture/src/main/java/com/google/android/fhir/datacapture/QuestionnaireViewModel.kt
```diff
@@ -176,13 +176,37 @@ private fun Questionnaire.QuestionnaireItemComponent.createQuestionnaireResponse
   QuestionnaireResponse.QuestionnaireResponseItemComponent {
   return QuestionnaireResponse.QuestionnaireResponseItemComponent().apply {
     linkId = this@createQuestionnaireResponseItem.linkId
-    if (this@createQuestionnaireResponseItem.initial.isNotEmpty()) {
+    if (this@createQuestionnaireResponseItem.type != Questionnaire.QuestionnaireItemType.GROUP &&
```
jingtang10, line 179: I find this entire block to be difficult to read. Can you try to improve it? For example, seems like the whole block should be inside 
```
if (this@createQuestionnaireResponseItem.initial.isNotEmpty()) {
  ...
}
```

You can also use early returns, or create utility functions.

datacapture/src/main/java/com/google/android/fhir/datacapture/QuestionnaireViewModel.kt
```diff
@@ -176,13 +176,37 @@ private fun Questionnaire.QuestionnaireItemComponent.createQuestionnaireResponse
...
         mutableListOf(
           QuestionnaireResponse.QuestionnaireResponseItemAnswerComponent().apply {
             value = this@createQuestionnaireResponseItem.initial[0].value
           }
         )
+    } else if (initial.isNotEmpty() &&
+        this@createQuestionnaireResponseItem.type == Questionnaire.QuestionnaireItemType.GROUP
+    ) {
+      throw IllegalArgumentException(
+        "Questionnaire item ${this@createQuestionnaireResponseItem.linkId} breaks the rule (type!='group' and type!='display') or initial.empty()"
```
jingtang10, line 194: what is "the rule"? is it written in FHIR somewhere? I'm not seeing it in https://www.hl7.org/fhir/questionnaire.html#invs.

if it's our rule, then just use plain language to explain it.

jingtang10 (changes requested):

datacapture/src/main/java/com/google/android/fhir/datacapture/QuestionnaireViewModel.kt
MuhammadSalman-7214: okay

datacapture/src/main/java/com/google/android/fhir/datacapture/QuestionnaireViewModel.kt
MuhammadSalman-7214: https://www.hl7.org/fhir/questionnaire-definitions.html#Questionnaire.item.initial

@joiskash is saying to write the way I have written it now, previously I tried to wrote in simple words

joiskash: @jingtang10  yes this "rule" is found in the description of the questionnaire.item.initial .It is not our rule, it is a rule from the FHIR spec. You will find it in the link posted by Muhammad

MuhammadSalman-7214: so @jingtang10 should the exception be like this or is it need to be changed?

MuhammadSalman-7214:
@Test
  fun questionnareHasMoreThanOneInitialValues_shouldThrowError() {
    val questionnaire =
      Questionnaire().apply {
        id = "a-questionnaire"
        addItem(
          Questionnaire.QuestionnaireItemComponent().apply {
            linkId = "a-link-id"
            text = "Basic question"
            type = Questionnaire.QuestionnaireItemType.BOOLEAN
            initial =
              mutableListOf(
                Questionnaire.QuestionnaireItemInitialComponent().setValue(BooleanType(true)),
                Questionnaire.QuestionnaireItemInitialComponent().setValue(BooleanType(true))
              )
          }
        )
      }
    val serializedQuestionniare = printer.encodeResourceToString(questionnaire)
    state.set(QuestionnaireFragment.BUNDLE_KEY_QUESTIONNAIRE, serializedQuestionniare)
    
    val errorMessage =
      assertFailsWith<IllegalArgumentException> { QuestionnaireViewModel(state) }.localizedMessage

    assertThat(errorMessage)
      .isEqualTo("Questionnaire item a-link-id have more than one initial value which is not allowed")
  }
  
  @joiskash I have written this test case to verify the code is throwing exception when there are more than one initial values, it is throwing right exception, but it is not asserting. Can you please help me out in this? Thanks.

datacapture/src/main/java/com/google/android/fhir/datacapture/QuestionnaireViewModel.kt
jingtang10: Questionnaire item ${this@createQuestionnaireResponseItem.linkId} has initial value(s) and is a group or display item. See rule que-8 at https://www.hl7.org/fhir/questionnaire-definitions.html#Questionnaire.item.initial.

joiskash:
There are some tests in the view model that say "shouldThrowError" have a look at them and then add similar tests. So basically you are asserting that a proper error message is thrown in the unit test.

MuhammadSalman-7214:
@joiskash the tests in the view model are for the case when we pass both Questionnaire and Questionnaire response together, but we are testing for the other case in which we are only passing question. This test is not there in ViewModel Test class as well, so that is why need a bit help in this. Thanks.

MuhammadSalman-7214:
@jingtang10 and @joiskash, please review, thanks.

MuhammadSalman-7214:
@jingtang10 and @joiskash, please check the PR. Everything has been done. Thanks

datacapture/src/main/java/com/google/android/fhir/datacapture/QuestionnaireViewModel.kt
```diff
@@ -177,12 +177,32 @@ private fun Questionnaire.QuestionnaireItemComponent.createQuestionnaireResponse
...
+            }
+          )
+      } else if (this@createQuestionnaireResponseItem.type ==
+          Questionnaire.QuestionnaireItemType.GROUP ||
+          this@createQuestionnaireResponseItem.type == Questionnaire.QuestionnaireItemType.DISPLAY
+      ) {
+        throw IllegalArgumentException(
+          "Questionnaire item ${this@createQuestionnaireResponseItem.linkId} has initial value(s) and is a group or display item. See rule que-8 at https://www.hl7.org/fhir/questionnaire-definitions.html#Questionnaire.item.initial."
         )
+      } else if (initial.size > 1 &&
```
joiskash, line 198: Please refer to rule 13 mentioned [here ](https://www.hl7.org/fhir/questionnaire-definitions.html#Questionnaire.item.initial)

datacapture/src/test/java/com/google/android/fhir/datacapture/QuestionnaireViewModelTest.kt
```diff
@@ -369,6 +369,135 @@ class QuestionnaireViewModelTest {
...
+    state.set(QuestionnaireFragment.BUNDLE_KEY_QUESTIONNAIRE, serializedQuestionniare)
+    state.set(
+      QuestionnaireFragment.BUNDLE_KEY_QUESTIONNAIRE_RESPONSE,
+      serializedQuestionniareResponse
+    )
+    QuestionnaireViewModel(state)
+  }
+
+  @Test
+  fun questionnareHasMoreThanOneInitialValues_shouldThrowError() {
```
joiskash, line 414: same as above refer to rule 13 [here](https://www.hl7.org/fhir/questionnaire-definitions.html#Questionnaire.item.initial) you need to check  if item.repeats is true or false. 
not item type.

joiskash (dismissed):

datacapture/src/main/java/com/google/android/fhir/datacapture/QuestionnaireViewModel.kt
MuhammadSalman-7214: @joiskash currently we are handling only one initial value, not more than one. That is the reason of the check.

datacapture/src/test/java/com/google/android/fhir/datacapture/QuestionnaireViewModelTest.kt
MuhammadSalman-7214: @joiskash currently we are handling only one initial value, not more than one. That is the reason for this test.

joiskash: I think we should test the rule 13. If there is more than one initial item and repeats == false then we should throw and error. If there is more than one initial item and repeats == true we should not throw an error and should just use the first initial value. Later on when #216

joiskash: We do not want to reject a valid questionnaire just because we are unable to handle repeats. That is it what I feel. @jingtang10 please comment.

datacapture/src/main/java/com/google/android/fhir/datacapture/QuestionnaireViewModel.kt
MuhammadSalman-7214: @joiskash check updated as per the rule, thanks

datacapture/src/test/java/com/google/android/fhir/datacapture/QuestionnaireViewModelTest.kt
MuhammadSalman-7214: @joiskash test case updated as per the rule, thanks.

MuhammadSalman-7214: yes @joiskash, updated the PR as per the rules and as you are suggesting.

joiskash: Can you re-request review? also one small suggestion. Do you think you can move the if-else logic into a function? basically 3 if else statements will be there

1. rule 8 i.e. item (type!='group' and type!='display') or initial.empty()
2. rule 13 i.e. item repeats=true or initial.count() <= 1
3. happy scenario else for all other conditions . 

This way the code will be very readable and will be directly related to the rules defined in the spec. You can directly use these logical statements as defined.

MuhammadSalman-7214: @joiskash Sure, I can re request review and move all the if else logic to a seprate function, but I don't think we shouldn't change the format of checks as every check has been verified with unit tests. Also all the checks are as per the rules, though the format of check can be different so I don't think we should change the rules again.

MuhammadSalman-7214: Screenshot from 2021-04-06 13-49-29
@joiskash, when creating a function, it is not picking answer or setAnswe. Can you tell me the reason why is it not picking it this way?

joiskash:
```
private fun Questionnaire.QuestionnaireItemComponent.createAnswerForItemWithInitialValue():  
MutableList<QuestionnaireResponse.QuestionnaireResponseItemAnswerComponent>? {
  if (valid condition) {
  ....
 return  mutableListOf(
   ...
  ) }
 else error:
  return null
}
```

try this syntax and call the function and set answer with the returned value

joiskash: create an extension function for Questionnaire.QuestionnaireItemComponent

MuhammadSalman-7214:
@joiskash, I have updated the PR, please check, thanks.

datacapture/src/test/java/com/google/android/fhir/datacapture/QuestionnaireViewModelTest.kt
MuhammadSalman-7214: @joiskash, please check the PR, thanks.

joiskash (commented):
Look fine to me except for my comment about restructuring the if-else to match the rules in the spec exactly with rule 8 and rule 13. Currently you only refer to rule 8 in the error message. @jingtang10 can take a decision on this

MuhammadSalman-7214:
> Look fine to me except for my comment about restructuring the if-else to match the rules in the spec exactly with rule 8 and rule 13. Currently you only refer to rule 8 in the error message. @jingtang10 can take a decision on this

@joiskash rule 13 is covering as well in case for repeat true, we are setting the value in that case

joiskash:
Yes I am just talking about the error message. Check both error message, both refer to rule 8. You don't refer to rule 13 anywhere even though you are checking for it.

MuhammadSalman-7214:
Well I can add another check in case for throwing exception for rule 13

MuhammadSalman-7214:
@joiskash, I have checked all the rules and found out the exception of rule 13 is covering as well. It's just that the exception wording can be changed only to "Item have more than one Initial Values and is not Repeating Item", something like that. What do you suggest about the text of the exception? as every check is covering already

joiskash:
I think "Questionnaire item ${this@createQuestionnaireResponseWithInitialItem.linkId} Can only have multiple initial values for repeating items See rule que-13 at https://www.hl7.org/fhir/questionnaire-definitions.html#Questionnaire.item.initial."

datacapture/src/main/java/com/google/android/fhir/datacapture/QuestionnaireViewModel.kt
```diff
@@ -176,20 +176,62 @@ private fun Questionnaire.QuestionnaireItemComponent.createQuestionnaireResponse
...
+          value = this@createQuestionnaireResponseWithInitialItem.initial[0].value
+        }
+      )
+    } else if (this@createQuestionnaireResponseWithInitialItem.type ==
+        Questionnaire.QuestionnaireItemType.GROUP ||
+        this@createQuestionnaireResponseWithInitialItem.type ==
+          Questionnaire.QuestionnaireItemType.DISPLAY
+    ) {
+      throw IllegalArgumentException(
+        "Questionnaire item ${this@createQuestionnaireResponseWithInitialItem.linkId} has initial value(s) and is a group or display item. See rule que-8 at https://www.hl7.org/fhir/questionnaire-definitions.html#Questionnaire.item.initial."
```
jingtang10, line 218: move this to the beginning of this function and throw exception.

i.e.

```
if (is group or is display and has initial value) {
  throw exception
}
```

datacapture/src/main/java/com/google/android/fhir/datacapture/QuestionnaireViewModel.kt
```diff
@@ -176,20 +176,62 @@ private fun Questionnaire.QuestionnaireItemComponent.createQuestionnaireResponse
...
+    answer = createQuestionnaireResponseWithInitialItem()
     this@createQuestionnaireResponseItem.item.forEach {
       this.addItem(it.createQuestionnaireResponseItem())
     }
   }
 }
 
+private fun Questionnaire.QuestionnaireItemComponent.createQuestionnaireResponseWithInitialItem():
+  MutableList<QuestionnaireResponse.QuestionnaireResponseItemAnswerComponent>? {
+  if (this@createQuestionnaireResponseWithInitialItem.initial.isNotEmpty()) {
```
jingtang10, line 188: this is too long -- think of ways to make this block more readable. right now it's very difficult to figure out what this if block is doing.

use a combination of 1) creating utility functions 2) early return / throw 3) reorganising branches logically

datacapture/src/main/java/com/google/android/fhir/datacapture/QuestionnaireViewModel.kt
```diff
@@ -176,20 +176,62 @@ private fun Questionnaire.QuestionnaireItemComponent.createQuestionnaireResponse
...
+        !this@createQuestionnaireResponseWithInitialItem.repeats &&
+        this@createQuestionnaireResponseWithInitialItem.type !=
+          Questionnaire.QuestionnaireItemType.GROUP &&
+        this@createQuestionnaireResponseWithInitialItem.type !=
+          Questionnaire.QuestionnaireItemType.DISPLAY
+    ) {
+      throw IllegalArgumentException(
+        "Questionnaire item ${this@createQuestionnaireResponseWithInitialItem.linkId} has initial value(s). See rule que-8 at https://www.hl7.org/fhir/questionnaire-definitions.html#Questionnaire.item.initial."
+      )
+    }
```
jingtang10, line 230: same here. do an early return.

datacapture/src/main/java/com/google/android/fhir/datacapture/QuestionnaireViewModel.kt
jingtang10, line 188: do you need this at all? 
```suggestion
  if (initial.isNotEmpty()) {
```

datacapture/src/main/java/com/google/android/fhir/datacapture/QuestionnaireViewModel.kt
```diff
@@ -176,20 +176,62 @@ private fun Questionnaire.QuestionnaireItemComponent.createQuestionnaireResponse
...
     this@createQuestionnaireResponseItem.item.forEach {
       this.addItem(it.createQuestionnaireResponseItem())
     }
   }
 }
 
+private fun Questionnaire.QuestionnaireItemComponent.createQuestionnaireResponseWithInitialItem():
+  MutableList<QuestionnaireResponse.QuestionnaireResponseItemAnswerComponent>? {
+  if (this@createQuestionnaireResponseWithInitialItem.initial.isNotEmpty()) {
+    if (this@createQuestionnaireResponseWithInitialItem.type !=
```
jingtang10, line 189:
```suggestion
    if (type !=
```

datacapture/src/main/java/com/google/android/fhir/datacapture/QuestionnaireViewModel.kt
```diff
@@ -176,20 +176,62 @@ private fun Questionnaire.QuestionnaireItemComponent.createQuestionnaireResponse
...
-        )
-    }
+    answer = createQuestionnaireResponseWithInitialItem()
     this@createQuestionnaireResponseItem.item.forEach {
       this.addItem(it.createQuestionnaireResponseItem())
     }
   }
 }
 
+private fun Questionnaire.QuestionnaireItemComponent.createQuestionnaireResponseWithInitialItem():
```
jingtang10, line 186: this function name is a bit of a misnomer. it's creating a list of answers rather than the questionnaire response item.

```suggestion
private fun Questionnaire.QuestionnaireItemComponent.createQuestionnaireResponseItemAnswers():
```

and add documentation to this function to say you are using inital items.

MuhammadSalman-7214:
> I think "Questionnaire item ${[this@createQuestionnaireResponseWithInitialItem.linkId](mailto:this@createQuestionnaireResponseWithInitialItem.linkId)} Can only have multiple initial values for repeating items See rule que-13 at https://www.hl7.org/fhir/questionnaire-definitions.html#Questionnaire.item.initial."

yes that rule and the exception has already been handled

datacapture/src/main/java/com/google/android/fhir/datacapture/QuestionnaireViewModel.kt
```diff
@@ -176,20 +176,62 @@ private fun Questionnaire.QuestionnaireItemComponent.createQuestionnaireResponse
...
+        "Questionnaire item ${this@createQuestionnaireResponseWithInitialItem.linkId} has initial value(s) and is a group or display item. See rule que-8 at https://www.hl7.org/fhir/questionnaire-definitions.html#Questionnaire.item.initial."
+      )
+    } else if (initial.size > 1 &&
+        !this@createQuestionnaireResponseWithInitialItem.repeats &&
+        this@createQuestionnaireResponseWithInitialItem.type !=
+          Questionnaire.QuestionnaireItemType.GROUP &&
+        this@createQuestionnaireResponseWithInitialItem.type !=
+          Questionnaire.QuestionnaireItemType.DISPLAY
+    ) {
+      throw IllegalArgumentException(
```
jingtang10, line 227: your error msg is wrong here. this is breaking que-13

datacapture/src/main/java/com/google/android/fhir/datacapture/QuestionnaireViewModel.kt
```diff
@@ -176,20 +176,62 @@ private fun Questionnaire.QuestionnaireItemComponent.createQuestionnaireResponse
...
+        this@createQuestionnaireResponseWithInitialItem.type !=
+          Questionnaire.QuestionnaireItemType.DISPLAY &&
+        this@createQuestionnaireResponseWithInitialItem.repeats &&
+        initial.size > 1
+    ) {
+      return mutableListOf(
+        QuestionnaireResponse.QuestionnaireResponseItemAnswerComponent().apply {
+          value = this@createQuestionnaireResponseWithInitialItem.initial[0].value
+        }
+      )
```
jingtang10, line 211: this seems wrong? if there're multiple intial values why do you only copy one?

datacapture/src/main/java/com/google/android/fhir/datacapture/QuestionnaireViewModel.kt
```diff
@@ -176,20 +176,62 @@ private fun Questionnaire.QuestionnaireItemComponent.createQuestionnaireResponse
...
+          Questionnaire.QuestionnaireItemType.GROUP &&
+        this@createQuestionnaireResponseWithInitialItem.type !=
+          Questionnaire.QuestionnaireItemType.DISPLAY
+    ) {
+      throw IllegalArgumentException(
+        "Questionnaire item ${this@createQuestionnaireResponseWithInitialItem.linkId} has initial value(s). See rule que-8 at https://www.hl7.org/fhir/questionnaire-definitions.html#Questionnaire.item.initial."
+      )
+    }
+  }
+  return null
```
jingtang10, line 232: Again, this can be made into an early return.

```
if (no initial value) {
  return null
}
```

this should probably be the very first thing in this function

jingtang10 (changes requested):
thanks @MuhammadSalman-7214

datacapture/src/main/java/com/google/android/fhir/datacapture/QuestionnaireViewModel.kt
jingtang10: it seems like @joiskash had pointed this out. can you change this please?

datacapture/src/main/java/com/google/android/fhir/datacapture/QuestionnaireViewModel.kt
MuhammadSalman-7214: because at the moment, we are only catering the first initial value for repeat true case

datacapture/src/main/java/com/google/android/fhir/datacapture/QuestionnaireViewModel.kt
MuhammadSalman-7214: sure

datacapture/src/main/java/com/google/android/fhir/datacapture/QuestionnaireViewModel.kt
jingtang10: thanks for the explanation. please create a follow up issue and add a comment here in that case.

datacapture/src/main/java/com/google/android/fhir/datacapture/QuestionnaireViewModel.kt
MuhammadSalman-7214: Done

datacapture/src/main/java/com/google/android/fhir/datacapture/QuestionnaireViewModel.kt
MuhammadSalman-7214: Done

datacapture/src/main/java/com/google/android/fhir/datacapture/QuestionnaireViewModel.kt
MuhammadSalman-7214: Done

datacapture/src/main/java/com/google/android/fhir/datacapture/QuestionnaireViewModel.kt
MuhammadSalman-7214: Yes and done

datacapture/src/main/java/com/google/android/fhir/datacapture/QuestionnaireViewModel.kt
MuhammadSalman-7214: Done

datacapture/src/main/java/com/google/android/fhir/datacapture/QuestionnaireViewModel.kt
MuhammadSalman-7214: Done

datacapture/src/main/java/com/google/android/fhir/datacapture/QuestionnaireViewModel.kt
MuhammadSalman-7214: Updated

datacapture/src/main/java/com/google/android/fhir/datacapture/QuestionnaireViewModel.kt
MuhammadSalman-7214: Done

datacapture/src/main/java/com/google/android/fhir/datacapture/QuestionnaireViewModel.kt
MuhammadSalman-7214: okay, will create an issue for this

jingtang10 (changes requested):

datacapture/src/main/java/com/google/android/fhir/datacapture/QuestionnaireViewModel.kt
```diff
@@ -176,20 +176,57 @@ private fun Questionnaire.QuestionnaireItemComponent.createQuestionnaireResponse
...
+        initial.size > 1
+    ) {
+      return mutableListOf(
+        QuestionnaireResponse.QuestionnaireResponseItemAnswerComponent().apply {
+          value = initial[0].value
+        }
+      )
+    }
+  }
+  return null
```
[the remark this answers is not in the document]
MuhammadSalman-7214: Thanks @jingtang10 for the article. It helped in understanding the concept.

jingtang10 (changes requested):
please update the PR title to something readable.

also please add in PR description what's the added functionality from https://github.com/google/android-fhir/pull/393.

datacapture/src/main/java/com/google/android/fhir/datacapture/QuestionnaireViewModel.kt
```diff
@@ -176,20 +176,49 @@ private fun Questionnaire.QuestionnaireItemComponent.createQuestionnaireResponse
...
-        )
-    }
+    answer = createQuestionnaireResponseItemAnswers()
     this@createQuestionnaireResponseItem.item.forEach {
       this.addItem(it.createQuestionnaireResponseItem())
     }
   }
 }
 
+/** Here we are setting initial value as an Answer to a Question */
```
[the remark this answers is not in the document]
MuhammadSalman-7214: Done

datacapture/src/main/java/com/google/android/fhir/datacapture/QuestionnaireViewModel.kt
```diff
@@ -176,20 +176,49 @@ private fun Questionnaire.QuestionnaireItemComponent.createQuestionnaireResponse
...
+    )
+  }
+
+  if (initial.size == 1) {
+    return mutableListOf(
+      QuestionnaireResponse.QuestionnaireResponseItemAnswerComponent().apply {
+        value = initial[0].value
+      }
+    )
+  }
```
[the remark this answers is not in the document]
MuhammadSalman-7214: Done

MuhammadSalman-7214:
> please update the PR title to something readable.
> 
> also please add in PR description what's the added functionality from #393.

Done

jingtang10 (approved):

PR #419, opened 2021-04-02
Add `DateTimeType` to search

### LINKED ISSUE #375: Support date and date time in search

epicadk:

Fixes #375 

**Description**

Add date to search 

**Alternative(s) considered**

Followed a similar approach to that of string and reference  

**Type**
Choose one: Feature 

**Checklist**


jingtang10:
Well done in figuring out how the index tables work.

epicadk:
> Well done in figuring out how the index tables work.

Thankyou 😄. I'm not sure how I should handle the case when the Prefix is null should we throw an error or have a default value?

core/src/main/java/com/google/android/fhir/search/MoreSearch.kt
```diff
@@ -101,6 +105,30 @@ fun ReferenceFilter.query(type: ResourceType): SearchQuery {
...
+      ParamPrefixEnum.APPROXIMATE -> "BETWEEN index_tsLow AND index_tsHigh"
+      ParamPrefixEnum.STARTS_AFTER -> " < index_tsLow"
+      ParamPrefixEnum.ENDS_BEFORE -> " > index_tsHigh"
+      ParamPrefixEnum.NOT_EQUAL -> "NOT BETWEEN index_tsLow AND index_tsHigh"
+      ParamPrefixEnum.EQUAL -> "BETWEEN index_tsLow AND index_tsHigh"
+      ParamPrefixEnum.GREATERTHAN -> " > index_tsLow"
+      ParamPrefixEnum.GREATERTHAN_OR_EQUALS -> " >= index_tsLow"
+      ParamPrefixEnum.LESSTHAN -> "< index_tsHigh"
+      ParamPrefixEnum.LESSTHAN_OR_EQUALS -> "<= index_tsHigh"
+      null -> "" // Possibly throw an error or provide a default value?
```
jingtang10, line 120: have you tested these?

core/src/test/java/com/google/android/fhir/search/SearchTest.kt
```diff
@@ -119,6 +119,43 @@ class SearchTest {
...
+    assertThat(query.query)
+      .isEqualTo(
+        """
+    SELECT a.serializedResource
+    FROM ResourceEntity a
+    WHERE a.resourceType = ?
+    AND a.resourceId IN (
+    SELECT resourceId form DateIndexEntity
+    WHERE resourceType = ? AND index_name = ?
+    AND ? BETWEEN index_tsLow AND index_tsHigh
```
jingtang10, line 143: does this even work?

epicadk: Ahh I see I have misspelled `from`. I 'll correct it. Apart from that is does work. I tested it using FhirEngineImpl but I am not sure that is where I should add the tests.

epicadk:
@jingtang10 I've added tests for the SQL query. But I think we should resolve #433 before we merge this.

epicadk:
> @jingtang10 I've added a test for the SQL query. But I think we should resolve #433 before we merge this. Converting to draft will add tests for all the cases. I just wanted to know whether I should continue adding the tests in `FhirEngineImplTest` or should I add them to `DatabaseImplTest` ?

Hi, can I have some help here?

jingtang10:
Hi @epicadk sorry for the delay. I approved and merged #433. Can you mark this draft as ready for review? thanks

epicadk:
> Hi @epicadk sorry for the delay. I approved and merged #433. Can you mark this draft as ready for review? thanks

Marked as ready to review. But I haven't really added tests. I'm having trouble in figuring out where I should test if the query works as intended in the `FHIRengineImpl` test or the `DatabaseImpl` test?

epicadk:
Added an empty commit because build test was failing here but seemed to be working fine locally.

epicadk:
PR ready for review

engine/src/main/java/com/google/android/fhir/search/MoreSearch.kt
```diff
@@ -119,6 +121,66 @@ fun ReferenceFilter.query(type: ResourceType): SearchQuery {
   )
 }
 
+fun DateFilter.query(type: ResourceType): SearchQuery {
```
jingtang10, line 124: to make this PR simpler, i think we could restrict the precision field a little bit. For example, I would think having precision year, month, or day are rare for date time data type (see comment below).

So you can do something like `check(precision == MINUTE || ... SECOND || ... MILLISECOND)`

engine/src/main/java/com/google/android/fhir/search/MoreSearch.kt
```diff
@@ -119,6 +121,66 @@ fun ReferenceFilter.query(type: ResourceType): SearchQuery {
...
+  val tsHigh = value!!.precision.add(value!!.value, 1).time
+  var useHigh = false
```
jingtang10, line 126: I would much rather just create a var that is null and assign the right value to it instead of using this boolean flag.

engine/src/main/java/com/google/android/fhir/search/MoreSearch.kt
```diff
@@ -119,6 +121,66 @@ fun ReferenceFilter.query(type: ResourceType): SearchQuery {
...
+              AND index_from BETWEEN ? AND ? AND index_to BETWEEN ? AND ?""",
+          listOf(
+            type.name,
+            parameter.paramName,
+            value!!.value.time,
+            tsHigh - 1,
+            value!!.value.time,
+            tsHigh - 1
+          )
+        )
```
jingtang10, line 168: I think the meaning of equal and not equal can be interpreted in many ways -- and the standard doesn't really help here because it's kinda vague. but i'm happy for us to take a simple approach and use a very strict definition of equal and not equal for date time: literally requiring the start and end timestamps to be both equal. what do you think?

jingtang10 (commented):
some comments - sorry for the delay!

engine/src/main/java/com/google/android/fhir/search/MoreSearch.kt
epicadk: It says in the docs that `YYYY`, `YYYY-MM`, `YYYY-MM-DD` but yeah I think that a user would use Date rather than DateTime for these cases. Should I still add these constrains?

engine/src/main/java/com/google/android/fhir/search/MoreSearch.kt
epicadk: I think this implementation makes sense to have this implementation. 

For example if I wanted to search for an event that occurred on `2001-01-01` I would expect an event that occurred at `2001-01-01T13:10` to be included in the results as well.

engine/src/main/java/com/google/android/fhir/search/MoreSearch.kt
```diff
@@ -123,6 +125,58 @@ fun ReferenceFilter.query(type: ResourceType): SearchQuery {
...
+      ParamPrefixEnum.LESSTHAN_OR_EQUALS ->
+        ">= index_to".also { tsHigh = value!!.precision.add(value!!.value, 1).time }
+    }
+  return SearchQuery(
+    """
+    SELECT resourceId FROM DateIndexEntity 
+    WHERE resourceType = ? AND index_name = ? AND ? $condition 
+    """,
+    listOf(type.name, parameter.paramName, tsHigh ?: value!!.value.time)
+  )
```
jingtang10, line 177: I find the structure here a bit difficult to read, especially given the early returns you have in some (but not all) of the branches in the `when` block. can you create another helper function that returns a condition string as well as a list of params which you will concat to the end of the param list?

jingtang10 (changes requested):

Tarun-Bhardwaj:
Hi @epicadk, could you please provide an ETA for completion of changes mentioned in the PR?

epicadk:
> Hi @epicadk, could you please provide an ETA for completion of changes mentioned in the PR?

I think this and #436 are pretty similar. Once that is approved and ready to merge I'll make similar changes to this pr. I'd say we can get this done in the next 4-5 days

Tarun-Bhardwaj:
@aditya-07 , could you please review the changes?

engine/src/main/java/com/google/android/fhir/search/MoreSearch.kt
```diff
@@ -169,6 +183,39 @@ val Order?.sqlString: String
...
+        listOf(
+          value.value.time,
+          value.precision.add(value.value, 1).time - 1,
+          value.value.time,
+          value.precision.add(value.value, 1).time - 1
+        )
+    ParamPrefixEnum.GREATERTHAN ->
+      "? <= index_from" to listOf(value.precision.add(value.value, 1).time)
+    ParamPrefixEnum.GREATERTHAN_OR_EQUALS -> "? <= index_from" to listOf(value.value.time)
+    ParamPrefixEnum.LESSTHAN -> "? >= index_to" to listOf(value.value.time)
```
jingtang10, line 214:
```suggestion
    ParamPrefixEnum.LESSTHAN -> "? >= index_from" to listOf(value.value.time)
```

because: `[parameter]=lt2013-01-14T10:00 | 2013-01-14 matches, because it includes the part of 14-Jan 2013 before 10am`

what this says to me is that if any part of the range is less than the given value, the range is considered less than the given value.

same for le

engine/src/main/java/com/google/android/fhir/search/MoreSearch.kt
```diff
@@ -169,6 +183,39 @@ val Order?.sqlString: String
...
+    ParamPrefixEnum.EQUAL ->
+      "index_from BETWEEN ? AND ? AND index_to BETWEEN ? AND ?" to
+        listOf(
+          value.value.time,
+          value.precision.add(value.value, 1).time - 1,
+          value.value.time,
+          value.precision.add(value.value, 1).time - 1
+        )
+    ParamPrefixEnum.GREATERTHAN ->
+      "? <= index_from" to listOf(value.precision.add(value.value, 1).time)
```
jingtang10, line 212:
```suggestion
      "? <= index_to" to listOf(value.precision.add(value.value, 1).time)
```

see this: `[parameter]=gt2013-01-14T10:00 | 2013-01-14 matches, because it includes the part of 14-Jan 2013 after 10am`

what this says to me is that if any part of the range is greater than the given value, the range is considered greater than the given value.

same for ge

engine/src/main/java/com/google/android/fhir/search/MoreSearch.kt
```diff
@@ -169,6 +183,39 @@ val Order?.sqlString: String
...
     }
 
+private fun getConditionParamPair(
+  prefix: ParamPrefixEnum,
+  value: DateTimeType
+): Pair<String, List<Any>> {
+  return when (prefix) {
+    ParamPrefixEnum.APPROXIMATE -> TODO("Not Implemented")
+    ParamPrefixEnum.STARTS_AFTER ->
+      "? <= index_from" to listOf(value.precision.add(value.value, 1).time)
```
jingtang10, line 193: can you change the order of things a little bit I find this hard to read...

`index_from >= ?` is easier to read for me than `? <= index_from` because it follows the prefix names such as start after, greater than etc.

same for the rest

jingtang10 (changes requested):

engine/src/main/java/com/google/android/fhir/search/MoreSearch.kt
epicadk: Ahh yes that's how this and ends before are different.

engine/src/main/java/com/google/android/fhir/search/MoreSearch.kt
```diff
@@ -230,3 +277,14 @@ private fun BigDecimal.getRange(): BigDecimal {
...
+/**
+ * Returns the upper bound of the range of the Date's epoch Timestamp. The value is related to the
+ * precision of the DateTimeType
+ *
+ * For example 2001-01-01 includes all values on the given day and thus this functions will return
+ * 978373799 ( which is one second less than the epoch of 2001-01-02)
+ */
+private fun DateTimeType.getTsHigh(): Long = precision.add(value, 1).time - 1
+
+private data class ConditionParam<T>(val condition: String, val params: List<T>)
```
jingtang10, line 290: is it possible to do something like this: https://kotlinlang.org/docs/functions.html#variable-number-of-arguments-varargs

or create a new function? so that you don't need to use `listOf` repeatedly above?

jingtang10 (changes requested):

engine/src/main/java/com/google/android/fhir/search/MoreSearch.kt
epicadk: I think data classes can only have var/val in their primary constructor, I'll define another one.

engine/src/test/java/com/google/android/fhir/search/SearchTest.kt
```diff
@@ -537,6 +538,300 @@ class SearchTest {
...
+    assertThat(query.query)
+      .isEqualTo(
+        """
+    SELECT a.serializedResource
+    FROM ResourceEntity a
+    WHERE a.resourceType = ?
+    AND a.resourceId IN (
+    SELECT resourceId FROM DateIndexEntity
+    WHERE resourceType = ? AND index_name = ? AND index_from >= ?
+    )
```
jingtang10, line 562: add 4 spaces to align slightly better - same below

jingtang10 (approved):
Awesome! 😎

Thanks for the PR!

engine/src/androidTest/java/com/google/android/fhir/db/impl/DatabaseImplTest.kt
```diff
@@ -964,6 +965,357 @@ class DatabaseImplTest {
...
+        deceased = DateTimeType("2013-03-23T10:00:00-05:30")
+      }
+    database.insert(patient)
+    val result =
+      database.search<Patient>(
+        Search(ResourceType.Patient)
+          .apply {
+            filter(Patient.DEATH_DATE) {
+              prefix = ParamPrefixEnum.STARTS_AFTER
+              value = DateTimeType("2013-03-14")
```
deepankarb, line 982: The `DateTimeType` supplied above to `patient` has time zone but not here. Will running the query in different time zones produce consistent results?

Can we add tests for that case or create an issue to add them later?

engine/src/main/java/com/google/android/fhir/search/MoreSearch.kt
```diff
@@ -169,14 +182,46 @@ val Order?.sqlString: String
       null -> ""
     }
 
+private fun getConditionParamPair(
+  prefix: ParamPrefixEnum,
+  value: DateTimeType
+): ConditionParam<Long> {
+  val (start, end) = value.rangeEpochMillis
+  return when (prefix) {
+    ParamPrefixEnum.APPROXIMATE -> TODO("Not Implemented")
```
deepankarb, line 191: Can we add more details here about `ParamPrefixEnum.APPROXIMATE` e.g. link to spec and reason for not implementing.

deepankarb: Also, an issue to track it :)

engine/src/main/java/com/google/android/fhir/search/MoreSearch.kt
```diff
@@ -230,3 +283,18 @@ private fun BigDecimal.getRange(): BigDecimal {
...
+ * the DateTimeType
+ *
+ * For example 2001-01-01 includes all values on the given day and thus this functions will return
+ * 978307200 (epoch timestamp of 2001-01-01) and 978393599 ( which is one second less than the epoch
+ * of 2001-01-02)
+ */
+private val DateTimeType.rangeEpochMillis
+  get() = value.time to precision.add(value, 1).time - 1
+
+private data class ConditionParam<T>(val condition: String, val params: List<T>) {
```
deepankarb, line 298: Can we use [`varargs`](https://kotlinlang.org/docs/functions.html#variable-number-of-arguments-varargs) for params?

deepankarb (approved):
Thanks @epicadk this is cool!
I've left a few non-blocking comments. LGTM, approved :rocket:

engine/src/main/java/com/google/android/fhir/search/MoreSearch.kt
epicadk: We can't use varargs in the primary constructor of a data class. I've created a secondary constructor for that reason. I realize I was still using listOf in the dateTime code, I've updated that now.

engine/src/androidTest/java/com/google/android/fhir/db/impl/DatabaseImplTest.kt
epicadk: Sorry I just did that to show to it is compatible with different timezones. By default the timezone will always be UTC unless specified. I think we can leave it up to the consumer whether they would like to search in the devices timezone of any other timezone?

epicadk: Oh, Hapi will use the local timezone for DateTimeTypes where the timezone isn't specified. I think this could cause issues. We would also have to change the way we index dates and datetimes in order to have more consistent results across timezones.

Even for date types `DateType("2001-01-01").value.time` will return `978287400000` (for me in india) which is actually `December 31, 2000 6:30:00 PM` GMT

epicadk: Thanks, I have opened #567

PR #426, opened 2021-04-05
Support Contains for string search

### LINKED ISSUE #376: Support contains for string search
related to https://github.com/OpenSRP/fhircore/issues/20 but this shoud not block it as the first version would target at simplest use case with exact match.

See spec https://www.hl7.org/fhir/search.html#modifiers

epicadk:

Fixes #376 
**Description**

Support contains for string search.

**Alternative(s) considered**

Explained in the comment on linked issue.  I've removed the ParamPrefixEnum as a parameter of string filter because those params are only meant for Number Quantity and date parameters 
refer https://www.hl7.org/fhir/search.html#prefix or the ParamPrefixEnum Javadoc

I also cannot find a `StringParamModifier` Enum like the `TokenParamModifier` and `UriParamQualifierEnum` Enums that's why I used the string param class. I think we could possibly create our own implementation of the `StringParamModifier` class. 

The tests I have added to the FHIREngineImpl are from https://www.hl7.org/fhir/search.html#string

**Type**
Choose one: Feature

**Checklist**


engine/src/main/java/com/google/android/fhir/search/SearchDsl.kt
```diff
@@ -56,7 +56,7 @@ data class Search(val type: ResourceType, var count: Int? = null, var from: Int?
 @SearchDslMarker
 data class StringFilter(
   val parameter: StringClientParam,
-  var prefix: ParamPrefixEnum? = null,
+  var modifier: StringParam? = null,
```
jingtang10, line 59: I'm not certain this is the right way to represent this. Looking at the API, you have something like this: 

```
filter(Patient.GIVEN) {
  value = "eve"
  modifier = StringParam().setContains(true)
}
```

I would say it looks a little clumsy. can we just create a new enum? Similar to `Order` here. And just have `MATCHES, MATCHES_EXACTLY, CONTAINS`.

jingtang10 (changes requested):
Thanks @epicadk for this change! Just a comment on the syntax for the search API.

engine/src/main/java/com/google/android/fhir/search/SearchDsl.kt
epicadk: Yes will do.

engine/src/androidTest/java/com/google/android/fhir/db/impl/DatabaseImplTest.kt
```diff
@@ -360,6 +362,95 @@ class DatabaseImplTest {
       .inOrder()
   }
 
+  @Test
+  fun search_HumanName() {
```
epicadk, line 366: @jingtang10 I changed the indexing pattern. This test would have failed with the previous way but now it passes.

jingtang10:
Thanks @epicadk - I would really love to focus on the contains issue and handle the change of indexing in a separate PR. Is it ok?

epicadk:
> Thanks @epicadk - I would really love to focus on the contains issue and handle the change of indexing in a separate PR. Is it ok?

Yep sorry. Reverting the change.

epicadk:
@jingtang10 done.

engine/src/main/java/com/google/android/fhir/search/params/StringSearchModifier.kt
```diff
@@ -0,0 +1,22 @@
+/*
```
jingtang10, line 1: does it make sense for this to be in `SearchDsl.kt`? `Order` is already there.

engine/src/main/java/com/google/android/fhir/search/params/StringSearchModifier.kt
```diff
@@ -0,0 +1,22 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.google.android.fhir.search.params
+
+enum class StringSearchModifier {
```
jingtang10, line 19: prefer for this to be exhaustive - add STARTS_WITH and set it as default in the DSL? Feel like that's much more explicit. so you'll have

```
STARTS_WITH,
MATCHES_EXACTLY,
CONTAINS
```

for a developer looking at these three options it's immediately clear what the behavior is. not having a default makes it a little harder to understand the behavior.

jingtang10, line 19: nit: `StringFilterModifier`?

engine/src/test/java/com/google/android/fhir/search/SearchTest.kt
```diff
@@ -46,6 +46,99 @@ class SearchTest {
...
+    assertThat(query.query)
+      .isEqualTo(
+        """
+    SELECT a.serializedResource
+    FROM ResourceEntity a
+    WHERE a.resourceType = ?
+    AND a.resourceId IN (
+    SELECT resourceId FROM StringIndexEntity
+    WHERE resourceType = ? AND index_name = ? AND index_value LIKE '%' || ? || '%' COLLATE NOCASE
+    )
```
jingtang10, line 70: does spotless give you problems here?

if not, can you make the indentation look better by adding 4 spaces so this block is in line with the `"""`.

engine/src/test/java/com/google/android/fhir/search/SearchTest.kt
```diff
@@ -46,6 +46,99 @@ class SearchTest {
...
+    assertThat(query.query)
+      .isEqualTo(
+        """
+    SELECT a.serializedResource
+    FROM ResourceEntity a
+    WHERE a.resourceType = ?
+    AND a.resourceId IN (
+    SELECT resourceId FROM StringIndexEntity
+    WHERE resourceType = ? AND index_name = ? AND index_value LIKE ? || '%' COLLATE NOCASE
+    )
```
jingtang10, line 131: same here

engine/src/androidTest/java/com/google/android/fhir/db/impl/DatabaseImplTest.kt
```diff
@@ -371,6 +372,110 @@ class DatabaseImplTest {
...
+              value = "Eve"
+              modifier = StringSearchModifier.EXACT
+            }
+          }
+          .getQuery()
+      )
+    }
+    assertThat(res).hasSize(1)
+    assertThat(res[0].id).isEqualTo("Patient/${patient1.id}")
+    assertThat(res[0].nameFirstRep.given.any { it.toString() == "Eve" }).isTrue()
```
jingtang10, line 407: this line seems to be superfulous if you have asserted that `patient1` is returned?

engine/src/androidTest/java/com/google/android/fhir/db/impl/DatabaseImplTest.kt
```diff
@@ -371,6 +372,110 @@ class DatabaseImplTest {
...
+    val patient2 =
+      Patient().apply {
+        id = "test_2"
+        addName(HumanName().addGiven("EVE"))
+      }
+    val patient3 =
+      Patient().apply {
+        id = "test_3"
+        addName(HumanName().addGiven("eve"))
+      }
```
jingtang10, line 391: i think i would prefer to break this into 2 different test cases.

```
fun search_string_exact // for matching Eve
fun search_string_exact_differentCase // for not matching EVE 
```

`EVE` and `eve` are frankly the same test case since it's about not matching the case.

engine/src/androidTest/java/com/google/android/fhir/db/impl/DatabaseImplTest.kt
```diff
@@ -371,6 +372,110 @@ class DatabaseImplTest {
...
+    val patient2 =
+      Patient().apply {
+        id = "test_2"
+        addName(HumanName().addGiven("Evelyn"))
+      }
+    val patient3 =
+      Patient().apply {
+        id = "test_3"
+        addName(HumanName().addGiven("Eve"))
+      }
```
jingtang10, line 426: hm i think you could just have one most generic case, which is that you can match a substring in a case-insensitive way.

engine/src/androidTest/java/com/google/android/fhir/db/impl/DatabaseImplTest.kt
```diff
@@ -371,6 +372,110 @@ class DatabaseImplTest {
         )
     }
 
+  @Test
+  fun search_String_Exact() {
```
jingtang10, line 376:
```suggestion
  fun search_string_exact() {
```

engine/src/androidTest/java/com/google/android/fhir/db/impl/DatabaseImplTest.kt
```diff
@@ -371,6 +372,110 @@ class DatabaseImplTest {
...
+          .getQuery()
+      )
+    }
+    assertThat(res).hasSize(1)
+    assertThat(res[0].id).isEqualTo("Patient/${patient1.id}")
+    assertThat(res[0].nameFirstRep.given.any { it.toString() == "Eve" }).isTrue()
+  }
+
+  @Test
+  fun search_String_Contains() {
```
jingtang10, line 411:
```suggestion
  fun search_string_contains() {
```

engine/src/androidTest/java/com/google/android/fhir/db/impl/DatabaseImplTest.kt
```diff
@@ -371,6 +372,110 @@ class DatabaseImplTest {
...
+    assertThat(
+        res.all { patient ->
+          patient.nameFirstRep.given.any { it.toString().toLowerCase().contains("eve") }
+        }
+      )
+      .isTrue()
+  }
+
+  @Test
+  fun search_String_Default() {
```
jingtang10, line 450:
```suggestion
  fun search_string_default() {
```

engine/src/androidTest/java/com/google/android/fhir/db/impl/DatabaseImplTest.kt
```diff
@@ -371,6 +372,110 @@ class DatabaseImplTest {
...
+    val patient2 =
+      Patient().apply {
+        id = "test_2"
+        addName(HumanName().addGiven("Evelyn"))
+      }
+    val patient3 =
+      Patient().apply {
+        id = "test_3"
+        addName(HumanName().addGiven("Severine"))
+      }
```
jingtang10, line 465: same thing here - let's split the cases.

i think you can have a matching case and a non-matching case. 2 should be enough.

jingtang10 (changes requested):
thanks!

epicadk:
@jingtang10 done

engine/src/androidTest/java/com/google/android/fhir/db/impl/DatabaseImplTest.kt
```diff
@@ -371,6 +372,141 @@ class DatabaseImplTest {
         )
     }
 
+  @Test
+  fun search_String_default() {
```
jingtang10, line 376:
```suggestion
  fun search_string_default() {
```

engine/src/androidTest/java/com/google/android/fhir/db/impl/DatabaseImplTest.kt
```diff
@@ -371,6 +372,141 @@ class DatabaseImplTest {
...
+    }
+    assertThat(res).hasSize(1)
+    assertThat(
+        res.all { patient -> patient.nameFirstRep.given.any { it.toString().startsWith("Eve") } }
+      )
+      .isTrue()
+  }
+
+  @Test
+  fun search_String_default_no_match() {
```
jingtang10, line 396:
```suggestion
  fun search_string_default_no_match() {
```

engine/src/androidTest/java/com/google/android/fhir/db/impl/DatabaseImplTest.kt
```diff
@@ -371,6 +372,141 @@ class DatabaseImplTest {
...
+      database.insert(patient)
+      database.search<Patient>(
+        Search(ResourceType.Patient).apply { filter(Patient.GIVEN) { value = "eve" } }.getQuery()
+      )
+    }
+    assertThat(res).hasSize(0)
+  }
+
+  @Test
+  fun search_String_exact() {
```
jingtang10, line 412:
```suggestion
  fun search_string_exact() {
```

engine/src/androidTest/java/com/google/android/fhir/db/impl/DatabaseImplTest.kt
```diff
@@ -371,6 +372,141 @@ class DatabaseImplTest {
...
+            }
+          }
+          .getQuery()
+      )
+    }
+    assertThat(res).hasSize(1)
+    assertThat(res[0].nameFirstRep.given.any { it.toString() == "Eve" }).isTrue()
+  }
+  @Test
+  fun search_String_exact_no_match() {
```
jingtang10, line 435:
```suggestion
  fun search_string_exact_no_match() {
```

engine/src/androidTest/java/com/google/android/fhir/db/impl/DatabaseImplTest.kt
```diff
@@ -371,6 +372,141 @@ class DatabaseImplTest {
...
+            }
+          }
+          .getQuery()
+      )
+    }
+    assertThat(res).hasSize(0)
+  }
+
+  @Test
+  fun search_String_contains() {
```
jingtang10, line 458:
```suggestion
  fun search_string_contains() {
```

engine/src/androidTest/java/com/google/android/fhir/db/impl/DatabaseImplTest.kt
```diff
@@ -371,6 +372,141 @@ class DatabaseImplTest {
...
+    assertThat(
+        res.all { result ->
+          result.nameFirstRep.given.any { it.toString().toLowerCase().contains("eve") }
+        }
+      )
+      .isTrue()
+  }
+
+  @Test
+  fun search_String_contains_no_match() {
```
jingtang10, line 488:
```suggestion
  fun search_string_contains_no_match() {
```

engine/src/androidTest/java/com/google/android/fhir/db/impl/DatabaseImplTest.kt
```diff
@@ -371,6 +372,141 @@ class DatabaseImplTest {
...
+        id = "test_1"
+        addName(HumanName().addGiven("Evelyn"))
+      }
+    val res = runBlocking {
+      database.insert(patient)
+      database.search<Patient>(
+        Search(ResourceType.Patient).apply { filter(Patient.GIVEN) { value = "eve" } }.getQuery()
+      )
+    }
+    assertThat(res).hasSize(1)
```
jingtang10, line 388: nit: add black line before assertions - and same for other test cases.

engine/src/androidTest/java/com/google/android/fhir/db/impl/DatabaseImplTest.kt
```diff
@@ -371,6 +372,141 @@ class DatabaseImplTest {
...
+    assertThat(
+        res.all { patient -> patient.nameFirstRep.given.any { it.toString().startsWith("Eve") } }
+      )
+      .isTrue()
```
jingtang10, line 392: kinda feel like you just need to assert that the right patient (with id `test_1`) is returned. no need to actually assert the name indeed starts with `Eve`. actually the name might start with `EVE` or `eve`. so this assertion doesn't need to be here.

engine/src/androidTest/java/com/google/android/fhir/db/impl/DatabaseImplTest.kt
jingtang10: i think just checking the patient id is right is enough?

engine/src/androidTest/java/com/google/android/fhir/db/impl/DatabaseImplTest.kt
```diff
@@ -371,6 +372,141 @@ class DatabaseImplTest {
...
+          .getQuery()
+      )
+    }
+    assertThat(res).hasSize(1)
+    assertThat(
+        res.all { result ->
+          result.nameFirstRep.given.any { it.toString().toLowerCase().contains("eve") }
+        }
+      )
+      .isTrue()
```
jingtang10, line 484: same here

jingtang10 (approved):
Great! Thanks so much for this. Please feel free to merge after addressing the final few (nit) comments.

jingtang10:
Thanks again @epicadk for this PR and merging!

PR #436, opened 2021-04-10
Feat: Number Search

### LINKED ISSUE #435: Support Search for Number Type

Currently we do not support search for Number types

Support Search for Number types

NA

**Additional context**

https://www.hl7.org/fhir/search.html#number

Yes

epicadk:

Fixes #435 

**Description**

Add support for Number search.

**Alternative(s) considered**

Followed a similar approach to strings

**Type**
Choose one: Feature

**Checklist**


epicadk:
What should we set as the threshold for the Approximate search prefix?

epicadk:
Will add tests to show that the SQL query works

epicadk:
Will finish this by today.

epicadk:
Pr ready for review

epicadk:
I'll make changes to the tests so that they are similar to the string search ones.

jingtang10:
@kevinmost can you please leave a review because this is related to your work on indexing numbers. Thanks!

engine/src/main/java/com/google/android/fhir/search/MoreSearch.kt
```diff
@@ -109,6 +113,81 @@ fun StringFilter.query(type: ResourceType): SearchQuery {
   )
 }
 
+fun NumberFilter.query(type: ResourceType): SearchQuery {
+  val precision =
+    when {
+      value!!.scale() >= 0 ->
```
kevinmost, line 119: There are a lot of `value!!` throughout this method. Can we just null-assert once, at the top of the file, and capture it in a local val?

```kt
val value = value!!
val precision = 
  when {
    value.scale() >= 0 -> 
  // and so on
```

engine/src/main/java/com/google/android/fhir/search/MoreSearch.kt
```diff
@@ -109,6 +113,81 @@ fun StringFilter.query(type: ResourceType): SearchQuery {
...
+            type.name,
+            parameter.paramName,
+            (value!! - precision).toDouble(),
+            (value!! + precision).toDouble()
+          )
+        )
+      // Ends_Before and Starts_After are not used with integer values.
+      ParamPrefixEnum.ENDS_BEFORE ->
+        if (value!!.scale() <= 0)
+          throw IllegalArgumentException("Prefix not allowed for Integer type")
```
kevinmost, line 159: You can use `require` here. I would also put the prefix into the error message.

```
require(value.scale() > 0) { "Prefix $prefix not allowed for Integer type" }
```

engine/src/main/java/com/google/android/fhir/search/MoreSearch.kt
```diff
@@ -109,6 +113,81 @@ fun StringFilter.query(type: ResourceType): SearchQuery {
...
+        else "<"
+      ParamPrefixEnum.STARTS_AFTER ->
+        if (value!!.scale() <= 0)
+          throw IllegalArgumentException("Prefix not allowed for Integer type")
```
kevinmost, line 163: Ditto

kevinmost (approved):

kevinmost:
@jingtang10 LGTM, not sure if you wanted to approve before this lands.

epicadk:
Ran the integration tests they pass locally, can't figure out why they fail in the CI

kevinmost:
I think something may be wrong with CI (again). I'm seeing CI fail on one of my PRs even though I don't think it should.

epicadk:
I checked the test results in the artifacts and the test in the reference app seems to be failing.

engine/src/main/java/com/google/android/fhir/search/MoreSearch.kt
```diff
@@ -109,6 +113,81 @@ fun StringFilter.query(type: ResourceType): SearchQuery {
   )
 }
 
+fun NumberFilter.query(type: ResourceType): SearchQuery {
```
jingtang10, line 115: similar to the comment i left in the date time PR. i find the structure a bit difficult to read because of the early returns in some of the branches in `when`. can you create a utility function to make the logic a bit more readable (and unify the code paths). thanks!

jingtang10 (changes requested):

engine/src/main/java/com/google/android/fhir/search/MoreSearch.kt
```diff
@@ -149,3 +172,37 @@ val Order?.sqlString: String
       Order.DESCENDING -> "DESC"
       null -> ""
     }
+
+fun getConditionParamPair(
```
jingtang10, line 176: private

engine/src/main/java/com/google/android/fhir/search/MoreSearch.kt
```diff
@@ -113,6 +116,26 @@ fun StringFilter.query(type: ResourceType): SearchQuery {
...
 }
 
+fun NumberFilter.query(type: ResourceType): SearchQuery {
+  val value = value!!
+  val precision =
+    when {
+      value.scale() >= 0 ->
+        BigDecimal(1).divide(BigDecimal(10).pow(value.scale())).divide(BigDecimal(2))
+      else -> BigDecimal(5)
+    }
```
jingtang10, line 126: I feel this is a little too lenient. 0.49 would be within the margin to be equal to 0.5 (scale would be 1, 10^(-1) = 0.1, so the precision is 0.05). I feel like we need to increase the scale by at least one or two decimal points. So that maybe 0.499 can be deemed equivalent to 0.5, but not 0.49.

Also, can you create a utility function with a little bit of explanation in the documentation (with examples) to explain what's happening here?

Thanks!

engine/src/main/java/com/google/android/fhir/search/MoreSearch.kt
```diff
@@ -149,3 +172,37 @@ val Order?.sqlString: String
...
+fun getConditionParamPair(
+  prefix: ParamPrefixEnum?,
+  value: BigDecimal,
+  precision: BigDecimal
+): Pair<String, List<Double>> {
+  return when (prefix) {
+    ParamPrefixEnum.EQUAL, null ->
+      "index_value >= ? AND index_value < ?" to
+        listOf((value - precision).toDouble(), (value + precision).toDouble())
+    ParamPrefixEnum.GREATERTHAN -> "index_value > ?" to listOf(value.toDouble())
```
jingtang10, line 185:
```suggestion
    ParamPrefixEnum.GREATERTHAN -> "index_value > ?" to listOf((value - precision).toDouble())
```

same should be applied to everything below

jingtang10 (changes requested):

engine/src/main/java/com/google/android/fhir/search/MoreSearch.kt
epicadk: according to the [standard](https://www.hl7.org/fhir/search.html#number)

> When a comparison prefix in the set lgt, lt, ge, le, sa & eb is provided, the implicit precision of the number is ignored, and they are treated as if they have arbitrarily high precision

engine/src/main/java/com/google/android/fhir/search/MoreSearch.kt
epicadk: Actually 0.5 would simply anything in the range of [0.45,0.55). I think this mainly has to do with [Significant figures](https://en.wikipedia.org/wiki/Significant_figures#Rounding_to_significant_figures)

engine/src/main/java/com/google/android/fhir/search/MoreSearch.kt
jingtang10: oh i didn't realise this is described in the standard. can you maybe add a comment in the code to link to https://www.hl7.org/fhir/search.html#number?

engine/src/main/java/com/google/android/fhir/search/MoreSearch.kt
jingtang10: oh i see this is mandated in the standard as well https://www.hl7.org/fhir/search.html#number.

engine/src/main/java/com/google/android/fhir/search/MoreSearch.kt
```diff
@@ -149,3 +172,37 @@ val Order?.sqlString: String
...
+    ParamPrefixEnum.GREATERTHAN_OR_EQUALS -> "index_value >= ?" to listOf(value.toDouble())
+    ParamPrefixEnum.LESSTHAN -> "index_value < ?" to listOf(value.toDouble())
+    ParamPrefixEnum.LESSTHAN_OR_EQUALS -> "index_value <= ?" to listOf(value.toDouble())
+    ParamPrefixEnum.NOT_EQUAL ->
+      "index_value < ? OR index_value >= ?" to
+        listOf((value - precision).toDouble(), (value + precision).toDouble())
+    // Ends_Before and Starts_After are not used with integer values.
```
jingtang10, line 192: where in the standard says this?

jingtang10 (changes requested):
thanks very much! a few more comments.

engine/src/main/java/com/google/android/fhir/search/MoreSearch.kt
epicadk: [This section](https://www.hl7.org/fhir/search.html#prefix) the small paragraph immediately after the table

engine/src/main/java/com/google/android/fhir/search/MoreSearch.kt
```diff
@@ -113,6 +116,25 @@ fun StringFilter.query(type: ResourceType): SearchQuery {
...
 
+/**
+ * Extension function that returns a SearchQuery based on the value and prefix of the NumberFilter
+ */
+fun NumberFilter.query(type: ResourceType): SearchQuery {
+  val value = value!!
+
+  val precision = value.getRange()
+
+  val conditionParamPair = getConditionParamPair(prefix, value, precision)
```
jingtang10, line 127:
```suggestion
  val conditionAndParams = getConditionParamPair(prefix, value!!)
```

and call `value.getRange()` inside `getConditionParamPair`

engine/src/main/java/com/google/android/fhir/search/MoreSearch.kt
```diff
@@ -149,3 +171,60 @@ val Order?.sqlString: String
       Order.DESCENDING -> "DESC"
       null -> ""
     }
+
+/**
+ * returns the condition and list of params required in NumberFilter.query see
+ * https://www.hl7.org/fhir/search.html#number
```
jingtang10, line 177:
```suggestion
 * Returns the condition and list of params required in NumberFilter.query see
 * https://www.hl7.org/fhir/search.html#number.
```

engine/src/test/java/com/google/android/fhir/search/SearchTest.kt
```diff
@@ -633,4 +635,342 @@ class SearchTest {
...
+        listOf(
+          ResourceType.RiskAssessment.name,
+          ResourceType.RiskAssessment.name,
+          RiskAssessment.PROBABILITY.paramName,
+          BigDecimal("100.00").toDouble()
+        )
+      )
+  }
+
+  @Test(expected = java.lang.IllegalArgumentException::class)
```
jingtang10, line 848: use this: https://github.com/google/android-fhir/search?q=assertthrows

engine/src/test/java/com/google/android/fhir/search/SearchTest.kt
```diff
@@ -633,4 +635,342 @@ class SearchTest {
...
```
jingtang10, line 895: use this: https://github.com/google/android-fhir/search?q=assertthrows

engine/src/main/java/com/google/android/fhir/search/MoreSearch.kt
```diff
@@ -149,3 +171,60 @@ val Order?.sqlString: String
...
+    // Ends_Before and Starts_After are not used with integer values.
+    ParamPrefixEnum.ENDS_BEFORE -> {
+      require(value.scale() > 0) { "Prefix $prefix not allowed for Integer type" }
+      "index_value < ?" to listOf(value.toDouble())
+    }
+    ParamPrefixEnum.STARTS_AFTER -> {
+      require(value.scale() > 0) { "Prefix $prefix not allowed for Integer type" }
+      "index_value > ?" to listOf(value.toDouble())
+    }
+    // Approximate to a 10% range
```
jingtang10, line 204: add a link to the standard where this is explained.

jingtang10 (changes requested):
Apologies really should have sent these in the previous rounds... but I think this might be the last review :)

engine/src/main/java/com/google/android/fhir/search/MoreSearch.kt
```diff
@@ -149,3 +171,60 @@ val Order?.sqlString: String
...
+ * Extension function that returns the range in which the value should lie for it to be considered a
+ * match (@see NumberFilter.query) . The value is directly related to the scale of the BigDecimal.
+ *
+ * For example A search for 100.00 (has a scale of 2) would match any value in [99.995, 100.005) and
+ * the function returns 0.005
+ *
+ * For Big integers which have a negative scale the function returns 5 For example A search for 1e3
+ * would match any value in [995, 1005) and the function returns 5
+ *
+ * Note 100 is considered to have 2 significant digits
```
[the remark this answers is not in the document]
epicadk: Sorry matches upto 2. Yep, I'll remove it. I mainly included it to highlight the difference between 1e2 and 100

engine/src/main/java/com/google/android/fhir/search/MoreSearch.kt
```diff
@@ -149,3 +171,60 @@ val Order?.sqlString: String
...
+ */
+private fun BigDecimal.getRange(): BigDecimal {
+  return when {
+    scale() >= 0 -> BigDecimal(0.5).divide(BigDecimal(10).pow(scale()))
+    else -> BigDecimal(5)
```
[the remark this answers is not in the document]
epicadk: Sure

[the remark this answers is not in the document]
epicadk: 100 has a scale of 0 but 1e2 will have a scale of -2

Tarun-Bhardwaj:
@epicadk , it seems there are some changes requested on this PR. Is there an ETA by when you think this can be closed?

epicadk:
> @epicadk , it seems there are some changes requested on this PR. Is there an ETA by when you think this can be closed?

Yup working on it. Will be done by the end of the day.

engine/src/test/java/com/google/android/fhir/search/SearchTest.kt
```diff
@@ -633,4 +636,361 @@ class SearchTest {
         )
       )
   }
+
+  @Test
+  fun search_number_equals() {
+    /* x contains pairs of values and their corresponding range (see BigDecimal.getRange() in
+    MoreSearch.KT) */
+    for (x in
```
jingtang10, line 644: actually slightly prefer to split this into 3 different test cases.

i think `1e2` and `1e3` can be merged into one test case.

engine/src/main/java/com/google/android/fhir/search/MoreSearch.kt
```diff
@@ -149,3 +168,62 @@ val Order?.sqlString: String
...
+    ParamPrefixEnum.LESSTHAN -> "index_value < ?" to listOf(value.toDouble())
+    ParamPrefixEnum.LESSTHAN_OR_EQUALS -> "index_value <= ?" to listOf(value.toDouble())
+    ParamPrefixEnum.NOT_EQUAL -> {
+      val precision = value.getRange()
+      "index_value < ? OR index_value >= ?" to
+        listOf((value - precision).toDouble(), (value + precision).toDouble())
+    }
+    // Ends_Before and Starts_After are not used with integer values.
+    ParamPrefixEnum.ENDS_BEFORE -> {
+      require(value.scale() > 0) { "Prefix $prefix not allowed for Integer type" }
```
jingtang10, line 197: maybe move this to the front of the function

`require(value.scale() <= 0 || (prefix != STARTS_AFTER && prefix != ENDS_BEFORE))`

jingtang10 (approved):
thanks @epicadk! great work!

👍

engine/src/test/java/com/google/android/fhir/search/SearchTest.kt
jingtang10: or use proper parameterized tests? may be an overkill though.

epicadk: I removed the 1e2 test case for now, maybe we can discuss parametrized tests in a different issue?

engine/src/main/java/com/google/android/fhir/search/MoreSearch.kt
epicadk: I think the condition would remain value.scale() > 0 right?

jingtang10: oh yeah that's right

engine/src/test/java/com/google/android/fhir/search/SearchTest.kt
jingtang10: i mean you could even just have 3 different test cases so test failures are more explicit.

eg. `search_number_equals_int`, `search_number_equals_decimal` etc

epicadk:
Keeping only the relevant commit messages in the commit body. I hope that's okay?

epicadk:
@Tarun-Bhardwaj done with this, can focus on adding Date Search now

epicadk:
> I checked the test results in the artifacts and the test in the reference app seems to be failing. Temporarily deleted the test to check if that is indeed the issue. cc @jingtang10

This is still failing removing it for now, we can add it back later. (It's just an example anyways.)

PR #457, opened 2021-04-28
chore : Add relevant files to `paths-ignore`

epicadk:

trivial

**Description**

Skip the workflow if changes are made only to .md files ( like `readme.md`) 

**Alternative(s) considered**
 
We could add the `license.txt` file as well but spotless check might fail if changes are made to the file so I've decided to include it. 

**Type**
Choose one: (Other)

**Checklist**

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -20,7 +20,7 @@ on:
   push:
     branches: [ master ]
     paths-ignore: 
-      - '*.md'
+      - '**.md'
     tags:
       run_ci*
   pull_request:
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -19,10 +19,14 @@ name: CI
 on:
   push:
     branches: [ master ]
+    paths-ignore: 
+      - '*.md'
     tags:
       run_ci*
   pull_request:
     branches: [ master ]
+    paths-ignore: 
+      - '*.md'
 
 # A workflow run is made up of one or more jobs that can run sequentially or in parallel
 jobs:
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -26,7 +26,7 @@ on:
   pull_request:
     branches: [ master ]
     paths-ignore: 
-      - '*.md'
+      - '**.md'
 
 # A workflow run is made up of one or more jobs that can run sequentially or in parallel
 jobs:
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -19,10 +19,14 @@ name: CI
 on:
   push:
     branches: [ master ]
+    paths-ignore: 
+      - '*.md'
```
jingtang10, line 23:
```suggestion
      - '**.md'
```

.github/workflows/ci.yml
```diff
@@ -19,10 +19,14 @@ name: CI
...
     tags:
       run_ci*
   pull_request:
     branches: [ master ]
+    paths-ignore: 
+      - '*.md'
```
jingtang10, line 29:
```suggestion
      - '**.md'
```

jingtang10 (changes requested):
Thanks for this! Great idea using `paths-ignore`.

Please do modify the paths though to use `**` so that it covers subdirectories.

epicadk:
> Thanks for this! Great idea using `paths-ignore`.
> 
> Please do modify the paths though to use `**` so that it covers subdirectories.

Yup sorry I wasn't aware of the difference between `*` and `**`. I've made the changes.

epicadk:
I also added `CODEOWNERS` and `LICENSE` because they wouldn't really affect the outcome of the CI

jingtang10:
Great! Thanks for this!

jingtang10 (approved):

jingtang10:
This PR is causing some issues. If a PR is raised with changes only to the ignored files, no CI workflow will be triggered. But because we have a branch rule for `master` to ensure only PRs with passing tests can be merged, the PR cannot actually be merged.

I'm not sure this PR is still the right thing to do given this. Either we revert this PR, or if we can find a way for GitHub to not enforce the testing rule for PRs with only documentation change.

jingtang10:
@deepankarb had this problem for #720

epicadk:
I had a look at some of the actions repos and even they have removed the path filter due to a similar issue. (lastest commit on https://github.com/actions/upload-artifact). So I think removing the filter is the way to go.

The logic behind this change was to avoid running the tests because the code wouldn't have changed. So if we do find a way to implement this properly we could revisit this later.

deepankarb:
Thanks @epicadk @jingtang10 The worst case scenario is when there are lots of small changes in documentation in multiple PRs (single PR updation will cancel any subsequent runs AFAIK so not an issue). Such a scenario is IMO rare. I think it's better to remove the `paths-ignore`.

PR #482, opened 2021-05-18
Rework sync API to be separate from data access API

jingtang10:

Fixes #[issue number]

**Description**

**Alternative(s) considered**

**Type**

**Checklist**


engine/src/main/java/com/google/android/fhir/FhirEngine.kt
```diff
@@ -55,20 +55,13 @@ interface FhirEngine {
...
-
-  /** Attempts to upload locally created and modified resources. */
-  suspend fun syncUpload(): Result
+  suspend fun <R : Resource> search(search: Search): List<R>
 
-  suspend fun periodicSync(): Result
+  suspend fun syncDownload(download: suspend (SyncDownloadContext) -> List<Resource>)
 
-  fun updatePeriodicSyncConfiguration(syncConfig: PeriodicSyncConfiguration)
+  suspend fun syncUpload(upload: (suspend (List<SquashedLocalChange>) -> List<LocalChangeToken>))
```
deepankarb, line 62: I think we're leaking DB abstraction here. We'll have to think of a way to do decouple keeping the abstractions.

build.gradle.kts
```diff
@@ -1,5 +1,6 @@
 // Top-level build file where you can add configuration options common to all sub-projects/modules.
 buildscript {
+  val kotlin_version by extra("1.4.31")
```
yigit, line 3: for future reference, gradle 7.0 adds a much nicer way to define dependencies:

https://docs.gradle.org/7.0/release-notes.html#dependency-management-improvements

engine/src/main/java/com/google/android/fhir/db/impl/dao/SyncedResourceDao.kt
```diff
@@ -28,6 +29,11 @@ interface SyncedResourceDao {
 
   @Insert(onConflict = OnConflictStrategy.REPLACE) suspend fun insert(entity: SyncedResourceEntity)
 
+  @Transaction
+  suspend fun insertAll(resources: List<SyncedResourceEntity>) {
+    resources.forEach { resource -> insert(resource) }
+  }
```
yigit, line 35:
```suggestion
  @Insert(onConflict = OnConflictStrategy.REPLACE)
  suspend fun insertAll(resources: List<SyncedResourceEntity>)
```

you don't need to do it manually, room can receive a list to insert.

engine/src/main/java/com/google/android/fhir/impl/FhirEngineImpl.kt
```diff
@@ -69,53 +56,29 @@ constructor(
...
+      download(
+        object : SyncDownloadContext {
+          override suspend fun getLatestTimestamptFor(type: ResourceType) =
+            database.lastUpdate(type)
+        }
+      )
+
+    val timeStamps =
+      stuff.groupBy { it.resourceType }.entries.map {
+        SyncedResourceEntity(it.key, it.value.last().meta.lastUpdated.toTimeZoneString())
```
yigit, line 74: are these guaranteed to be ordered?

engine/src/main/java/com/google/android/fhir/impl/FhirEngineImpl.kt
```diff
@@ -69,53 +56,29 @@ constructor(
...
       }
+    database.insertSyncedResources(timeStamps, stuff)
+  }
 
-    WorkManager.getInstance(context)
-      .enqueueUniqueWork(SyncWorkType.DOWNLOAD.workerName, ExistingWorkPolicy.KEEP, downloadRequest)
+  override suspend fun syncUpload(
+    upload: (suspend (List<SquashedLocalChange>) -> List<LocalChangeToken>)
+  ) {
+    upload(database.getAllLocalChanges()).forEach { database.deleteUpdates(it) }
```
yigit, line 82: you probably want this in a transaction ?

yigit: also, you may want to check again for local changes and call it back again if there are new local changes after the sync call? (e.g. while(localHcanges.isNotEmpty() { .. }

engine/src/main/java/com/google/android/fhir/sync/FhirSynchronizer.kt
```diff
@@ -33,70 +36,106 @@ data class ResourceSyncException(val resourceType: ResourceType, val exception:
...
-              FHIRException(
-                "Could not infer response \"${response.resourceType}/${response.logicalId}\" as success."
+        while (nextUrl != null) {
+          val bundle = dataSource.loadData(nextUrl)
+          nextUrl = bundle.link.firstOrNull { component -> component.relation == "next" }?.url
+          if (bundle.type == Bundle.BundleType.SEARCHSET) {
+            result.addAll(bundle.entry.map { it.resource })
+          }
+        }
+      } catch (e: IOException) {}
```
yigit, line 74: we need to figure and and document what the expectation is here.
if they need to catch exceptions, lets document it in syncDownload

reference/src/main/java/com/google/android/fhir/reference/PatientListActivity.kt
```diff
@@ -40,12 +52,28 @@ class PatientListActivity() : AppCompatActivity() {
...
+      )
+    )
+
     val toolbar = findViewById<Toolbar>(R.id.toolbar)
     setSupportActionBar(toolbar)
     toolbar.title = title
 
     fhirEngine = fhirEngine(this)
 
+    lifecycleScope.launch {
```
yigit, line 69: you don't want lifecycle scope here (as it will be cancelled if activity disappears).

I know this is just tesst code but you usually want an application level scope here to run the work and observe the reesult. if UI disappears, sync still happens just no notification in the UI.

yigit: keeep in mind, if you change this to be an application level scope, then the code shouldn't be here as it would leak the activity.

yigit (changes requested):

build.gradle.kts
jingtang10: reverting this change. added a new issue #492

engine/src/main/java/com/google/android/fhir/db/impl/dao/SyncedResourceDao.kt
jingtang10: done

engine/src/main/java/com/google/android/fhir/impl/FhirEngineImpl.kt
jingtang10: fixed.

engine/src/main/java/com/google/android/fhir/impl/FhirEngineImpl.kt
jingtang10: pushed this into the db class for transaction.

engine/src/main/java/com/google/android/fhir/sync/FhirSynchronizer.kt
jingtang10: ack. i made note - @deepankarb is working on this.

engine/src/main/java/com/google/android/fhir/db/impl/DatabaseImpl.kt
```diff
@@ -128,6 +130,34 @@ internal class DatabaseImpl(context: Context, private val iParser: IParser, data
...
   }
 
+  override suspend fun syncUpload(
+    upload: suspend (List<SquashedLocalChange>) -> List<LocalChangeToken>
+  ) {
+    db.withTransaction {
+      var localChanges = getAllLocalChanges()
+      while (localChanges.isNotEmpty()) {
+        upload(localChanges).forEach { deleteUpdates(it) }
+      }
```
yigit, line 140: Lets not call user code in a transaction, if they try to access back into the database, bad things might happen.

```suggestion
    var localChanges = getAllLocalChanges()
      while (localChanges.isNotEmpty()) {
         val uploaded = upload(localChanges)
         deleteUpdates(uploaded)
         localChanges = getAllLocalChanges()
      }
```

now that suggestion hits into a race condition hence we need to have a mutex around this to avoid concurrent access.

engine/src/main/java/com/google/android/fhir/db/impl/DatabaseImpl.kt
```diff
@@ -128,6 +130,34 @@ internal class DatabaseImpl(context: Context, private val iParser: IParser, data
...
+    }
+  }
+
+  override suspend fun syncDownload(download: suspend (SyncDownloadContext) -> List<Resource>) {
+    db.withTransaction {
```
yigit, line 145: same here. basically, during this method we are holding a database lock and because they are doing network, it might lock for a very long time.

engine/src/main/java/com/google/android/fhir/sync/Config.kt
```diff
@@ -17,28 +17,58 @@
...
+
+/** Configuration for synchronization. */
+data class SyncConfiguration(
+  /** Data that needs to be synchronised */
+  val resourceSyncParams: List<ResourceSyncParams> = emptyList(),
+  /**
+   * true if the SDK needs to retry a failed sync attempt, false otherwise If this is set to true,
+   * then the result of the sync will be reported after the retry.
+   */
+  val retry: Boolean = false
```
yigit, line 48: where is this used?

engine/src/main/java/com/google/android/fhir/sync/FhirSynchronizer.kt
```diff
@@ -33,70 +35,103 @@ data class ResourceSyncException(val resourceType: ResourceType, val exception:
...
-        ResourceSynchronizer(syncData, dataSource, database, syncConfiguration.retry)
+
+    upload()
+
+    resourceSyncParams.forEach {
       try {
-        resourceSynchroniser.sync()
+        downloadResourceType(it.key, it.value)
       } catch (exception: Exception) {
-        exceptions.add(ResourceSyncException(syncData.resourceType, exception))
```
yigit, line 48: you should avoid catching CancelationException here. it would be also good to have a test that we handle cancelation properly (coroutine cancelation happens via CancelationExceptions)

yigit (changes requested):
sorry i miseed it in the first CL but can you add more tests here. this part is fairly tricky so would be good to test well.

engine/src/main/java/com/google/android/fhir/db/impl/DatabaseImpl.kt
jingtang10: done

engine/src/main/java/com/google/android/fhir/db/impl/DatabaseImpl.kt
jingtang10: done

engine/src/main/java/com/google/android/fhir/sync/Config.kt
jingtang10: https://github.com/google/android-fhir/issues/506

engine/src/main/java/com/google/android/fhir/sync/FhirSynchronizer.kt
jingtang10: https://github.com/google/android-fhir/issues/507

yigit (approved):

PR #511, opened 2021-05-28
Added onFocusChangeListner and RequiredConstraintValidator.kt

### LINKED ISSUE #470: Handle required fields in questionnaires
Handle the required field in the questionnaires: https://www.hl7.org/fhir/questionnaire-definitions.html#Questionnaire.item.required

Add a validator so that if a required field is not filled in display some kind of warning.

NA

**Additional context**
NA

NA

kunjan8794:
Fixes #470 

**Description**
Handle the required field in the questionnaires - add validator

**Type**
Feature

**Checklist**
- [X] I have read and acknowledged the [Code of conduct](https://github.com/google/android-fhir/blob/master/docs/CODE_OF_CONDUCT.md)
- [X] I have read [How to Contribute](https://github.com/google/android-fhir/blob/master/docs/contributing.md)
- [X] I have read the [Developer's guide](https://github.com/google/android-fhir/wiki/Developer's-Guide)
- [X] I have signed the Google [Individual CLA](https://cla.developers.google.com/about/google-individual), or I am covered by my company's [Corporate CLA](https://cla.developers.google.com/about/google-corporate )
- [X] I have discussed my proposed solution with code owners in the linked issue(s) and we have agreed upon the general approach
- [X] I have run `./gradlew spotlessApply` and `./gradlew spotlessCheck` to check my code follows the style guide of this project
- [X] I have run `./gradlew check` and `./gradlew connectedCheck` to test my changes locally
- [X] I have built and run the reference app(s) to verify my change fixes the issue and/or does not break the reference app(s)


datacapture/src/main/java/com/google/android/fhir/datacapture/views/QuestionnaireItemEditTextViewHolderFactory.kt
```diff
@@ -59,6 +60,19 @@ internal abstract class QuestionnaireItemEditTextViewHolderDelegate(
         )
       )
     }
+    textInputEditText.setOnFocusChangeListener { _, hasFocus ->
+      if (!hasFocus) {
+        val validate =
+          RequiredConstraintValidator.validate(
```
jingtang10, line 66: i don't think it's right to call the RequiredConstraintValidator individually. it should be added to the list of validators used to validate any question item and answer.

datacapture/src/test/java/com/google/android/fhir/datacapture/validation/RequiredConstraintValidatorTest.kt
```diff
@@ -0,0 +1,60 @@
...
+        addAnswer(
+          QuestionnaireResponse.QuestionnaireResponseItemAnswerComponent().apply {
+            value = IntegerType(9)
+          }
+        )
+      }
+
+    val validationResult =
+      RequiredConstraintValidator.validate(questionnaireItem, questionnaireResponseItem)
+    Truth.assertThat(validationResult.isValid).isTrue()
```
jingtang10, line 47: import Truth

jingtang10 (changes requested):

datacapture/src/main/java/com/google/android/fhir/datacapture/views/QuestionnaireItemEditTextViewHolderFactory.kt
```diff
@@ -52,12 +52,16 @@ internal abstract class QuestionnaireItemEditTextViewHolderDelegate(
     textInputEditText.doAfterTextChanged { editable: Editable? ->
       questionnaireItemViewItem.singleAnswerOrNull = getValue(editable.toString())
       questionnaireItemViewItem.questionnaireResponseItemChangedCallback()
```
jingtang10, line 54: do you think we can move line 53-54 to the on focus change listener too?

jingtang10 (approved):
Thanks! Just a comment on enablement evaluation as well.

datacapture/src/main/java/com/google/android/fhir/datacapture/views/QuestionnaireItemEditTextViewHolderFactory.kt
kunjan8794: @jingtang10 
We could do that, but possibly, can there be a scenario where a user is filling up the form, and then doesn't really lose focus on the last element and then clicks on the submit button?

jingtang10: yes and the same argument can be made for the required field validator. should we move the required field validation back as well?

i think we were leaning towards only validate when the user is moving away from the widget. but having thought about this a little more, on mobile we don't really have the signal for moving away from a widget... the user could scroll down but the 'focus' is not lost on the widget. so maybe we should just keep everything in the change listener?

kunjan8794: So again in that case, if we think about the RequiredConstraintValidator, if someone does not ever touch a required component, then the text change listener might never get called and the warning might never be displayed.

jingtang10: discussed this with @fredhersch.. he raised the point that if the user starts to type we don't want to show an error message before they finish typing (it would seem quite annoying).

i'm happy for us to leave it as is in this PR and maybe come back to refine where exactly the validation is called (and potentially adding an option for the developer to specify a mode)

datacapture/src/main/java/com/google/android/fhir/datacapture/views/QuestionnaireItemDateTimePickerViewHolderFactory.kt
```diff
@@ -157,12 +165,6 @@ internal object QuestionnaireItemDateTimePickerViewHolderFactory :
               )
             )
         questionnaireItemViewItem.questionnaireResponseItemChangedCallback()
-        applyValidationResult(
-          QuestionnaireResponseItemValidator.validate(
-            questionnaireItemViewItem.questionnaireItem,
-            questionnaireItemViewItem.questionnaireResponseItem
-          )
-        )
```
jingtang10, line 165: slightly odd this is removed by the same isn't removed in the date picker file?

jingtang10 (approved):

Tarun-Bhardwaj:
Hi @kunjan8794 , did you get a chance to resolve this issue?

datacapture/src/main/res/values/strings.xml
```diff
@@ -0,0 +1,6 @@
+<resources>
+    <string name="required_constraint_validation_message">Missing answer for required field.</string>
+    <string name="min_value_validation_prefix">Minimum value allowed is:</string>
+    <string name="max_value_validation_prefix">Maximum value allowed is:</string>
```
jingtang10, line 4: use this: https://developer.android.com/guide/topics/resources/localization#mark-message-parts

```suggestion
    <string name="min_value_validation_error_msg">Minimum value allowed is: <xliff:g id="min_value">%1$f</xliff></string>
    <string name="max_value_validation_error_msg">Maximum value allowed is: <xliff:g id="max_value">%1$f</xliff></string>
```

datacapture/src/main/java/com/google/android/fhir/datacapture/validation/MinValueConstraintValidator.kt
```diff
@@ -27,7 +29,9 @@ internal object MinValueConstraintValidator :
       answer: QuestionnaireResponse.QuestionnaireResponseItemAnswerComponent ->
       answer.value < extension.value
     },
-    { extension: Extension -> "Minimum value allowed is:" + extension.value.primitiveValue() }
+    { extension: Extension, context: Context ->
+      context.getString(R.string.min_value_validation_prefix) + extension.value.primitiveValue()
```
jingtang10, line 33:
```suggestion
      context.getString(R.string.min_value_validation_error_msg, extension.value.primitiveValue())
```

jingtang10 (approved):

jingtang10 (approved):

jingtang10 (approved):

PR #526, opened 2021-06-01
Add max/min length validators for questionnaire

### LINKED ISSUE #185: Support max length and min length
In our views for different question types, we need to validate the lengths of the answers by respecting the min and max length constraints.

The max length constraint is part of the `Questionnaire` resource in FHIR: https://www.hl7.org/fhir/questionnaire-definitions.html#Questionnaire.item.maxLength
The min length constarint is part of the structured data capture implementation guide (SDC IG): http://hl7.org/fhir/uv/sdc/2019May/behavior.html#value-constraints Notice the difference: it's an extension, not part of the core resource.

@kunjan8794 already created a validation framework: https://github.com/google/android-fhir/tree/master/datacapture/src/main/java/com/google/android/fhir/datacapture/validation

We will need to create possibly 2 more validators, which will be triggered if the max length field is set or the min length extension exists.

The result of the validation will then be displayed in the UI in the appropriate places in different types of input.

To begin with, we can probably focus on EditText types of controls.

stevenckngaa:

Fixes #185 

**Description**
The max/min length validators are only enforced on answers that
are in the the following FHIR primitive types:
1. BooleanType
2. DecimalType
3. IntegerType
4. DateType
5. TimeType
6. StringType
7. UriType

This list is an intersection of the FHIR primitive types & the FHIR
questionnaire response supported types.

**Alternative(s) considered**
No, I haven't.

**Type**
Choose one: Feature

N/A

**Checklist**
- [x]  I have signed the Google [Individual CLA](https://cla.developers.google.com/about/google-individual), or I am covered by my company's [Corporate CLA](https://cla.developers.google.com/about/google-corporate )


jingtang10 (changes requested):

datacapture/src/test/java/com/google/android/fhir/datacapture/validation/QuestionnaireResponseItemValidatorTest.kt
```diff
@@ -96,4 +97,32 @@ class QuestionnaireResponseItemValidatorTest {
...
+
+  @Test
+  fun exceededMaxMinLength_shouldReturnInvalidResultWithMessages() {
+    val questionnaireItem =
+      Questionnaire.QuestionnaireItemComponent().apply {
+        maxLength = 10
+        addExtension(
+          Extension().apply {
+            url = MIN_LENGTH_EXTENSION_URL
+            this.setValue(IntegerType(20))
```
joiskash, line 109: Can we use `value = IntegertType(20)` instead of `this.setValue`?

datacapture/src/main/java/com/google/android/fhir/datacapture/validation/PrimitiveTypeAnswerMaxLengthValidator.kt
```diff
@@ -0,0 +1,48 @@
...
+import org.hl7.fhir.r4.model.QuestionnaireResponse
+
+/**
+ * A validator to check if the answer exceeds the maximum number of permitted characters.
+ *
+ * <p>Only primitive types are subjected to this validation.
+ */
+internal object PrimitiveTypeAnswerMaxLengthValidator : ConstraintValidator {
+  override fun validate(
+    questionnaireItem: Questionnaire.QuestionnaireItemComponent,
```
joiskash, line 31: Why are we not passing the questionnaireItemViewItem but instead passing the questionnaireItem and questionnaireResponseItem separately? Seems like that is the current design in the ConstraintValidator.. just wanted to know the thought behind this.

joiskash (commented):
Thanks @stevenckngaa , great work. I just pointed out a few nits which probably got missed out when the validation package was added. Do have a look

datacapture/src/test/java/com/google/android/fhir/datacapture/validation/QuestionnaireResponseItemValidatorTest.kt
jingtang10: agree with this. i think this is more idiomatic.

stevenckngaa: For some reason value is a val. I can't use the assignment operator. This is really strange. :(

datacapture/src/main/java/com/google/android/fhir/datacapture/validation/PrimitiveTypeAnswerMaxLengthValidator.kt
jingtang10: I see `questionnaireItemViewItem` as a UI thing. Its purpose is to encapsulate information required for the recycler view to display an item in the questionnaire item list. It does contain both the question item and the response item. But it may also contain other things that are used for the UI (for example call back to refresh the visibility of other questions depending on the answer to the current question).

The validation API here should not depend on UI classes. It can be used separately (in fact @joiskash is building a separate API for validation) so I think it would be wrong to use the questionnaireItemViewItem.

Open for discussion of course - but this is my thought process.

jingtang10 (approved):
Thanks @stevenckngaa!

jingtang10:
please run `./gradlew spotlessApply` or simply `./gradlew sA`

Issue #273, opened 2021-02-23
Questionnaire population with another resource


mberg:
This is related to the draft stage suggestion in #227. 

One of the things about SDC in FHIR is that a submission will not just create a questionnaire reponse but also potentially create corresponding resources.   We will need to figure out the logic to revert / cancel resources if a questionnaire is edited or potentially deleted.

jingtang10:
Thanks @mberg for creating this. I want to break this down a little bit more. The workflow currently is as follows

1. You have a `Questionnaire` from somewhere (loaded from db, downloaded from a server, hardcoded in app, wherever you get it from)
2. You call the sdc lib to draw the form from the `Questionnaire`
3. The user fills out the form
4. You call the sdc lib or the user clicks some button to generate a `QuestionnaireResponse`
5. You call the sdc lib to create a (as an example) `Patient`

So what does reverting mean here technically for the sdc lib? 

If it's about deleting either the `QuestionnaireResponse`, or the `Patient`, then it's totally outside the remit of the sdc lib. The app is free to do whatever they want.

So what does editing mean here technically for the sdc lib? 

That means we'll need to populate the `Questionnaire` with either the `QuestionnaireResponse` or the `Patient`.

pld:
It sounds like the SDK appropriate task here is populating a Questionnaire resources with another resource. Related to that is defining how the other resource maps to the Questionnaire resource.

@jingtang10 what do you think? does it makes sense to update/replace this issue with one more focused on that task?

jingtang10:
#377

Rkareko:
@jingtang10 the proposed strategy is to use [Expression-based Population](http://build.fhir.org/ig/HL7/sdc/populate.html#expression-based-population) as defined in the SDC spec. We shall be using FHIR path to set the initial value of a Questionnaire item an example of which is given on [this article](https://developer.commure.com/docs/apis/sdc/examples/#using-a-fhir-path-to-set-the-initial-value-of-a-questionnaire-item)

Steps

1. Add a function `populate` in the ResourceMapper which takes a `Questionnaire` and `Resource` and returns a `QuestionnaireResponse`
2. For each item in the questionnaire use `FHIRPathEngine` to evaluate the `questionnaire-initialExpression` `valueExpression.expression` and retrieve the value from the resource
3. Create an item in the `QuestionnaireResponse` for the `Questionnaire` item in step 2 and populate the `item.answer` with the retrieved value.

[...]

PR #536, opened 2021-06-04
Implement Expression-based population

MuhammadSalman-7214:

Fixes https://github.com/google/android-fhir/issues/273

**Description**
To get all the information about the patient and then generate a Questionnaire Response based on the patient details fetched using fhirpathengine

**Alternative(s) considered**
Not considered any alternatives, as this approach was already finalized for this issue
**Type**
Choose one: (Feature)

**Checklist**


jingtang10 (commented):
Thanks @MuhammadSalman-7214! Can you add some tests for this?

MuhammadSalman-7214:
sure @jingtang10

jingtang10 (changes requested):
can you revert the changes to delete the issue templates?

datacapture/src/test/java/com/google/android/fhir/datacapture/mapping/ResourceMapperTest.kt
```diff
@@ -714,4 +722,286 @@ class ResourceMapperTest {
...
+    val name = HumanName()
+    name.given = mutableListOf(StringType("Salman"))
+    name.family = "Ali"
+    val nameList: List<HumanName> = listOf(name)
+    patientResource.name = nameList
+    val contactPoint = ContactPoint()
+    contactPoint.value = "12345"
+    val contactList: List<ContactPoint> = listOf(contactPoint)
+    patientResource.telecom = contactList
+    return patientResource
```
jingtang10, line 1005:
```suggestion
    return Patient().apply {
      active = true
      birthDate = Date(1996, 8, 17)
      gender = Enumerations.AdministrativeGender.MALE
      address = Address().apply {
        ...
      }
    }
```

datacapture/build.gradle.kts
```diff
@@ -90,6 +90,7 @@ dependencies {
   implementation(Dependencies.Kotlin.stdlib)
   implementation(Dependencies.Lifecycle.viewModelKtx)
   implementation(Dependencies.material)
+  implementation(Dependencies.fhirEngine)
```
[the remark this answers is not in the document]
MuhammadSalman-7214: I have added this dependency because to test fhirpathengine.evaluate function in unit test, it requires fhir engine dependency to be added in the gradle to have it available in the data capture module, otherwise the test doesn't pass.

buildSrc/src/main/kotlin/Dependencies.kt
```diff
@@ -75,6 +75,7 @@ object Dependencies {
   const val httpInterceptor = "com.squareup.okhttp3:logging-interceptor:${Versions.httpInterceptor}"
   const val jsonToolsPatch = "com.github.java-json-tools:json-patch:${Versions.jsonToolsPatch}"
   const val material = "com.google.android.material:material:${Versions.material}"
+  const val fhirEngine = "com.google.android.fhir:engine:${Versions.fhirEngine}"
```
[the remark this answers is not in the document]
MuhammadSalman-7214: if I'll revert it here, then where can I add the fhir engine dependency?

buildSrc/src/main/kotlin/Dependencies.kt
```diff
@@ -124,6 +125,7 @@ object Dependencies {
     const val httpInterceptor = "4.0.0"
     const val jsonToolsPatch = "1.13"
     const val material = "1.3.0"
+    const val fhirEngine = "0.1.0-alpha02"
```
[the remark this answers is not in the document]
MuhammadSalman-7214: if I'll revert it here, then where can I add the fhir engine dependency?

datacapture/src/test/java/com/google/android/fhir/datacapture/mapping/ResourceMapperTest.kt
MuhammadSalman-7214: Okay

datacapture/build.gradle.kts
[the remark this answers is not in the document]
jingtang10: can you check fhir path engine's namespace? it's not in our fhir engine lib. it comes from hapi structures lib.

jingtang10:
please address the comments from the last review in the code. thanks!

jingtang10:
especially revert the deletion of the issue template files.

datacapture/build.gradle.kts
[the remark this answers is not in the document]
MuhammadSalman-7214: @jingtang10 can you please elaborate? I understand that fhirpathengine is not in our local fhir engine lib, but what should we do to make test run successfully?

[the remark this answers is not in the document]
jingtang10: you need to add dependency to hapi. you can find it in fhir engine's dependencies.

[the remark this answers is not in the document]
MuhammadSalman-7214: @jingtang10 facing the following error, if I don't include the dependency in sdc build.gradle file explicitly

java.lang.NoClassDefFoundError: org/fhir/ucum/UcumException

	at com.google.android.fhir.datacapture.mapping.ResourceMapper.extractAnswersFromExpression(ResourceMapper.kt:110)
	at com.google.android.fhir.datacapture.mapping.ResourceMapper.populate(ResourceMapper.kt:92)
	at com.google.android.fhir.datacapture.mapping.ResourceMapperTest.populateResourceAnswers(ResourceMapperTest.kt:968)
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
	at java.base/jdk.internal.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.base/java.lang.reflect.Method.invoke(Method.java:566)
	at org.junit.runners.model.FrameworkMethod$1.runReflectiveCall(FrameworkMethod.java:59)
	at org.junit.internal.runners.model.ReflectiveCallable.run(ReflectiveCallable.java:12)
	at org.junit.runners.model.FrameworkMethod.invokeExplosively(FrameworkMethod.java:56)
	at org.junit.internal.runners.statements.InvokeMethod.evaluate(InvokeMethod.java:17)
	at org.junit.runners.ParentRunner$3.evaluate(ParentRunner.java:306)
	at org.robolectric.RobolectricTestRunner$HelperTestRunner$1.evaluate(RobolectricTestRunner.java:546)
	at org.robolectric.internal.SandboxTestRunner$2.lambda$evaluate$0(SandboxTestRunner.java:252)
	at org.robolectric.internal.bytecode.Sandbox.lambda$runOnMainThread$0(Sandbox.java:89)
	at java.base/java.util.concurrent.FutureTask.run(FutureTask.java:264)
	at java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1128)
	at java.base/java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:628)
	at java.base/java.lang.Thread.run(Thread.java:834)
Caused by: java.lang.ClassNotFoundException: org.fhir.ucum.UcumException
	at org.robolectric.internal.bytecode.SandboxClassLoader.getByteCode(SandboxClassLoader.java:164)
	at org.robolectric.internal.bytecode.SandboxClassLoader.maybeInstrumentClass(SandboxClassLoader.java:119)
	at org.robolectric.internal.bytecode.SandboxClassLoader.lambda$findClass$0(SandboxClassLoader.java:112)
	at org.robolectric.util.PerfStatsCollector.measure(PerfStatsCollector.java:53)
	at org.robolectric.internal.bytecode.SandboxClassLoader.findClass(SandboxClassLoader.java:111)
	at java.base/java.lang.ClassLoader.loadClass(ClassLoader.java:589)
	at java.base/java.lang.ClassLoader.loadClass(ClassLoader.java:522)
	... 19 more
	
	whereas api(Dependencies.hapiFhirStructuresR4) is already present in sdc build.gradle file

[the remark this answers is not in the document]
jingtang10: ok that's odd -- it maybe because of this file that is missing https://github.com/google/android-fhir/blob/master/engine/src/main/java/org/fhir/ucum/UcumException.kt

Can you try to upgrade our hapi dependency to 5.4.0 which was released end of may?

If that doesn't fix it can you update this bug https://github.com/hapifhir/hapi-fhir/issues/2443 and also send an email to the hapi fhir discussion group? https://groups.google.com/g/hapi-fhir?pli=1

also try to include the ucum exeption file to the sdc lib maybe the only other local fix we can do (and any client using both engine and sdc migth need to exclude the dupilcated file which might be a bit nasty...)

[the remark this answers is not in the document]
MuhammadSalman-7214: Thanks @jingtang10. After adding UcumException to sdc and also adding caffeine dependency, it is now working and I have removed the explicitly added dependency of fhir engine from sdc build.gradle as well. I have pushed the PR as well. You can review. Thanks.

datacapture/src/main/java/com/google/android/fhir/datacapture/mapping/ResourceMapper.kt
```diff
@@ -66,6 +76,87 @@ object ResourceMapper {
     }
   }
 
+  fun populate(questionnaire: Questionnaire, resource: Resource): HashMap<String, Any> {
+    val expressionMap = HashMap<String, String>()
+    val questionnaireItemList = questionnaire.item[0].item
```
jingtang10, line 81: why only care about the first item?

datacapture/src/main/java/com/google/android/fhir/datacapture/mapping/ResourceMapper.kt
```diff
@@ -66,6 +76,87 @@ object ResourceMapper {
...
+      }
+    }
+
+    return extractAnswersFromExpression(expressionMap, resource)
+  }
+
+  private fun fetchExpression(
+    question: Questionnaire.QuestionnaireItemComponent,
+    expressionMap: HashMap<String, String>
+  ) {
```
jingtang10, line 98: two things about this function signature

1. i dislike the idea that you are using an implicit out param... why not just return a map?
2. i think this should be an extension function on `QuestionnaireItemComponent`

datacapture/src/main/java/com/google/android/fhir/datacapture/mapping/ResourceMapper.kt
```diff
@@ -66,6 +76,87 @@ object ResourceMapper {
...
+  ): HashMap<String, Any> {
+    val context = FhirContext.forR4()
+    val fhirPathEngine =
+      FHIRPathEngine(HapiWorkerContext(context, DefaultProfileValidationSupport(context)))
+    val answersHashMap = HashMap<String, Any>()
+    val expressionMapIterator = expressionMap.keys.iterator()
+    while (expressionMapIterator.hasNext()) {
+      val key = expressionMapIterator.next()
+      val answerExtracted = fhirPathEngine.evaluate(resource, expressionMap[key])
+      val questionType = answerExtracted[0]
```
jingtang10, line 117: shouldn't all the extracte values be used if the question has repeated answers?

datacapture/src/main/java/com/google/android/fhir/datacapture/mapping/ResourceMapper.kt
```diff
@@ -66,6 +76,87 @@ object ResourceMapper {
...
+        questionType.isDateTime -> {
+          answersHashMap[key] = (questionType as DateType)
+        }
+        questionType.isBooleanPrimitive -> {
+          answersHashMap[key] = (questionType as BooleanType)
+        }
+        questionType.isPrimitive -> {
+          val gender = (questionType as Enumeration<*>).code
+          answersHashMap[key] = StringType(gender)
+        }
```
jingtang10, line 128: always extract this type of operation into a utility function. this is making the logic of this function very hard to read.

jingtang10 (changes requested):
thanks @MuhammadSalman-7214 - some high-level comments first.

can you ask @ekigamba and/or @shoaibmushtaq25 to leave a review since they worked on the extraction process? thanks!

datacapture/build.gradle.kts
```diff
@@ -67,6 +67,11 @@ android {
     // See https://developer.android.com/studio/write/java8-support
     jvmTarget = JavaVersion.VERSION_1_8.toString()
   }
+
+  packagingOptions {
+    exclude("META-INF/ASL-2.0.txt")
+    exclude("META-INF/LGPL-3.0.txt")
+  }
```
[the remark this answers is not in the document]
MuhammadSalman-7214: if I don't add this, then files duplication error comes

.idea/jarRepositories.xml
```diff
@@ -1,61 +0,0 @@
-<?xml version="1.0" encoding="UTF-8" ?>
```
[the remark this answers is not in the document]
MuhammadSalman-7214: was facing some problems in code, delete the build, ideas folder, because of which this file got removed

MuhammadSalman-7214:
> thanks @MuhammadSalman-7214 - some high-level comments first.
> 
> can you ask @ekigamba and/or @shoaibmushtaq25 to leave a review since they worked on the extraction process? thanks!

Sure @jingtang10

datacapture/src/main/java/com/google/android/fhir/datacapture/mapping/ResourceMapper.kt
MuhammadSalman-7214: currently I am only handling it for a single answer, not including multiple answers

datacapture/src/main/java/com/google/android/fhir/datacapture/mapping/ResourceMapper.kt
MuhammadSalman-7214: because all questionnaires currently have questions inside first item, so that is why just taking the first item, and then catering all the questions within it

datacapture/src/main/java/com/google/android/fhir/datacapture/mapping/ResourceMapper.kt
```diff
@@ -66,6 +76,87 @@ object ResourceMapper {
     }
   }
 
+  fun populate(questionnaire: Questionnaire, resource: Resource): HashMap<String, Any> {
```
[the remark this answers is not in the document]
MuhammadSalman-7214: The reason not generating QuestionnaireResponse is that QuestionnaireResponse wants answers against each and every link Id, whereas we don't have answer for every link Id that is present in the Questionnaire, so I am returning Hashmap, and setting the values as initial values to Questionnaire.

datacapture/src/main/java/com/google/android/fhir/datacapture/mapping/ResourceMapper.kt
MuhammadSalman-7214: okay

datacapture/src/main/java/com/google/android/fhir/datacapture/mapping/ResourceMapper.kt
MuhammadSalman-7214: okay

shoaibmushtaq25:
I have reviewed this PR and it looks good to me. Let's wait for @ekigamba to also review it as he did the major work on ResourceMapper in parallel to me. cc: @MuhammadSalman-7214 @jingtang10

MuhammadSalman-7214:
@jingtang10 update the PR as per your comments, kindly review, thanks.

datacapture/src/main/java/com/google/android/fhir/datacapture/mapping/ResourceMapper.kt
ekigamba: I also think that you should care about all the items

datacapture/src/main/java/com/google/android/fhir/datacapture/mapping/ResourceMapper.kt
```diff
@@ -66,6 +77,119 @@ object ResourceMapper {
     }
   }
 
+  fun populate(questionnaire: Questionnaire, resource: Resource): QuestionnaireResponse {
```
ekigamba, line 80: 1. Change the PR title to `Implement Expression-based population`
2. Is this supposed to support population from multiple resources eg. edit patient functionality will require populating using both the Patient and RelatedPerson. I think this is clearly stated https://github.com/opensrp/fhircore/issues/184 and https://github.com/opensrp/fhircore/issues/185

datacapture/src/main/java/com/google/android/fhir/datacapture/mapping/ResourceMapper.kt
```diff
@@ -66,6 +77,119 @@ object ResourceMapper {
...
+    val expressionMap = HashMap<String, String>()
+    val questionnaireItemList = questionnaire.item[0].item
```
ekigamba, line 82: I think that Expression-based population should only be activated if there is a `questionnaire-launchContext`. I still think that this should be OK since the application knows the Resource used in population

cc @Rkareko

ekigamba (changes requested):

datacapture/src/main/java/com/google/android/fhir/datacapture/mapping/ResourceMapper.kt
```diff
@@ -66,6 +77,119 @@ object ResourceMapper {
...
+    val questionnaireItemListIterator = questionnaireItemList.iterator()
+    while (questionnaireItemListIterator.hasNext()) {
+      val singleQuestion = questionnaireItemListIterator.next()
+      if (singleQuestion.type == Questionnaire.QuestionnaireItemType.GROUP) {
+        singleQuestion.item.forEach { question ->
```
ekigamba, line 87: I think we have a case where we might have multiple levels eg. `Patient.Contact.HumanName` where HumanName group has it's own questions i.e. given and family. The example can be found in the `ResourceMapperTest#extract()`

datacapture/src/main/java/com/google/android/fhir/datacapture/mapping/ResourceMapper.kt
MuhammadSalman-7214: @ekigamba I have done that for editing of Related Person as well, will be pushing it shortly

datacapture/src/main/java/com/google/android/fhir/datacapture/mapping/ResourceMapper.kt
```diff
@@ -66,6 +77,119 @@ object ResourceMapper {
     }
   }
 
+  fun populate(questionnaire: Questionnaire, resource: Resource): QuestionnaireResponse {
+    val expressionMap = HashMap<String, String>()
+    val questionnaireItemList = questionnaire.item[0].item
+    val questionnaireItemListIterator = questionnaireItemList.iterator()
+    while (questionnaireItemListIterator.hasNext()) {
+      val singleQuestion = questionnaireItemListIterator.next()
+      if (singleQuestion.type == Questionnaire.QuestionnaireItemType.GROUP) {
```
ekigamba, line 86: If we decide to support population from multiple resources, we might need to support the extension `questionnaire-itemPopulationContext`. I think @Rkareko and @jingtang10 can provided better direction on this. One of it's purpose would be switching the resource population context

ekigamba (changes requested):

datacapture/src/main/java/com/google/android/fhir/datacapture/mapping/ResourceMapper.kt
MuhammadSalman-7214: but that already has been cared of in this extension function fetchExpression() which will find expression in every item in a group

datacapture/src/main/java/com/google/android/fhir/datacapture/mapping/ResourceMapper.kt
MuhammadSalman-7214: I don't think so we need that, as fetchexpression() is already fetching extensions for GROUP as well as single items base on the loop, but still we can discuss

datacapture/src/main/java/com/google/android/fhir/datacapture/mapping/ResourceMapper.kt
MuhammadSalman-7214: as we have to define initial expression extension for all the questions that we want an answer or that we want to edit

datacapture/src/main/java/com/google/android/fhir/datacapture/mapping/ResourceMapper.kt
MuhammadSalman-7214: this should be okay then as we are checking for the ITEM_CONTEXT_EXTENSION_URL for the same way as above

datacapture/build.gradle.kts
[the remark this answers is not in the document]
ekigamba: Does this still happen even with the removed fhir engine dependency?

datacapture/src/main/java/com/google/android/fhir/datacapture/mapping/ResourceMapper.kt
```diff
@@ -66,6 +77,119 @@ object ResourceMapper {
...
+      }
+      this.isBooleanPrimitive -> {
+        return (this as BooleanType)
+      }
+      this.isPrimitive -> {
+        return StringType((this as Enumeration<*>).code)
+      }
+      this is HumanName -> {
+        return if (questionLinkId.contains("family")) {
+          StringType(this.family)
```
ekigamba, line 174: I see that we are returning a `StringType`. If the type is `HumanName` is the type returned supposed to be a `StringType` or a `HumanName`?

datacapture/src/main/java/com/google/android/fhir/datacapture/mapping/ResourceMapper.kt
```diff
@@ -66,6 +77,119 @@ object ResourceMapper {
...
+        return (this as DateType)
+      }
+      this.isBooleanPrimitive -> {
+        return (this as BooleanType)
+      }
+      this.isPrimitive -> {
+        return StringType((this as Enumeration<*>).code)
+      }
+      this is HumanName -> {
+        return if (questionLinkId.contains("family")) {
```
ekigamba, line 173: This should not be dependent on the linkId. Kindly change this to something more generic

datacapture/src/main/java/com/google/android/fhir/datacapture/mapping/ResourceMapper.kt
```diff
@@ -66,6 +77,119 @@ object ResourceMapper {
...
+          StringType(this.family)
+        } else {
+          this.given[0]
+        }
+      }
+      this is Address -> {
+        return if (questionLinkId.contains("country")) {
```
ekigamba, line 180: This should not be dependent on the linkId. Kindly change this to something more generic

datacapture/src/main/java/com/google/android/fhir/datacapture/mapping/ResourceMapper.kt
```diff
@@ -66,6 +77,119 @@ object ResourceMapper {
...
+          StringType(this.country)
+        } else {
+          StringType(this.city)
+        }
+      }
+      this is ContactPoint -> {
+        return StringType(this.value)
```
ekigamba, line 187: It is also not clear why this is returning a `StringType` if the expected type is `ContactPoint`. I see that we return the correct types here https://github.com/google/android-fhir/blob/feature/add-edit-patient-functionality/datacapture/src/main/java/com/google/android/fhir/datacapture/mapping/ResourceMapper.kt#L163

datacapture/src/test/java/com/google/android/fhir/datacapture/mapping/ResourceMapperTest.kt
```diff
@@ -714,4 +722,289 @@ class ResourceMapperTest {
...
+                  "url": "http://hl7.org/fhir/uv/sdc/StructureDefinition/sdc-questionnaire-initialExpression",
+                  "valueExpression": {
+                    "language": "text/fhirpath",
+                    "expression": "Patient.address",
+                    "name": "patientCity"
+                  }
+                }
+              ],
+              "linkId": "PR-address-country",
+              "definition": "http://hl7.org/fhir/StructureDefinition/Patient#Patient.address.country",
```
ekigamba, line 933: Add a question for `Patient.contact` with `Patient.contact.HumanName`

ekigamba (changes requested):

datacapture/src/test/java/com/google/android/fhir/datacapture/mapping/ResourceMapperTest.kt
```diff
@@ -714,4 +722,289 @@ class ResourceMapperTest {
...
+                }
+              ]
+            },
+            {
+              "extension": [
+                {
+                  "url": "http://hl7.org/fhir/uv/sdc/StructureDefinition/sdc-questionnaire-initialExpression",
+                  "valueExpression": {
+                    "language": "text/fhirpath",
+                    "expression": "Patient.telecom",
```
ekigamba, line 878: I believe this should be something like `Patient.telecom.value`

datacapture/src/main/java/com/google/android/fhir/datacapture/mapping/ResourceMapper.kt
MuhammadSalman-7214: it's supposed to be a String type here as I'll set the String type value based on Question linkId

datacapture/src/main/java/com/google/android/fhir/datacapture/mapping/ResourceMapper.kt
MuhammadSalman-7214: yeah can you suggest what should it be?

datacapture/src/main/java/com/google/android/fhir/datacapture/mapping/ResourceMapper.kt
MuhammadSalman-7214: yeah can you suggest what should it be?

datacapture/src/main/java/com/google/android/fhir/datacapture/mapping/ResourceMapper.kt
MuhammadSalman-7214: because I'll set all these values as per their link Id's. Will elaborate you on call

ekigamba (changes requested):

datacapture/src/test/java/com/google/android/fhir/datacapture/mapping/ResourceMapperTest.kt
```diff
@@ -714,4 +722,289 @@ class ResourceMapperTest {
...
+            }
+          ],
+          "item": [
+            {
+              "extension": [
+                {
+                  "url": "http://hl7.org/fhir/uv/sdc/StructureDefinition/sdc-questionnaire-initialExpression",
+                  "valueExpression": {
+                    "language": "text/fhirpath",
+                    "expression": "Patient.name",
```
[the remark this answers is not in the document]
MuhammadSalman-7214: we cannot say, it is wrong, though we can use it directly as Patient.name.given as well

datacapture/src/test/java/com/google/android/fhir/datacapture/mapping/ResourceMapperTest.kt
MuhammadSalman-7214: no it does not work with this as the value is for contact point not for telecome

datacapture/src/test/java/com/google/android/fhir/datacapture/mapping/ResourceMapperTest.kt
```diff
@@ -714,4 +722,289 @@ class ResourceMapperTest {
...
+              "type": "string",
+              "text": "City"
+            },
+            {
+              "extension": [
+                {
+                  "url": "http://hl7.org/fhir/uv/sdc/StructureDefinition/sdc-questionnaire-initialExpression",
+                  "valueExpression": {
+                    "language": "text/fhirpath",
+                    "expression": "Patient.address",
```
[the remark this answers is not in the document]
MuhammadSalman-7214: private fun Base.getType(questionLinkId: String): Type? {
    when {
      this.isDateTime -> {
        return (this as DateType)
      }
      this.isBooleanPrimitive -> {
        return (this as BooleanType)
      }
      this.isPrimitive -> {
        return StringType((this as Enumeration<*>).code)
      }
      this is HumanName -> {
        return if (questionLinkId.contains("family") && questionLinkId.contains("related")) {
          StringType(this.family)
        } else if (questionLinkId.contains("name") && questionLinkId.contains("related")) {
          this.given[0]
        } else if (questionLinkId.contains("family")) {
          StringType(this.family)
        } else {
          this.given[0]
        }
      }
      this is Address -> {
        return if (questionLinkId.contains("country")) {
          StringType(this.country)
        } else {
          StringType(this.city)
        }
      }
      this is ContactPoint -> {
        return StringType(this.value)
      }
    }
    return null
  }
  this function is already calling from populate function, so I don't think there is a separate need for this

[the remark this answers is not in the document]
MuhammadSalman-7214: for all your above comments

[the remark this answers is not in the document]
MuhammadSalman-7214: only thing to change is to make linkId thing more generic

datacapture/src/test/java/com/google/android/fhir/datacapture/mapping/ResourceMapperTest.kt
```diff
@@ -714,4 +722,289 @@ class ResourceMapperTest {
...
+          "type": "string",
+          "text": "Gender"
+        },
+        {
+          "linkId": "PR-telecom",
+          "type": "group",
+          "definition": "http://hl7.org/fhir/StructureDefinition/Patient#Patient.telecom",
+          "extension": [
+            {
+              "url": "http://hl7.org/fhir/uv/sdc/StructureDefinition/sdc-questionnaire-itemContext",
```
ekigamba, line 845: Kindly remove these extensions that specify the Base type. This is no longer supported and the extraction should work without this

datacapture/src/test/java/com/google/android/fhir/datacapture/mapping/ResourceMapperTest.kt
```diff
@@ -714,4 +722,289 @@ class ResourceMapperTest {
...
+            }
+          ],
+          "item": [
+            {
+              "extension": [
+                {
+                  "url": "http://hl7.org/fhir/uv/sdc/StructureDefinition/sdc-questionnaire-initialExpression",
+                  "valueExpression": {
+                    "language": "text/fhirpath",
+                    "expression": "Patient.address",
```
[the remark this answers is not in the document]
ekigamba: Is there an official example that can lead on how the population of compound/composite questions is supposed to be authored?

ekigamba (changes requested):

datacapture/src/test/java/com/google/android/fhir/datacapture/mapping/ResourceMapperTest.kt
```diff
@@ -714,4 +722,289 @@ class ResourceMapperTest {
...
+              "type": "string",
+              "text": "First Name"
+            },
+            {
+              "extension": [
+                {
+                  "url": "http://hl7.org/fhir/uv/sdc/StructureDefinition/sdc-questionnaire-initialExpression",
+                  "valueExpression": {
+                    "language": "text/fhirpath",
+                    "expression": "Patient.name",
```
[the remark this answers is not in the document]
Rkareko: We should look into creating a FHIRPath expression that fetches the family name

datacapture/src/main/java/com/google/android/fhir/datacapture/mapping/ResourceMapper.kt
jingtang10: I support doing multiple resources in a following PR (trying to separate things)

datacapture/src/main/java/com/google/android/fhir/datacapture/mapping/ResourceMapper.kt
jingtang10: thanks @MuhammadSalman-7214 I think in your reply you just stated _what it is now_ rather than responding to the suggest of _what it should/can/ought to be_.

datacapture/src/main/java/com/google/android/fhir/datacapture/mapping/ResourceMapper.kt
jingtang10: could potentially add a check and throw an exception @ekigamba what do you think?

e.g. `check(questionnaire.hasLaunchContextEtc(), "Does not have launch context etc etc")`

datacapture/src/test/java/com/google/android/fhir/datacapture/mapping/ResourceMapperTest.kt
jingtang10: @ekigamba can this be `Patient.telecom.where(system='phone')` or something similar?

datacapture/src/test/java/com/google/android/fhir/datacapture/mapping/ResourceMapperTest.kt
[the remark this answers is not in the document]
MuhammadSalman-7214: @jingtang10 I mean to say we can extract the value of name in both ways as Patient.name.given from expression as well as from Patient.name expression as well, are you clear on it now?

MuhammadSalman-7214:
@ekigamba and @jingtang10 all changes are done as per your comments, please review, thanks.

datacapture/src/main/java/com/google/android/fhir/datacapture/mapping/ResourceMapper.kt
ekigamba: I agree @jingtang10, I guess we can probably have helper functions on the SDK. It seems that the extension is not referenced on this function since this happens outside this function?

datacapture/src/test/java/com/google/android/fhir/datacapture/mapping/ResourceMapperTest.kt
ekigamba: @jingtang10 Yea, I think that's a better expression since `Patient.telecom` is a list

datacapture/src/main/java/com/google/android/fhir/datacapture/mapping/ResourceMapper.kt
```diff
@@ -78,56 +76,77 @@ object ResourceMapper {
...
     val questionnaireResponse: QuestionnaireResponse =
       QuestionnaireResponse().apply { questionnaire = questions.id }
-
     questions.item.forEach { questionnaireResponse.addItem(it.createQuestionnaireResponseItem()) }
     return questionnaireResponse
   }
 
-  private fun setInitialValueToQuestionnaire(
-    answersHashMap: HashMap<String, Any>,
+  private fun setInitialValue(
```
ekigamba, line 95: I also think this function name should be `setInitialValueForItemList`

ekigamba (changes requested):

datacapture/src/main/java/com/google/android/fhir/datacapture/mapping/ResourceMapper.kt
```diff
@@ -78,56 +76,77 @@ object ResourceMapper {
...
+    }
+    return expressionPair
+  }
+
+  private fun setInitialValueToAllItems(
+    question: Questionnaire.QuestionnaireItemComponent,
+    answersHashMap: HashMap<String, Type>
+  ) {
+    if (question.type == Questionnaire.QuestionnaireItemType.GROUP) {
+      setInitialValue(question.item, answersHashMap)
```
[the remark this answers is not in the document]
MuhammadSalman-7214: thanks

datacapture/src/main/java/com/google/android/fhir/datacapture/mapping/ResourceMapper.kt
```diff
@@ -78,56 +76,77 @@ object ResourceMapper {
...
+  ): HashMap<String, String> {
+    val itemsIterator = questionnaire.iterator()
+    while (itemsIterator.hasNext()) {
+      val pair = fetchExpressionsFromQuestionnaire(itemsIterator.next())
+      pair?.let { expressionMap[it.first] = it.second }
+    }
+    return expressionMap
+  }
+
+  private fun fetchExpressionsFromQuestionnaire(
```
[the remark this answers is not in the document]
MuhammadSalman-7214: okay

datacapture/src/main/java/com/google/android/fhir/datacapture/mapping/ResourceMapper.kt
```diff
@@ -78,56 +76,77 @@ object ResourceMapper {
...
+    questions: List<Questionnaire.QuestionnaireItemComponent>,
+    answersHashMap: HashMap<String, Type>
+  ) {
+    val itemsIterator = questions.iterator()
+    while (itemsIterator.hasNext()) {
+      setInitialValueToAllItems(itemsIterator.next(), answersHashMap)
+    }
+  }
+
+  private fun fetchExpressions(
```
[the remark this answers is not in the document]
MuhammadSalman-7214: okay

datacapture/src/main/java/com/google/android/fhir/datacapture/mapping/ResourceMapper.kt
```diff
@@ -78,56 +76,77 @@ object ResourceMapper {
...
+      pair?.let { expressionMap[it.first] = it.second }
+    }
+    return expressionMap
+  }
+
+  private fun fetchExpressionsFromQuestionnaire(
+    question: Questionnaire.QuestionnaireItemComponent
+  ): Pair<String, String>? {
+    var expressionPair: Pair<String, String>? = null
+    if (question.type == Questionnaire.QuestionnaireItemType.GROUP) {
```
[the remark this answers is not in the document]
MuhammadSalman-7214: thanks :-)

datacapture/src/main/java/com/google/android/fhir/datacapture/mapping/ResourceMapper.kt
```diff
@@ -78,56 +76,77 @@ object ResourceMapper {
...
+      fetchExpressions(question.item)
+    } else {
+      question.fetchExpression()?.let { exp ->
+        expressionPair = Pair(question.linkId, exp.expression)
+      }
+    }
+    return expressionPair
+  }
+
+  private fun setInitialValueToAllItems(
```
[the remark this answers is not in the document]
MuhammadSalman-7214: okay

datacapture/src/main/java/com/google/android/fhir/datacapture/mapping/ResourceMapper.kt
MuhammadSalman-7214: okay

datacapture/src/main/java/com/google/android/fhir/datacapture/mapping/ResourceMapper.kt
```diff
@@ -70,6 +66,8 @@ object ResourceMapper {
    * This method assumes there is only one FHIR resource to be extracted from the given
    * `questionnaire` and `questionnaireResponse`.
    */
+  private val expressionMap: HashMap<String, String> = hashMapOf()
```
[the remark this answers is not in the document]
MuhammadSalman-7214: okay, that is fine

ekigamba (approved):

datacapture/src/main/java/com/google/android/fhir/datacapture/mapping/ResourceMapper.kt
```diff
@@ -66,6 +73,109 @@ object ResourceMapper {
...
+  }
+
+  private fun setInitialValueForItemList(
+    questions: List<Questionnaire.QuestionnaireItemComponent>,
+    answersHashMap: HashMap<String, Type>
+  ) {
+    val itemsIterator = questions.iterator()
+    while (itemsIterator.hasNext()) {
+      setInitialValueForItem(itemsIterator.next(), answersHashMap)
+    }
```
jingtang10, line 100: no need to use iterator.

```
questions.forEach {
  setInitialValueForItem(it, answersHashMap)
}
```

jingtang10: Actually as per my previoius comment, I really think you should avoid using output parameter (i.e. `answerHashMap`). Can you please make that a return value?

datacapture/src/main/java/com/google/android/fhir/datacapture/mapping/ResourceMapper.kt
```diff
@@ -66,6 +73,109 @@ object ResourceMapper {
     }
   }
 
+  fun populate(questionnaire: Questionnaire, resource: Resource): QuestionnaireResponse {
+    val expressionMap = fetchExpressionsFromItemsList(questionnaire.item, hashMapOf())
+    val answersHashMap = extractAnswersFromExpression(expressionMap, resource)
+    return createQuestionnaireResponse(questionnaire, answersHashMap)
```
jingtang10, line 79: a suggestion: you are basically going through all the questions in the questionnaire and creating an expression map, and then you're going through it to create the answer map, and then you're going through the questions in the questionnaire again to create the response. why not just do this in one traverse?

jingtang10 (changes requested):
Thanks @MuhammadSalman-7214!

MuhammadSalman-7214:
https://github.com/google/android-fhir/pull/536#discussion_r657893734
@jingtang10 for this, I am using questionnaire, if I rename the variable to questionnaire then there is a problem with actual questionnaire variable as this line:
QuestionnaireResponse().apply { questionnaire = questions.id }

datacapture/src/main/java/com/google/android/fhir/datacapture/QuestionnaireViewModel.kt
```diff
@@ -231,7 +231,7 @@ private fun QuestionnaireResponse.QuestionnaireResponseItemComponent.addNestedIt
  * The hierarchy and order of child items will be retained as specified in the standard. See
  * https://www.hl7.org/fhir/questionnaireresponse.html#notes for more details.
  */
-private fun Questionnaire.QuestionnaireItemComponent.createQuestionnaireResponseItem():
+fun Questionnaire.QuestionnaireItemComponent.createQuestionnaireResponseItem():
```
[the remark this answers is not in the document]
MuhammadSalman-7214: okay sure

datacapture/src/main/java/com/google/android/fhir/datacapture/mapping/ResourceMapper.kt
MuhammadSalman-7214: okay

MuhammadSalman-7214:
@jingtang10 I have updated the PR as per your comments except the suggestion comment. If that is necessary then I'll do that as well otherwise you can approve the PR or please tell me to implement that comment as well, thank you very much.

MuhammadSalman-7214:
https://github.com/google/android-fhir/pull/536#discussion_r657893564
For this, I have added this file as well, but it is added in gitignore, so that is why not pushing

jingtang10:
> [#536 (comment)](https://github.com/google/android-fhir/pull/536#discussion_r657893734)
> @jingtang10 for this, I am using questionnaire, if I rename the variable to questionnaire then there is a problem with actual questionnaire variable at this line:
> QuestionnaireResponse().apply { questionnaire = questions.id }

what's the error? i think the error might be type mismatch rather than variable naming issue. Even if it is a naming conflict, you can typically apply scope to clarify (use `@`)?

jingtang10:
> @jingtang10 I have updated the PR as per your comments except the suggestion comment. If that is necessary then I'll do that as well otherwise you can approve the PR or please tell me to implement that comment as well, thank you very much.

yes please implement it

datacapture/src/main/java/com/google/android/fhir/datacapture/mapping/ResourceMapper.kt
MuhammadSalman-7214: > a suggestion: you are basically going through all the questions in the questionnaire and creating an expression map, and then you're going through it to create the answer map, and then you're going through the questions in the questionnaire again to create the response. why not just do this in one traverse?

@jingtang10 I have pushed the code as per your suggestion, please review, thanks

MuhammadSalman-7214:
@jingtang10 code is pushed, you can check the PR, thanks

datacapture/src/main/java/com/google/android/fhir/datacapture/mapping/ResourceMapper.kt
```diff
@@ -66,6 +73,60 @@ object ResourceMapper {
...
+    questionsList: List<Questionnaire.QuestionnaireItemComponent>,
+    resource: Resource
+  ) {
+    val itemsIterator = questionsList.iterator()
+    while (itemsIterator.hasNext()) {
+      evaluatePopulateExpressionOnQuestionnaireItem(itemsIterator.next(), resource)
+    }
+  }
+
+  private fun evaluatePopulateExpressionOnQuestionnaireItem(
```
jingtang10, line 99:
```suggestion
  private fun populateInitialValue(
```

datacapture/src/main/java/com/google/android/fhir/datacapture/mapping/ResourceMapper.kt
```diff
@@ -66,6 +73,60 @@ object ResourceMapper {
...
+  fun populate(questionnaire: Questionnaire, resource: Resource): QuestionnaireResponse {
+    val questionnaireResponse = QuestionnaireResponse()
+    evaluatePopulateExpressionsInItemList(questionnaire.item, resource)
+    questionnaire.item.forEach {
+      questionnaireResponse.addItem(it.createQuestionnaireResponseItem())
+    }
+    return questionnaireResponse
+  }
+
+  private fun evaluatePopulateExpressionsInItemList(
```
jingtang10, line 89:
```suggestion
  private fun populateIntialValue(
```

datacapture/src/main/java/com/google/android/fhir/datacapture/mapping/ResourceMapper.kt
```diff
@@ -66,6 +73,60 @@ object ResourceMapper {
...
+    questionsList: List<Questionnaire.QuestionnaireItemComponent>,
```
jingtang10, line 90:
```suggestion
    questionnaireItems: List<Questionnaire.QuestionnaireItemComponent>,
```

datacapture/src/main/java/com/google/android/fhir/datacapture/mapping/ResourceMapper.kt
```diff
@@ -66,6 +73,60 @@ object ResourceMapper {
...
+    resource: Resource
+  ) {
+    val itemsIterator = questionsList.iterator()
+    while (itemsIterator.hasNext()) {
+      evaluatePopulateExpressionOnQuestionnaireItem(itemsIterator.next(), resource)
+    }
```
jingtang10, line 96:
```suggestion
    questionnaireItems.forEach {
      populateInitialValue(it, resource)
    }
```

datacapture/src/main/java/com/google/android/fhir/datacapture/MoreQuestionnaireItemComponents.kt
```diff
@@ -67,3 +67,82 @@ internal val Questionnaire.QuestionnaireItemComponent.localizedText: String?
...
   get() = prefixElement?.getLocalizedText()
+
+/**
+ * Creates a [QuestionnaireResponse.QuestionnaireResponseItemComponent] from the provided
+ * [Questionnaire.QuestionnaireItemComponent].
+ *
+ * The hierarchy and order of child items will be retained as specified in the standard. See
+ * https://www.hl7.org/fhir/questionnaireresponse.html#notes for more details.
+ */
+fun Questionnaire.QuestionnaireItemComponent.createQuestionnaireResponseItem():
```
jingtang10, line 78: this is where the function belongs - great!

can i ask you for a favour @MuhammadSalman-7214? is it possible for you to add some unit tests for this? thanks!

jingtang10 (changes requested):
thanks! i think the code structure is much better this way.

we discussed the issue with the removed file offline - if you can resolved that that'd be great. thanks!

MuhammadSalman-7214:
@jingtang10 PR updated as per your comments, and jarRepositories.xml file added as well, though it is showing some changes, but that is of format only as I have applied spotlessApply. you can check the changes, thank you very much.

datacapture/src/main/java/com/google/android/fhir/datacapture/MoreQuestionnaireItemComponents.kt
MuhammadSalman-7214: can you tell me in which feature, I can write unit test for this function? thanks

datacapture/src/main/java/com/google/android/fhir/datacapture/mapping/ResourceMapper.kt
MuhammadSalman-7214: Done

datacapture/src/main/java/com/google/android/fhir/datacapture/mapping/ResourceMapper.kt
MuhammadSalman-7214: Done

datacapture/src/main/java/com/google/android/fhir/datacapture/mapping/ResourceMapper.kt
MuhammadSalman-7214: Done

datacapture/src/main/java/com/google/android/fhir/datacapture/mapping/ResourceMapper.kt
MuhammadSalman-7214: Done

datacapture/src/main/java/com/google/android/fhir/datacapture/MoreQuestionnaireItemComponents.kt
```diff
@@ -67,3 +67,82 @@ internal val Questionnaire.QuestionnaireItemComponent.localizedText: String?
...
+  questionnaireItemComponent: Questionnaire.QuestionnaireItemComponent
+) {
+  if (answer.isNotEmpty()) {
+    answer.first().item = questionnaireItemComponent.listOfItemInAnswer()
+  }
+}
+
+/**
+ * Creates a List of [QuestionnaireResponse.QuestionnaireResponseItemComponent] from the provided
+ * [Questionnaire.QuestionnaireItemComponent].
```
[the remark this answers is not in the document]
MuhammadSalman-7214: Done

datacapture/src/main/java/com/google/android/fhir/datacapture/MoreQuestionnaireItemComponents.kt
```diff
@@ -67,3 +67,82 @@ internal val Questionnaire.QuestionnaireItemComponent.localizedText: String?
...
+ *
+ * The hierarchy and order of child items will be retained as specified in the standard. See
+ * https://www.hl7.org/fhir/questionnaireresponse.html#notes for more details.
+ */
+private inline fun Questionnaire.QuestionnaireItemComponent.listOfItemInAnswer() =
```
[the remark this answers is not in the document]
MuhammadSalman-7214: Done

datacapture/src/main/java/com/google/android/fhir/datacapture/MoreQuestionnaireItemComponents.kt
```diff
@@ -67,3 +67,82 @@ internal val Questionnaire.QuestionnaireItemComponent.localizedText: String?
...
+  item.map { it.createQuestionnaireResponseItem() }.toList()
```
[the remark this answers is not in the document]
MuhammadSalman-7214: Done

datacapture/src/main/java/com/google/android/fhir/datacapture/mapping/ResourceMapper.kt
```diff
@@ -66,6 +73,60 @@ object ResourceMapper {
...
+   * This function takes Questionnaire and Resource and returns Questionnaire Response that we can
+   * populate to Questionnaire
+   */
+  fun populate(questionnaire: Questionnaire, resource: Resource): QuestionnaireResponse {
+    val questionnaireResponse = QuestionnaireResponse()
+    evaluatePopulateExpressionsInItemList(questionnaire.item, resource)
+    questionnaire.item.forEach {
+      questionnaireResponse.addItem(it.createQuestionnaireResponseItem())
+    }
+    return questionnaireResponse
```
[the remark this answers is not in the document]
MuhammadSalman-7214: Done

datacapture/src/main/java/com/google/android/fhir/datacapture/mapping/ResourceMapper.kt
```diff
@@ -66,6 +73,60 @@ object ResourceMapper {
...
+        }
+      }
+    }
+  }
+
+  private fun Questionnaire.QuestionnaireItemComponent.fetchExpression(): Expression? {
+    this.extension.filter { it.url == ITEM_INITIAL_EXPRESSION_URL }.map {
+      return it.value as Expression
+    }
+    return null
```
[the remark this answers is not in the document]
MuhammadSalman-7214: Done

datacapture/src/main/java/com/google/android/fhir/datacapture/mapping/ResourceMapper.kt
```diff
@@ -66,6 +73,60 @@ object ResourceMapper {
...
+          initial =
+            mutableListOf(
+              Questionnaire.QuestionnaireItemInitialComponent().setValue(answerExtracted)
+            )
+        }
+      }
+    }
+  }
+
+  private fun Questionnaire.QuestionnaireItemComponent.fetchExpression(): Expression? {
```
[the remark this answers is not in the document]
MuhammadSalman-7214: Done

datacapture/src/main/java/com/google/android/fhir/datacapture/mapping/ResourceMapper.kt
```diff
@@ -66,6 +73,55 @@ object ResourceMapper {
     }
   }
 
+  /**
+   * This function takes Questionnaire and Resource and returns Questionnaire Response that we can
+   * populate to Questionnaire
```
jingtang10, line 78: You don't need to re-describe what the function signature is. Rather explain what it does:

```
Returns a `QuestionnaireResponse` to the [questionnaire] that is pre-filled from the [resource].

See http://build.fhir.org/ig/HL7/sdc/populate.html#expression-based-population.
```

datacapture/src/main/java/com/google/android/fhir/datacapture/mapping/ResourceMapper.kt
```diff
@@ -66,6 +73,55 @@ object ResourceMapper {
...
+    resource: Resource
+  ) {
+    questionnaireItems.forEach { populateInitialValue(it, resource) }
+  }
+
+  private fun populateInitialValue(
+    question: Questionnaire.QuestionnaireItemComponent,
+    resource: Resource
+  ) {
+    val expressionMap: HashMap<String, String> = hashMapOf()
```
jingtang10, line 98: is this map still used?

datacapture/src/main/java/com/google/android/fhir/datacapture/mapping/ResourceMapper.kt
```diff
@@ -66,6 +73,55 @@ object ResourceMapper {
...
+    } else {
+      question.fetchExpression?.let { exp ->
+        expressionMap[question.linkId] = exp.expression
+        val answerExtracted = fhirPathEngine.evaluate(resource, exp.expression)[0] as Type
+        question.apply {
+          initial =
+            mutableListOf(
+              Questionnaire.QuestionnaireItemInitialComponent().setValue(answerExtracted)
+            )
+        }
```
jingtang10, line 113: why do you need to use `apply`?

jingtang10 (changes requested):

datacapture/src/main/java/com/google/android/fhir/datacapture/mapping/ResourceMapper.kt
MuhammadSalman-7214: To set the initial value to the questionnaire

datacapture/src/main/java/com/google/android/fhir/datacapture/mapping/ResourceMapper.kt
MuhammadSalman-7214: Done

datacapture/src/main/java/com/google/android/fhir/datacapture/mapping/ResourceMapper.kt
MuhammadSalman-7214: No, and I have removed it as well

datacapture/src/main/java/com/google/android/fhir/datacapture/MoreQuestionnaireItemComponents.kt
MuhammadSalman-7214: Done

MuhammadSalman-7214:
https://github.com/google/android-fhir/pull/536#discussion_r659876257
@jingtang10 done

datacapture/src/main/java/com/google/android/fhir/datacapture/mapping/ResourceMapper.kt
```diff
@@ -66,6 +73,53 @@ object ResourceMapper {
     }
   }
 
+  /**
+   * Returns a `QuestionnaireResponse` to the [questionnaire] that is pre-filled from the [resource]
+   * . See http://build.fhir.org/ig/HL7/sdc/populate.html#expression-based-population.
```
jingtang10, line 78: nit: don't put the period on the new line.

datacapture/src/main/java/com/google/android/fhir/datacapture/mapping/ResourceMapper.kt
jingtang10: why not

`question.inital = ...`

rather than using `.apply{ ... }`?

is that not simpler and more straightforward? is using `apply` adding anything?

datacapture/src/main/java/com/google/android/fhir/datacapture/MoreQuestionnaireItemComponents.kt
jingtang10: where are the unit tests?

jingtang10 (changes requested):

MuhammadSalman-7214:
https://github.com/google/android-fhir/pull/536#discussion_r660424257
@jingtang10 the unit test for createQuestionnaireResponse() is in the ResoueceMapperTest class, should I create a new file for unit test of this function?

datacapture/src/main/java/com/google/android/fhir/datacapture/MoreQuestionnaireItemComponents.kt
MuhammadSalman-7214: @jingtang10 the unit test for createQuestionnaireResponse() is in the ResoueceMapperTest class, should I create a new file for unit test of this function?

jingtang10:
> [#536 (comment)](https://github.com/google/android-fhir/pull/536#discussion_r660424257)
> @jingtang10 the unit test for createQuestionnaireResponse() is in the ResoueceMapperTest class, should I create a new file for unit test of this function?

not that i can see. can you send a link?

datacapture/src/test/java/com/google/android/fhir/datacapture/mapping/ResourceMapperTest.kt
```diff
@@ -720,4 +716,249 @@ class ResourceMapperTest {
...
+    val iParser: IParser = FhirContext.forR4().newJsonParser()
+
+    val uriTestQuestionnaire =
+      iParser.parseResource(org.hl7.fhir.r4.model.Questionnaire::class.java, questionnaireJson) as
+        org.hl7.fhir.r4.model.Questionnaire
+
+    val patient = createPatientResource()
+    val response: QuestionnaireResponse = ResourceMapper.populate(uriTestQuestionnaire, patient)
+
+    assertThat(((response.item[0].item[0].item[0].answer[0]).value as StringType).valueAsString)
```
jingtang10, line 924: `response.item[0]` is repeated so many times - extract a local variable

jingtang10 (approved):

datacapture/src/test/java/com/google/android/fhir/datacapture/mapping/ResourceMapperTest.kt
MuhammadSalman-7214: sure

PR #546, opened 2021-06-12
hide nested elements after parent item deselected / parent answer removed

### LINKED ISSUE #530: Nested items are still revealed after deselecting / removing the answer for the parent item
**Describe the bug**
The SDC has a feature which hides nested items until the parent item has been answered. On removing the answer/ deselecting the option , the nested items are still visible and do not go back to the hidden state.

**Component**
 SDC library

**To Reproduce**
1. Go to datacapturegallery app. 
2. Click on Neonate record from New South Wales, Australia
3. Scroll down to Hep B give y/n
4. Press the checkbox - Date given will appear
5. Uncheck the check box - Date given remains

**Expected behavior**
Date given is supposed to go back to hidden state

**Screenshots**
Initial
answerSelected
answerUnselected

 - Device: One plus 7T pro
 - Android version: 10
 - Build number: QKQ1.200909.002

Originally reported by @f-odhiambo

maanuanubhav999:

Fixes #530

**Description**
Removes nested / child items when parent item deselected/ answer removed

**Alternative(s) considered**
No

**Type**
Choose one: Bug fix

**Checklist**


joiskash (changes requested):
This does not seem like the right fix. In the case of nested questionnaire items, if the parent items answer is empty, null, or false you need to remove all the nested items that are within the parent item,answers.items. The right place to do this is probably in the `questionnaireResponseItemChangedCallback`

maanuanubhav999:
okay sure @joiskash. I will work on that.

maanuanubhav999:
@joiskash PR is failing tests where the parent is false but still, we have child Items. will we actually have a case like this?? Is there any other approach that I could follow or should I change tests?.

joiskash:
> 
> 
> @joiskash PR is failing tests where the parent is false but still, we have child Items. will we actually have a case like this?? Is there any other approach that I could follow or should I change tests?.

You can change any test which does not seem logical with the new change.

datacapture/src/main/java/com/google/android/fhir/datacapture/QuestionnaireViewModel.kt
```diff
@@ -148,7 +154,10 @@ internal class QuestionnaireViewModel(state: SavedStateHandle) : ViewModel() {
       if (enabled) {
         questionnaireItemViewItemList.add(
           QuestionnaireItemViewItem(questionnaireItem, questionnaireResponseItem) {
-            questionnaireResponseItemChangedCallback(questionnaireItem.linkId)
+            questionnaireResponseItemChangedCallback(
+              questionnaireItem.linkId,
+              questionnaireResponseItem
```
joiskash, line 159: Revert this change, you can get the `questionnaireResponseItem` by using the linkId and  `linkIdToQuestionnaireResponseItemMap`

datacapture/src/main/java/com/google/android/fhir/datacapture/QuestionnaireViewModel.kt
```diff
@@ -73,9 +72,16 @@ internal class QuestionnaireViewModel(state: SavedStateHandle) : ViewModel() {
   private val modificationCount = MutableStateFlow(0)
 
   /** Callback function to update the UI. */
-  private val questionnaireResponseItemChangedCallback: (String) -> Unit = { linkId ->
+  private val questionnaireResponseItemChangedCallback:
+    (String, QuestionnaireResponse.QuestionnaireResponseItemComponent) -> Unit =
+      { linkId, questionnaireResponseItem ->
```
joiskash, line 77: Same as below, use the `linkIdToQuestionnaireResponseItemMap`

datacapture/src/main/java/com/google/android/fhir/datacapture/QuestionnaireViewModel.kt
```diff
@@ -73,9 +72,16 @@ internal class QuestionnaireViewModel(state: SavedStateHandle) : ViewModel() {
...
     linkIdToQuestionnaireItemMap[linkId]?.let {
       if (it.hasNestedItemsWithinAnswers) {
+        if (questionnaireResponseItem.answer.first().hasValueBooleanType() &&
```
joiskash, line 80: We also need to check another condition here.. OR `questionnaireResponseItem.answer.first().value.isEmpty`

datacapture/src/main/java/com/google/android/fhir/datacapture/QuestionnaireViewModel.kt
```diff
@@ -73,9 +72,16 @@ internal class QuestionnaireViewModel(state: SavedStateHandle) : ViewModel() {
...
+            !questionnaireResponseItem.answer.first().valueBooleanType.booleanValue()
+        ) {
+          questionnaireResponseItem.answer.removeFirstOrNull()
+        }
         linkIdToQuestionnaireResponseItemMap[linkId]!!.addNestedItemsToAnswer(it)
```
joiskash, line 85: this should be in the else branch.. so if there is no answer value then there should be no nested items in the answer. Makes sense?

joiskash (changes requested):
Please check my review comments and please Re-Request once you have made changes, thanks!

datacapture/src/main/java/com/google/android/fhir/datacapture/QuestionnaireViewModel.kt
maanuanubhav999: Thanks, I totally missed that.

datacapture/src/main/java/com/google/android/fhir/datacapture/QuestionnaireViewModel.kt
maanuanubhav999: Thanks, @joiskash but not quite sure about this or maybe I have misinterpreted. We want to **add Nested Items To Answer** when the condition **value is not empty** and **value is booleanType and its value is true** ?? If this is the case it will not add any nested values when it is not of BooleanType.

datacapture/src/main/java/com/google/android/fhir/datacapture/QuestionnaireViewModel.kt
```diff
@@ -72,17 +72,21 @@ internal class QuestionnaireViewModel(state: SavedStateHandle) : ViewModel() {
...
   /** Callback function to update the UI. */
-  private val questionnaireResponseItemChangedCallback:
-    (String, QuestionnaireResponse.QuestionnaireResponseItemComponent) -> Unit =
-      { linkId, questionnaireResponseItem ->
+  private val questionnaireResponseItemChangedCallback: (String) -> Unit = { linkId ->
     linkIdToQuestionnaireItemMap[linkId]?.let {
       if (it.hasNestedItemsWithinAnswers) {
-        if (questionnaireResponseItem.answer.first().hasValueBooleanType() &&
-            !questionnaireResponseItem.answer.first().valueBooleanType.booleanValue()
+        if (linkIdToQuestionnaireResponseItemMap[linkId]!!.answer.first().value.isEmpty ||
```
joiskash, line 78: i think this should be `condition_1 || (condition2 && condition3)`

joiskash (changes requested):
Please add a test case to check the new condition that you have written

datacapture/src/main/java/com/google/android/fhir/datacapture/QuestionnaireViewModel.kt
maanuanubhav999: Okay. Sure

maanuanubhav999: I have to push multiple PR as my local `./gradlew check`  is failing all tests. I think there is already an issue related to it. I will later squash them.

datacapture/src/test/java/com/google/android/fhir/datacapture/QuestionnaireViewModelTest.kt
```diff
@@ -740,17 +740,72 @@ class QuestionnaireViewModelTest {
...
+      }
+    )
+    viewModel.questionnaireItemViewItemList[0].questionnaireResponseItemChangedCallback()
+    viewModel.questionnaireItemViewItemList[0].questionnaireResponseItem.answer[0].item[0]
+      .addAnswer(
+        QuestionnaireResponse.QuestionnaireResponseItemAnswerComponent().apply {
+          this.value = valueBooleanType.setValue(true)
+        }
+      )
+    viewModel.questionnaireItemViewItemList[0].questionnaireResponseItem.answer.clear()
```
joiskash, line 802: I would suggest removing this line and instead putting `viewModel.questionnaireItemViewItemList[0].questionnaireResponseItem.answer[0].value = false` so that the nested item is removed and you can test it.. Over here it looks like you are just clearing the answer and then setting the value to false. Do let me know if my understanding is wrong.

joiskash (changes requested):
Put another test case for type string. But otherwise I think this PR is almost ready! Thanks!

datacapture/src/main/java/com/google/android/fhir/datacapture/QuestionnaireViewModel.kt
joiskash: Sorry missed this, so when the questionnaireItem has a nested item then the corresponding questionnaireResponseItem will not have a nested item, the questionnaireResponseItem.answer will have the nested item. It is the structure that the FHIR spec has defined. It is to make sure that if the parent questionnaireItem allows repeated answers, then all the answers will contain the nested items.

joiskash: Please have a look at this [discussion](https://chat.fhir.org/#narrow/stream/179255-questionnaire/topic/Questionnaire.20Response.20example)

datacapture/src/test/java/com/google/android/fhir/datacapture/QuestionnaireViewModelTest.kt
maanuanubhav999: sure thanks. I totally forgot that clear will erase the nested items also.

datacapture/src/main/java/com/google/android/fhir/datacapture/QuestionnaireViewModel.kt
maanuanubhav999: @joiskash thanks, do we need this(moving to else branch) even now?

joiskash (commented):
Look good to me :) , @jingtang10 please have a look if you feel any changes are needed. After that @maanuanubhav999 you can merge this to master. Thanks!

joiskash (approved):

datacapture/src/main/java/com/google/android/fhir/datacapture/QuestionnaireViewModel.kt
joiskash: yes it looks fine now. Just a small comment i missed in the review, `linkIdToQuestionnaireResponseItemMap[linkId]!!.answer.first()` could probably be a local variable to make the if-else more readable.

jingtang10 (commented):
We don't normally prefix PRs with things like `Fix:`. IMHO they actually make the PRs harder to read. We can use labels instead if needed.

Just do `Hide nested ...`

Also I'm not so sure that the fix is in this file... look at this block https://github.com/google/android-fhir/blob/f5116ed7aef1f93bedd840e8fde1046f65252350/datacapture/src/main/java/com/google/android/fhir/datacapture/views/QuestionnaireItemCheckBoxViewHolderFactory.kt#L40: 

when the check box is unticked, the answer is still there but with a false value. I think in this case having a nested item is kind of legitimate. I think if we want to fix this issue, we could make sure that `false` answers are simply clearing the answers (see below example) -- but is that what we want?

```
questionnaireItemViewItem.singleAnswerOrNull =
  if (checkBox.isChecked) {
    QuestionnaireResponse.QuestionnaireResponseItemAnswerComponent().apply {
      value = BooleanType(checkBox.isChecked)
    }
  } else {
    null
}
```

@joiskash - do you think we can ask on fhir chat about this?

joiskash:
> 
> 
> We don't normally prefix PRs with things like `Fix:`. IMHO they actually make the PRs harder to read. We can use labels instead if needed.
> 
> Just do `Hide nested ...`
> 
> Also I'm not so sure that the fix is in this file... look at this block
> 
> https://github.com/google/android-fhir/blob/f5116ed7aef1f93bedd840e8fde1046f65252350/datacapture/src/main/java/com/google/android/fhir/datacapture/views/QuestionnaireItemCheckBoxViewHolderFactory.kt#L40
> :
> 
> when the check box is unticked, the answer is still there but with a false value. I think in this case having a nested item is kind of legitimate. I think if we want to fix this issue, we could make sure that `false` answers are simply clearing the answers (see below example) -- but is that what we want?
> 
> ```
> questionnaireItemViewItem.singleAnswerOrNull =
>   if (checkBox.isChecked) {
>     QuestionnaireResponse.QuestionnaireResponseItemAnswerComponent().apply {
>       value = BooleanType(checkBox.isChecked)
>     }
>   } else {
>     null
> }
> ```
> 
> @joiskash - do you think we can ask on fhir chat about this?

Makes sense..I was actually looking at this only from the point of UI. There can be a question which if the answer is false, the nested items still make sense. I did not think this through. 

Current scenario is nested items remain hidden until the parent item has been answered.

Once the parent item is answered the nested items appear.

What should happen now when the parent answer is cleared (in case of edit text or integer) or set to false in case of a boolean.

@f-odhiambo please put your thoughts here.

jingtang10:
One thing we _could_ do is just to say, if it's a single checkbox, then we will clear the answer if it's unticked. If you want the answer to be set to false but present, maybe use a radio button group (with yes and no radio buttons).

maanuanubhav999:
> One thing we _could_ do is just to say, if it's a single checkbox, then we will clear the answer if it's unticked. If you want the answer to be set to false but remain present, maybe use a radio button group (with yes and no radio buttons).

sorry I messed up earlier. Summarizing from the points mentioned above 

We surely don't want to remove the nested element **every time** the parent is `false`. it will be required in many cases ( i.e. question:  did you take your medicine, if not why?). So when to remove it ??

But we may still require to clear nested elements when `(case of edit text or integer)`

Maybe we can have a **clear option** that would clear all the filled QuestionnaireResponse??  there may be other cases when one requires to clear the form.

joiskash:
> One thing we _could_ do is just to say, if it's a single checkbox, then we will clear the answer if it's unticked. If you want the answer to be set to false but remain present, maybe use a radio button group (with yes and no radio buttons).

This seems like a implementation detail that can be missed out.. @f-odhiambo  please comment. This issue was raised by you, what is the ideal behavior according to you?

f-odhiambo:
HI @joiskash. Sorry I may have missed this. 

So the ideal scenario for this would be

1. if the parent items answer is empty, null, or false you need to remove all the nested items that are within the parent item
2.  if the parent items answer is answered, checked, or true you need to show/make visible  all the nested items that are within the parent item

Thus 
1. Go to datacapturegallery app.
2. Click on Neonate record from New South Wales, Australia
3. Scroll down to Hep B give y/n
4. Press the checkbox - date field should be made visible 
5. Uncheck the check box - date field should be hidden

joiskash:
> 
> 
> HI @joiskash. Sorry I may have missed this.
> 
> So the ideal scenario for this would be
> 
>     1. if the parent items answer is empty, null, or false you need to remove all the nested items that are within the parent item
> 
>     2. if the parent items answer is answered, checked, or true you need to show/make visible  all the nested items that are within the parent item
> 
> 
> Thus
> 
>     1. Go to datacapturegallery app.
> 
>     2. Click on Neonate record from New South Wales, Australia
> 
>     3. Scroll down to Hep B give y/n
> 
>     4. Press the checkbox - date field should be made visible
> 
>     5. Uncheck the check box - date field should be hidden

So the main issue is with false, there may be a parent question which can have the answer as false and have valid nested items.. So in the current build if the parent answer is by default unchecked then the nested items will not be visible. And after you check and uncheck if we clear nested items then for a false answer we can never have the nested items..

Jing did suggest asking this to the fhir chat group, but I have not done it yet.

jingtang10:
https://chat.fhir.org/#narrow/stream/179255-questionnaire/topic/EnableWhen.20behavior.20with.20operator.20.60exists.60

let's see the response.

jingtang10:
please see the chat link above. one solution suggested by people on the fhir chat is that we need to be able to clear boolean questions (i.e. adding another option to the radio button group or use a different widget type) and indeed coding questions.

joiskash:
As per my understanding, from the FHIR chat we can conclude that 

1.  Boolean questions should be rendered as 3 options. The 3rd one being "not answered" or something similar
2. Choice questions should have a clear answer option.

maanuanubhav999:
> As per my understanding, from the FHIR chat we can conclude that
> 
> 1. Boolean questions should be rendered as 3 options. The 3rd one being "not answered" or something similar
> 2. Choice questions should have a clear answer option.
> 
> @jingtang10 please correct me if I am wrong

@joiskash So should I move forward with this?

jingtang10:
> > As per my understanding, from the FHIR chat we can conclude that
> > 
> > 1. Boolean questions should be rendered as 3 options. The 3rd one being "not answered" or something similar
> > 2. Choice questions should have a clear answer option.
> > 
> > @jingtang10 please correct me if I am wrong
> 
> @joiskash So should I move forward with this?

That's be great thanks!

jingtang10:
@maanuanubhav999 is there more to do here? if not can you close this?

PR #586, opened 2021-06-24
Horizontal layout support for RadioGroup and CheckBoxGroup

FikriMilano:
**IMPORTANT: All PRs must be linked to an issue (except for extremely trivial and straightforward changes).**

Fixes #542

**Description**
Add horizontal layout support for RadioGroup and CheckBoxGroup using `choiceOrientation` component.
To render each item in a single direction either vertical or horizontal, we use Constraint Layout Flow.

**Alternative(s) considered**
N/A

**Type**
Choose one: Feature

**Checklist**
- [x] I have read and acknowledged the [Code of conduct](https://github.com/google/android-fhir/blob/master/docs/CODE_OF_CONDUCT.md)
- [x] I have read [How to Contribute](https://github.com/google/android-fhir/blob/master/docs/contributing.md)
- [x] I have read the [Developer's guide](https://github.com/google/android-fhir/wiki/Developer's-Guide)
- [x] I have signed the Google [Individual CLA](https://cla.developers.google.com/about/google-individual), or I am covered by my company's [Corporate CLA](https://cla.developers.google.com/about/google-corporate )
- [x] I have discussed my proposed solution with code owners in the linked issue(s) and we have agreed upon the general approach
- [x] I have run `./gradlew spotlessApply` and `./gradlew spotlessCheck` to check my code follows the style guide of this project
- [x] I have run `./gradlew check` and `./gradlew connectedCheck` to test my changes locally
- [x] I have built and run the reference app(s) to verify my change fixes the issue and/or does not break the reference app(s)

PR #692, opened 2021-08-09
Support Date Approximate

### LINKED ISSUE #568: Support the Approximate prefix while searching for dates

Support Approximate prefix while searching for dates.

 ref https://www.hl7.org/fhir/search.html#prefix.

> the value for the parameter in the resource is approximately the same to the provided value.
Note that the recommended value for the approximation is 10% of the stated value (or for a date, 10% of the gap between now and the date), but systems may choose other values where appropriate

**Additional context**

We need to discuss how to implement this, and what approximate would imply. We could stick to what the spec recommends (stated above) or we could come up with our own definition. 

Maybe adter we decide what the strategy.

epicadk:
- Support date approximate
- add some comments

Fixes #568 

**Description**

**Alternative(s) considered**

**Type**
Choose one: Feature

**Checklist**


Tarun-Bhardwaj:
@epicadk , are you sure this PR fixes #586 (as mentioned in the description above)? Should it not be #568 ?

epicadk:
My bad, I was using the CLI and messed up the numbering.

epicadk:
@Tarun-Bhardwaj done with the changes. awating review

buildSrc/src/main/kotlin/Dependencies.kt
```diff
@@ -102,6 +102,8 @@ object Dependencies {
   const val robolectric = "org.robolectric:robolectric:${Versions.robolectric}"
   const val truth = "com.google.truth:truth:${Versions.truth}"
   const val flexBox = "com.google.android.flexbox:flexbox:${Versions.flexBox}"
+  const val mockk = "io.mockk:mockk:${Versions.mockk}"
+  const val mockkAndroid = "io.mockk:mockk-android:${Versions.mockk}"
```
jingtang10, line 106: the reason that mocking is needed here is that the `getCurrentDate` function is written as a global function. Can you add something like a "date provider" that the engine/search class depends on so you can easily use a custom implementation of that provider in tests?

@stevenckngaa @yigit  @kevinmost do you agree with this approach?

engine/src/androidTest/java/com/google/android/fhir/db/impl/DatabaseImplTest.kt
```diff
@@ -967,6 +974,52 @@ class DatabaseImplTest {
     assertThat(result).isEmpty()
   }
 
+  @Test
+  fun search_dateTime_Approximate() = runBlocking {
```
jingtang10, line 978:
```suggestion
  fun search_dateTime_approximate() = runBlocking {
```

engine/src/androidTest/java/com/google/android/fhir/db/impl/DatabaseImplTest.kt
```diff
@@ -967,6 +974,52 @@ class DatabaseImplTest {
...
+          Search(ResourceType.Patient)
+            .apply {
+              filter(Patient.BIRTHDATE, DateType("2013-03-14"), ParamPrefixEnum.APPROXIMATE)
+            }
+            .getQuery()
+        )
+      assertThat(result.single().id).isEqualTo("Patient/1")
+    }
+  }
+
```
jingtang10, line 1022: can you add two negative cases (where no result is returned) as well?

engine/src/main/java/com/google/android/fhir/search/MoreSearch.kt
```diff
@@ -216,7 +219,25 @@ private fun getConditionParamPair(prefix: ParamPrefixEnum, value: DateType): Con
   val start = value.rangeEpochDays.first
   val end = value.rangeEpochDays.last
   return when (prefix) {
-    ParamPrefixEnum.APPROXIMATE -> TODO("Not Implemented")
+    ParamPrefixEnum.APPROXIMATE -> {
+      val currentDateTime = DateType(getCurrentDate(), value.precision)
+      val diffStart =
+        (value.rangeEpochDays.first -
```
jingtang10, line 225: `value.rangeEpochDays.first` is already defined as `start`
and `value.rangeEpochDays.last` as `end`

same with below

engine/src/main/java/com/google/android/fhir/search/MoreSearch.kt
```diff
@@ -216,7 +219,25 @@ private fun getConditionParamPair(prefix: ParamPrefixEnum, value: DateType): Con
...
+            0.1 * (value.rangeEpochDays.first - currentDateTime.rangeEpochDays.first).absoluteValue)
```
jingtang10, line 226: define a APPROXIMATION_COEFFICIENT as 0.1 and add a comment that explains what it is and links to the spec.

same with below

jingtang10 (changes requested):

epicadk:
Sort of busy with the presentation will work on this after.

engine/src/main/java/com/google/android/fhir/MoreDates.kt
```diff
@@ -16,9 +16,13 @@
...
+import java.time.Instant
 import java.time.LocalDate
 import java.util.Date
 
 @Suppress("DEPRECATION") // java.util.Date API used by HAPI
 internal val Date.epochDay
   get() = LocalDate.of(year + 1900, month + 1, date).toEpochDay()
+
+/** returns the current date */
+internal fun getCurrentDate() = Date.from(Instant.now())
```
jingtang10, line 28: remove? this is not used any more i think

engine/src/main/java/com/google/android/fhir/search/MoreSearch.kt
```diff
@@ -26,11 +26,20 @@ import com.google.android.fhir.db.Database
...
+import java.util.Date
+import kotlin.math.absoluteValue
+import kotlin.math.roundToLong
 import org.hl7.fhir.r4.model.DateTimeType
 import org.hl7.fhir.r4.model.DateType
 import org.hl7.fhir.r4.model.Resource
 import org.hl7.fhir.r4.model.ResourceType
 
+/**
+ * multiplier required to Approximate to a 10% range see https://www.hl7.org/fhir/search.html#prefix
```
jingtang10, line 39:
```suggestion
 * The multiplier used to determine the range for the `ap` search prefix. See https://www.hl7.org/fhir/search.html#prefix for more details.
```

engine/src/main/java/com/google/android/fhir/search/MoreSearch.kt
```diff
@@ -212,11 +221,34 @@ val Order?.sqlString: String
...
   val start = value.rangeEpochDays.first
   val end = value.rangeEpochDays.last
   return when (prefix) {
-    ParamPrefixEnum.APPROXIMATE -> TODO("Not Implemented")
+    ParamPrefixEnum.APPROXIMATE -> {
+      val currentDateTime = DateType(Date.from(currentDate), value.precision)
+      val diffStart =
+        (start -
+            APPROXIMATION_COEFFICIENT *
+              (start - currentDateTime.rangeEpochDays.first).absoluteValue)
```
jingtang10, line 237: make this a private util function

engine/src/main/java/com/google/android/fhir/search/SearchDsl.kt
```diff
@@ -36,7 +38,12 @@ import org.hl7.fhir.r4.model.ResourceType
 import org.hl7.fhir.r4.model.UriType
 
 @SearchDslMarker
-data class Search(val type: ResourceType, var count: Int? = null, var from: Int? = null) {
+data class Search(
+  val type: ResourceType,
+  var count: Int? = null,
+  var from: Int? = null,
+  val dateProvider: Clock = Clock.systemDefaultZone()
```
jingtang10, line 45: @stevenckngaa @kevinmost @yigit do you think this should be part of the fhir engine rather than the search interface?

jingtang10 (changes requested):

engine/src/main/java/com/google/android/fhir/search/SearchDsl.kt
epicadk: I've added it to the fhir engine for now. I'll rebase if just having it in the search api is fine.

jingtang10 (approved):

PR #749, opened 2021-08-22
Single fhir engine per process

### LINKED ISSUE #100: Making Fhir engine a singleton per process per db name
Following discussions between @yigit @florina-muntenescu @jingtang10, we decided to make fhir engine a singleton per process + db name to avoid racing conditions when updating the db

- [x] Modify FhirEngine's API so that it's impossible to create another instance if there is already one (i.e. caching). At the moment in the referende app this is ensured by the initialization code in `FhirApplication.kt` using `by lazy`. But we want to be in the engine itself
- [x] Mark the `inMemory()` API as internal since we're only using it in tests and definitly don't want users to actually create in-memory db
- [x] Simplify the initialization code and remove the db name option (it's not used anywhere anyway)

deepankarb:
Fixes #100

**Description**
* Made the public API `FhirEngineBuilder` a singleton and exposed only a single instance of `FhirEngine` from it.
* Added tests to check the same instance is returned when `build(Context)` is called more than once with application and activity contexts.
* Made `inMemory()` internal since it's only for testing
* Removed option to configure DB name - it will always be `fhirEngine`

**Type**
Choose one: Improvement

**Checklist**
- [x ] I have read and acknowledged the [Code of conduct](https://github.com/google/android-fhir/blob/master/docs/CODE_OF_CONDUCT.md)
- [ x] I have read [How to Contribute](https://github.com/google/android-fhir/blob/master/docs/contributing.md)
- [ x] I have read the [Developer's guide](https://github.com/google/android-fhir/wiki/Developer's-Guide)
- [ x] I have signed the Google [Individual CLA](https://cla.developers.google.com/about/google-individual), or I am covered by my company's [Corporate CLA](https://cla.developers.google.com/about/google-corporate )
- [x ] I have discussed my proposed solution with code owners in the linked issue(s) and we have agreed upon the general approach
- [ x] I have run `./gradlew spotlessApply` and `./gradlew spotlessCheck` to check my code follows the style guide of this project
- [ x] I have run `./gradlew check` and `./gradlew connectedCheck` to test my changes locally
- [x ] I have built and run the reference app(s) to verify my change fixes the issue and/or does not break the reference app(s)

### DIFF
--- a/engine/src/main/java/com/google/android/fhir/FhirEngineBuilder.kt
+++ b/engine/src/main/java/com/google/android/fhir/FhirEngineBuilder.kt
@@ -19,9 +19,14 @@ package com.google.android.fhir
 import android.content.Context
 
 /** The builder for [FhirEngine] instance */
-class FhirEngineBuilder constructor(context: Context) {
-  private val services = FhirServices.builder(context)
+object FhirEngineBuilder {
+  lateinit var fhirEngine: FhirEngine
 
   /** Builds a new instance of the [FhirEngine]. */
-  fun build() = services.build().fhirEngine
+  fun build(context: Context): FhirEngine {
+    if (!::fhirEngine.isInitialized) {
+      fhirEngine = FhirServices.builder(context.applicationContext).build().fhirEngine
+    }
+    return fhirEngine
+  }
 }
--- a/engine/src/test/java/com/google/android/fhir/FhirEngineProviderTest.kt
+++ b/engine/src/test/java/com/google/android/fhir/FhirEngineProviderTest.kt
@@ -23,22 +23,23 @@ import org.junit.Test
 import org.junit.runner.RunWith
 import org.robolectric.RobolectricTestRunner
 
-/** Unit tests for [FhirEngineBuilder]. */
+/** Unit tests for [FhirEngineProvider]. */
 @RunWith(RobolectricTestRunner::class)
-class FhirEngineBuilderTest {
-  private val builder: FhirEngineBuilder = FhirEngineBuilder
+class FhirEngineProviderTest {
+  private val provider: FhirEngineProvider = FhirEngineProvider
 
   @Test
   fun build_twiceWithAppContext_shouldReturnSameFhirEngine() {
-    val engineOne = builder.build(ApplicationProvider.getApplicationContext())
-    val engineTwo = builder.build(ApplicationProvider.getApplicationContext())
+    val engineOne = provider.getInstance(ApplicationProvider.getApplicationContext())
+    val engineTwo = provider.getInstance(ApplicationProvider.getApplicationContext())
     Truth.assertThat(engineOne).isSameInstanceAs(engineTwo)
   }
 
   @Test
   fun build_withAppAndActivityContext_shouldReturnSameFhirEngine() {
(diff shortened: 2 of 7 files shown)


deepankarb:
@jingtang10 can you review this? Thanks!

engine/src/main/java/com/google/android/fhir/FhirEngineBuilder.kt
```diff
@@ -19,9 +19,14 @@ package com.google.android.fhir
...
-  private val services = FhirServices.builder(context)
+object FhirEngineBuilder {
+  lateinit var fhirEngine: FhirEngine
 
   /** Builds a new instance of the [FhirEngine]. */
-  fun build() = services.build().fhirEngine
+  fun build(context: Context): FhirEngine {
+    if (!::fhirEngine.isInitialized) {
+      fhirEngine = FhirServices.builder(context.applicationContext).build().fhirEngine
+    }
```
aditya-07, line 29: Is this block thread-safe? I know that lazy delegation is thread-safe, not sure about this.

aditya-07 (commented):
Added a comment regarding the thread-safety, will like to know your thoughts on it. Apart from that LGTM.

engine/src/main/java/com/google/android/fhir/FhirEngineBuilder.kt
deepankarb: I don't think this is thread safe hence have Synchronised it.

engine/src/test/java/com/google/android/fhir/FhirEngineProviderTest.kt
```diff
@@ -0,0 +1,45 @@
...
+/** Unit tests for [FhirEngineProvider]. */
+@RunWith(RobolectricTestRunner::class)
+class FhirEngineProviderTest {
+  private val provider: FhirEngineProvider = FhirEngineProvider
+
+  @Test
+  fun build_twiceWithAppContext_shouldReturnSameFhirEngine() {
+    val engineOne = provider.getInstance(ApplicationProvider.getApplicationContext())
+    val engineTwo = provider.getInstance(ApplicationProvider.getApplicationContext())
+    Truth.assertThat(engineOne).isSameInstanceAs(engineTwo)
```
kevinmost, line 35: Please import `Truth.assertThat` directly:

```kt
import com.google.common.truth.Truth.assertThat
```

```kt
assertThat(engineOne).isSameInstanceAs(engineTwo)
```

engine/src/test/java/com/google/android/fhir/FhirEngineProviderTest.kt
```diff
@@ -0,0 +1,45 @@
...
+  }
+
+  @Test
+  fun build_withAppAndActivityContext_shouldReturnSameFhirEngine() {
+    val engineAppContext = provider.getInstance(ApplicationProvider.getApplicationContext())
+    val engineActivityContext =
+      provider.getInstance(InstrumentationRegistry.getInstrumentation().context)
+    Truth.assertThat(engineAppContext).isSameInstanceAs(engineActivityContext)
```
kevinmost, line 43: Ditto

kevinmost (approved):

PR #756, opened 2021-08-26
Add response item initial value to questionnaire response only if enablewhen is true.

### LINKED ISSUE #755: Though enable when is false QuestionResponseItem with initial value present in questionnaire response at the time of data extraction.
**Describe the bug**
Found a new edge case which is not supported yet by sdk.
If the question item has enablewhen clause, and it also has initial coding value, in that case though enablewhen clause is false question response item with initial coding value present in questionnaire response at the time of data extraction.

Just consider the symptoms question as an example.
Symptoms main question has 4 answer options as checkboxes.
None of the answer options are selected initially.
There are hidden question items with initial coding value and enablewhen clause is false.
Bug = > though enable when is false, question response item with initial coding value present in question response at the time of data extraction.

**Component**
SDC library

**To Reproduce**

**Expected behavior**

**Screenshots**

**Additional context**

santosh-pingle:

Fixes #755 

**Description**
Add response item initial value to questionnaire response only if enablewhen is true.
Remove response item initial value from questionnaire response if enablewhen is false.

**Alternative(s) considered**

**Type**

**Checklist**

### DIFF
--- a/datacapture/src/main/java/com/google/android/fhir/datacapture/QuestionnaireViewModel.kt
+++ b/datacapture/src/main/java/com/google/android/fhir/datacapture/QuestionnaireViewModel.kt
@@ -135,7 +135,7 @@ internal class QuestionnaireViewModel(state: SavedStateHandle) : ViewModel() {
    */
   fun getQuestionnaireResponse(): QuestionnaireResponse {
     return questionnaireResponse.copy().apply {
-      item = getEnableResponseItems(this@QuestionnaireViewModel.questionnaire.item, item)
+      item = getEnabledResponseItems(this@QuestionnaireViewModel.questionnaire.item, item)
     }
   }
 
@@ -234,7 +234,7 @@ internal class QuestionnaireViewModel(state: SavedStateHandle) : ViewModel() {
     return QuestionnaireState(items = items, pagination = pagination)
   }
 
-  private fun getEnableResponseItems(
+  private fun getEnabledResponseItems(
     questionnaireItemList: List<Questionnaire.QuestionnaireItemComponent>,
     questionnaireResponseItemList: List<QuestionnaireResponse.QuestionnaireResponseItemComponent>,
   ): List<QuestionnaireResponse.QuestionnaireResponseItemComponent> {
@@ -243,24 +243,24 @@ internal class QuestionnaireViewModel(state: SavedStateHandle) : ViewModel() {
       .zip(questionnaireResponseItemList.asSequence())
       .filter { (questionnaireItem, questionnaireResponseItem) ->
         EnablementEvaluator.evaluate(questionnaireItem) { linkId ->
-          val questionnaireItemComponent = linkIdToQuestionnaireItemMap[linkId]
-          val questionnaireResponseItemComponent = linkIdToQuestionnaireResponseItemMap[linkId]
-          if (questionnaireItemComponent == null || questionnaireResponseItemComponent == null) {
+          val questionnaireItem = linkIdToQuestionnaireItemMap[linkId]
+          val questionnaireResponseItem = linkIdToQuestionnaireResponseItemMap[linkId]
+          if (questionnaireItem == null || questionnaireResponseItem == null) {
             return@evaluate QuestionnaireItemWithResponse(null, null)
           }
           QuestionnaireItemWithResponse(
-            questionnaireItem = questionnaireItemComponent,
-            questionnaireResponseItem = questionnaireResponseItemComponent
+            questionnaireItem = questionnaireItem,
+            questionnaireResponseItem = questionnaireResponseItem
           )
         }
       }
       .map { (questionnaireItem, questionnaireResponseItem) ->
         // Nested group items
         questionnaireResponseItem.item =
-          getEnableResponseItems(questionnaireItem.item, questionnaireResponseItem.item)
+          getEnabledResponseItems(questionnaireItem.item, questionnaireResponseItem.item)
         // Nested question items
         questionnaireResponseItem.answer.forEach {
-          it.item = getEnableResponseItems(questionnaireItem.item, it.item)
+          it.item = getEnabledResponseItems(questionnaireItem.item, it.item)
         }
         questionnaireResponseItem
       }
--- a/datacapture/src/main/java/com/google/android/fhir/datacapture/QuestionnaireViewModel.kt
+++ b/datacapture/src/main/java/com/google/android/fhir/datacapture/QuestionnaireViewModel.kt
@@ -126,8 +126,21 @@ internal class QuestionnaireViewModel(state: SavedStateHandle) : ViewModel() {
           )
       )
 
-  /** The current [QuestionnaireResponse] captured by the UI. */
-  fun getQuestionnaireResponse(): QuestionnaireResponse = questionnaireResponse
+  /**
+   * Returns current[QuestionnaireResponse] captured by the UI which includes answers of enabled
+   * questions.
+   *
+   * Set of answers will not be always same, the questions previously populated with answers can be
+   * re-enabled or disabled.
+   */
+  fun getQuestionnaireResponse(): QuestionnaireResponse {
+    val questionnaireResponseCopy = questionnaireResponse.copy()
+    removeAllNestedAnswersOfDisabledQuestionnaire(
+      questionnaire.item,
+      questionnaireResponseCopy.item
+    )
+    return questionnaireResponseCopy
+  }
 
   private fun createLinkIdToQuestionnaireResponseItemMap(
     questionnaireResponseItemList: List<QuestionnaireResponse.QuestionnaireResponseItemComponent>
@@ -198,33 +211,59 @@ internal class QuestionnaireViewModel(state: SavedStateHandle) : ViewModel() {
             }
 
           if (!enabled) {
-            questionnaireResponseItem.removeAllNestedAnswers()
-            emptyList()
-          } else {
-            if (questionnaireItem.hasEnableWhen() && !questionnaireResponseItem.hasAnswer()) {
-              questionnaireResponseItem.updateInitialValue(questionnaireItem)
-            }
-            listOf(
-              QuestionnaireItemViewItem(questionnaireItem, questionnaireResponseItem) {
-                questionnaireResponseItemChangedCallback(questionnaireItem.linkId)
-              }
-            ) +
-              getQuestionnaireState(
-                  questionnaireItemList = questionnaireItem.item,
-                  questionnaireResponseItemList =
-                    if (questionnaireResponseItem.answer.isEmpty()) {
-                      questionnaireResponseItem.item
-                    } else {
-                      questionnaireResponseItem.answer.first().item
-                    },
-                  // we're now dealing with nested items, so pagination is no longer a concern
-                  pagination = null,
-                )
-                .items
+            return@flatMap emptyList()
           }
+
+          listOf(
+            QuestionnaireItemViewItem(questionnaireItem, questionnaireResponseItem) {
+              questionnaireResponseItemChangedCallback(questionnaireItem.linkId)
+            }
+          ) +
+            getQuestionnaireState(
+                questionnaireItemList = questionnaireItem.item,
+                questionnaireResponseItemList =
+                  if (questionnaireResponseItem.answer.isEmpty()) {
+                    questionnaireResponseItem.item
+                  } else {
+                    questionnaireResponseItem.answer.first().item
+                  },
+                // we're now dealing with nested items, so pagination is no longer a concern
+                pagination = null,
+              )
+              .items
         }
         .toList()
     return QuestionnaireState(items = items, pagination = pagination)
+  }
+
+  private fun removeAllNestedAnswersOfDisabledQuestionnaire(
+    questionnaireItemList: List<Questionnaire.QuestionnaireItemComponent>,
+    questionnaireResponseItemList: List<QuestionnaireResponse.QuestionnaireResponseItemComponent>,
+  ) {
+    questionnaireItemList.asSequence().zip(questionnaireResponseItemList.asSequence()).forEach {
+      (questionnaireItem, questionnaireResponseItem) ->
+      val enabled =
+        EnablementEvaluator.evaluate(questionnaireItem) { linkId ->
+          QuestionnaireItemWithResponse(
+            questionnaireItem = (linkIdToQuestionnaireItemMap[linkId]
+                ?: return@evaluate QuestionnaireItemWithResponse(null, null)),
+            questionnaireResponseItem = (linkIdToQuestionnaireResponseItemMap[linkId]
+                ?: return@evaluate QuestionnaireItemWithResponse(null, null))
+          )
+        }
+      if (!enabled) {
+        return@forEach questionnaireResponseItem.removeAllNestedAnswers()
+      }
+      removeAllNestedAnswersOfDisabledQuestionnaire(
+        questionnaireItemList = questionnaireItem.item,
+        questionnaireResponseItemList =
+          if (questionnaireResponseItem.answer.isEmpty()) {
+            questionnaireResponseItem.item
+          } else {
+            questionnaireResponseItem.answer.first().item
+          }
+      )
+    }
   }
 }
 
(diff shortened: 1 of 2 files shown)


datacapture/src/main/java/com/google/android/fhir/datacapture/QuestionnaireViewModel.kt
```diff
@@ -216,6 +218,7 @@ internal class QuestionnaireViewModel(state: SavedStateHandle) : ViewModel() {
                 )
                 .items
           } else {
+            questionnaireResponseItem?.removeAnswers()
```
aditya-07, line 221: `questionnaireResponseItem` is not nullable. So we can get rid of the safe `?.` call.

aditya-07, line 221: `questionnaireResponseItem` is not nullable. So we can get rid of the safe `?.` call.

datacapture/src/test/java/com/google/android/fhir/datacapture/QuestionnaireViewModelTest.kt
```diff
@@ -813,6 +813,192 @@ class QuestionnaireViewModelTest {
...
+            )
+            addItem(
+              QuestionnaireResponse.QuestionnaireResponseItemComponent().apply {
+                linkId = "question-2"
+                addAnswer()
+              }
+            )
+          }
+        )
+      }
```
aditya-07, line 923: Instead of generating the `questionnaireResponse` here, can you add the static `questionnaireResponseJson` here so that exact expected value is know before hand. Later for `assertion`, you can either serialize `viewModel.getQuestionnaireResponse()` to a json to perform string comparison or convert `questionnaireResponseJson` to `questionnaireResponse` object.

datacapture/src/test/java/com/google/android/fhir/datacapture/QuestionnaireViewModelTest.kt
```diff
@@ -813,6 +813,192 @@ class QuestionnaireViewModelTest {
...
+                addAnswer(
+                  QuestionnaireResponse.QuestionnaireResponseItemAnswerComponent().apply {
+                    this.value = valueBooleanType.setValue(true)
+                  }
+                )
+              }
+            )
+          }
+        )
+      }
```
aditya-07, line 994: Same as above.

datacapture/src/main/java/com/google/android/fhir/datacapture/QuestionnaireViewModel.kt
```diff
@@ -198,6 +197,9 @@ internal class QuestionnaireViewModel(state: SavedStateHandle) : ViewModel() {
               )
             }
           if (enabled) {
+            if (questionnaireItem.hasEnableWhen() && !questionnaireResponseItem.hasAnswer()) {
+              questionnaireResponseItem.updateInitialValue(questionnaireItem)
+            }
```
aditya-07, line 202: Is this required here?It should have the value from init code 
```
questionnaire.item.forEach {
        questionnaireResponse.addItem(it.createQuestionnaireResponseItem())
      }
```

aditya-07 (changes requested):

datacapture/src/main/java/com/google/android/fhir/datacapture/QuestionnaireViewModel.kt
santosh-pingle: It's issue.
If question item is not enabled then it can not have answer, though it has initial value.
If question item is enabled then it can have answer, if it has initial value then just add those as initial answer.

We need it because of 
if not enabled then those answers are removed ( questionnaireResponseItem.removeAnswers()), 
and if enabled then just add those initial value as initial answers.

aditya-07 (approved):
LGTM.

datacapture/src/main/java/com/google/android/fhir/datacapture/QuestionnaireViewModel.kt
```diff
@@ -324,3 +327,20 @@ internal fun QuestionnairePagination.nextPage(): QuestionnairePagination {
...
+  answer = responseItemWithInitialValue.answer
+}
+
+internal fun QuestionnaireResponse.QuestionnaireResponseItemComponent.removeAnswers() {
+  if (hasItem()) {
+    item.forEach { it.removeAnswers() }
+  }
+  if (hasAnswer()) {
+    answer.removeAll { true }
+  }
```
jingtang10, line 345:
```suggestion
  answer.removeAll { true }
  item.forEach { it.removeAnswers() }
```

datacapture/src/main/java/com/google/android/fhir/datacapture/QuestionnaireViewModel.kt
```diff
@@ -324,3 +327,20 @@ internal fun QuestionnairePagination.nextPage(): QuestionnairePagination {
...
+
+internal fun QuestionnaireResponse.QuestionnaireResponseItemComponent.updateInitialValue(
+  questionnaireItem: Questionnaire.QuestionnaireItemComponent
+) {
+  val responseItemWithInitialValue = questionnaireItem.createQuestionnaireResponseItem()
+  item = responseItemWithInitialValue.item
+  answer = responseItemWithInitialValue.answer
+}
+
+internal fun QuestionnaireResponse.QuestionnaireResponseItemComponent.removeAnswers() {
```
jingtang10, line 339: this should be called `removeAllAnswers` or `removeAllNestedAnswers`.

also add kdoc.

datacapture/src/main/java/com/google/android/fhir/datacapture/QuestionnaireViewModel.kt
```diff
@@ -198,6 +197,9 @@ internal class QuestionnaireViewModel(state: SavedStateHandle) : ViewModel() {
               )
             }
           if (enabled) {
```
jingtang10, line 199: Early return to make the code more readable:

```suggestion
          if (!enabled) {
            questionnaireResponseItem.removeAnswers()
            emptyList()
          }
```

datacapture/src/main/java/com/google/android/fhir/datacapture/QuestionnaireViewModel.kt
jingtang10: wait - i'm not sure i understand why this is done this way.

the problem raised in the issue is that answers to disabled questions are being extracted.

this function `getQuestionnaireState` is called to render the questionnaire. according to the logic here, if the user answers a question, you will override their answer with the intial value? this seems incorrect to me.

it seems to me the fix should actually be in the extraction code. rather than in the view model.

jingtang10 (changes requested):
@santosh-pingle i'm not sure this fix is correct - can you please address the comments? thanks!

datacapture/src/main/java/com/google/android/fhir/datacapture/QuestionnaireViewModel.kt
santosh-pingle: Please check, this is what you meant.

datacapture/src/main/java/com/google/android/fhir/datacapture/QuestionnaireViewModel.kt
santosh-pingle, line 206: Ok.
(this was a discussion which we had offline.)

Let’s assume that, question item is not enabled (enablewhen clause is false) and it has answers (either initial value or user input answer.)
Now, at the time of data extraction, there is no way that QuestionnaireResponse will tell whether a given questionnaire item is enabled or not.
Therefore, if the questionnaire item is not enabled then the answer should not be present in the questionnaire response.

Whether answers get overridden?

- If the question item is not enabled, then it will not have answers in questionnaireResponseItem, answers are removed.
- Now, let's assume the question item is enabled, so just update the questionnaire item response with initial values. 
- If it has already answered it will not override the answer, it checks hasenablewhen and does not have an answer.

jingtang10: you're correct in saying that from the `QuestionnaireResponse` alone you can't tell if a question is enabled, but you can tell that from inspecting the `Questionnaire` itself no? so one solution is to do this in the resource mapper.

even if you do decide to do this in view model (which is an alternative solution), it should not be in this function, i think it probably should be when the answer changes (so that the enable when evaluation would change) so it probably should be in the `questionnaireResponseItemChangedCallback`?

overriding answer: i still don't think this is right. the user may decide to remove an initial answer. in your case you would then immediately add the intial answer back.

santosh-pingle: We can handle it in ```questionnaireResponseItemChangedCallback```, but still we need to update initial values in response items once it enabled, as answers were already removed when question item was disabled (as per fix).
Consider below case ->
There are question items which are enabled but not displayed to the user. Initial values are only answers and users can not input it. If these questionnaire items get disabled then their answers also get removed (as per fix). Now, if they get enabled again then we need to add those initial values back to response items as those are the only answers.

jingtang10: Can you try to not remove the intial answer when disabling the question?

santosh-pingle: https://chat.fhir.org/#narrow/stream/179255-questionnaire/topic/Questionnaire.20item.20response.20and.20disabled.20questionnaire.20item/near/251752110

santosh-pingle: As you suggested in developers sdk call and reference to above link, we will retain answers of disabled questions in questionnaire response, and from ``` getQuestionnaireResponse ``` we will return separate copy of questionnaire response which will have answers of enabled questions only.

santosh-pingle: Working on change.

santosh-pingle:
Can we please review it.

datacapture/src/main/java/com/google/android/fhir/datacapture/QuestionnaireViewModel.kt
```diff
@@ -197,31 +209,62 @@ internal class QuestionnaireViewModel(state: SavedStateHandle) : ViewModel() {
...
+        questionnaireResponseItemList =
+          if (questionnaireResponseItem.answer.isEmpty()) {
+            questionnaireResponseItem.item
+          } else {
+            questionnaireResponseItem.answer.first().item
+          }
+      )
+    }
+  }
 }
```
jingtang10, line 268: when do you actually remove the item from the list? i'm struggling to find the line that actually does it. all this does is that if something it's not enabled you early return... in that case the item is still in the list though?

what about turn this around and use the `filter` function and generate an enabled list like below:

```suggestion
  private fun getEnableResponseItems(
    questionnaireItemList: List<Questionnaire.QuestionnaireItemComponent>,
    questionnaireResponseItemList: List<QuestionnaireResponse.QuestionnaireResponseItemComponent>,
  ): List<QuestionnaireResponse.QuestionnaireResponseItemComponent> {
    return questionnaireItemList
      .asSequence()
      .zip(questionnaireResponseItemList.asSequence())
      .filter { (questionnaireItem, questionnaireResponseItem) ->
        EnablementEvaluator.evaluate(questionnaireItem) { linkId ->
          val questionnaireItemComponent = linkIdToQuestionnaireItemMap[linkId]
          val questionnaireResponseItemComponent = linkIdToQuestionnaireResponseItemMap[linkId]
          if (questionnaireItemComponent == null || questionnaireResponseItemComponent == null) {
            return@evaluate QuestionnaireItemWithResponse(null, null)
          }
          QuestionnaireItemWithResponse(
            questionnaireItem = questionnaireItemComponent,
            questionnaireResponseItem = questionnaireResponseItemComponent
          )
        }
      }
      .map { (questionnaireItem, questionnaireResponseItem) ->
        // Nested group items
        questionnaireResponseItem.item =
          getEnableResponseItems(questionnaireItem.item, questionnaireResponseItem.item)

        // Nested question items
        questionnaireResponseItem.answer.forEach {
          it.item = getEnableResponseItems(questionnaireItem.item, it.item)
        }

        questionnaireResponseItem
      }
      .toList()
  }
```

datacapture/src/test/java/com/google/android/fhir/datacapture/QuestionnaireViewModelTest.kt
```diff
@@ -813,6 +813,199 @@ class QuestionnaireViewModelTest {
     }
   }
 
+  @Test
+  fun removeAnswers_questionnaireResponseItemHasAnswers_removeAnswers() {
```
jingtang10, line 817: up to you but i don't think you need to test this private function

jingtang10 (changes requested):

datacapture/src/main/java/com/google/android/fhir/datacapture/QuestionnaireViewModel.kt
santosh-pingle: Please refer further discussion in below comments.

datacapture/src/main/java/com/google/android/fhir/datacapture/QuestionnaireViewModel.kt
santosh-pingle: Answers for disabled question was getting removed with below change.
return@forEach questionnaireResponseItem.removeAllNestedAnswers()

But as you suggested not to include questionnaireResponse item to questionnaire response for disabled question, this change is updated.

datacapture/src/main/java/com/google/android/fhir/datacapture/QuestionnaireViewModel.kt
```diff
@@ -197,31 +206,66 @@ internal class QuestionnaireViewModel(state: SavedStateHandle) : ViewModel() {
...
+                // we're now dealing with nested items, so pagination is no longer a concern
+                pagination = null,
+              )
+              .items
         }
         .toList()
     return QuestionnaireState(items = items, pagination = pagination)
   }
+
+  private fun getEnableResponseItems(
```
jingtang10, line 237:
```suggestion
  private fun getEnabledResponseItems(
```

datacapture/src/main/java/com/google/android/fhir/datacapture/QuestionnaireViewModel.kt
```diff
@@ -197,31 +206,66 @@ internal class QuestionnaireViewModel(state: SavedStateHandle) : ViewModel() {
...
+    questionnaireItemList: List<Questionnaire.QuestionnaireItemComponent>,
+    questionnaireResponseItemList: List<QuestionnaireResponse.QuestionnaireResponseItemComponent>,
+  ): List<QuestionnaireResponse.QuestionnaireResponseItemComponent> {
+    return questionnaireItemList
+      .asSequence()
+      .zip(questionnaireResponseItemList.asSequence())
+      .filter { (questionnaireItem, questionnaireResponseItem) ->
+        EnablementEvaluator.evaluate(questionnaireItem) { linkId ->
+          val questionnaireItemComponent = linkIdToQuestionnaireItemMap[linkId]
+          val questionnaireResponseItemComponent = linkIdToQuestionnaireResponseItemMap[linkId]
```
jingtang10, line 247:
```suggestion
          val questionnaireItem = linkIdToQuestionnaireItemMap[linkId]
          val questionnaireResponseItem = linkIdToQuestionnaireResponseItemMap[linkId]
```

datacapture/src/main/java/com/google/android/fhir/datacapture/QuestionnaireViewModel.kt
```diff
@@ -197,31 +206,66 @@ internal class QuestionnaireViewModel(state: SavedStateHandle) : ViewModel() {
...
+          if (questionnaireItemComponent == null || questionnaireResponseItemComponent == null) {
+            return@evaluate QuestionnaireItemWithResponse(null, null)
+          }
+          QuestionnaireItemWithResponse(
+            questionnaireItem = questionnaireItemComponent,
+            questionnaireResponseItem = questionnaireResponseItemComponent
```
jingtang10, line 253:
```suggestion
            questionnaireItem,
            questionnaireResponseItem
```

jingtang10 (approved):
a few small comments.. once addressed please feel free to merge!

nice work thanks @santosh-pingle!

Issue #723, opened 2021-08-17
Allow disjunction of mulitple search parameters (i.e. name is joe or address is france)


jingtang10:
Discussed with @ekigamba @f-odhiambo @aditya-07 and @fredhersch yesterday. We identified 2 different use cases

1. while searching using a search parameter (i.e. given name), provide multiple possible matches that are joined together as a disjunction/OR. This is the example @ekigamba provided `("name" = ("joe" or "john")) AND ("age" = (11 or 12))` and this is what the FHIR RESTful APIs already support (not that we must follow the RESTful spec in our offline search API)
2. having multiple search parameters in a disjunction (OR) rather than conjunction (AND). At the moment when you list multiple search parameters in the search API, we add them on top of each other. In other words we apply them as filters and each filter further narrows down the resources. @ekigamba gave the example - we might want to search for patients whose name is XXX OR whose patient id is XXX. This is useful when in the UI we use 1 search box that searches for patients using multiple fields. In our discussion, we identified a way to get around this which is simply issue multiple search queries. However, I think in more complex cases the need for the search API to handle this properly is going to be more important (so the onus of deduplicating is not on the developer).

OK, so in this issue - let's focus on number 1. I will create another issue for number 2.

_Originally posted by @jingtang10 in https://github.com/google/android-fhir/issues/292#issuecomment-900113580_

jingtang10:
In this issue we will focus on number 2.

Tarun-Bhardwaj:
Code changes merged with PR #827

PR #786, opened 2021-09-15
Disjunction of multiple values or params in search filter

### LINKED ISSUE #292: Handle composite search parameters in the indexing code
https://github.com/google/android-fhir/pull/281#discussion_r589413068

https://www.hl7.org/fhir/search.html#composite

aditya-07:

Fixes #292 #723

**Description**
`and` and `or` operators are now available to specify conjunction or disjunction between filters. 

**Alternative(s) considered**

**Type**
Choose one: Feature 

**Checklist**


jingtang10 (commented):
Thanks @aditya-07! Great work!

I really like the new syntax using infix opeartions. It's very cool 😎  But I do wonder if we're just opening a can of worms by allowing very complex nested queries. Would it be sufficient if we simply allow 1 level of filter but let the developer specify if the operation is AND or OR?

I would rather than we support something simple, instead of supporting something complex but with significant performance issues.

This is also true for the change introducing the `has` operator with multiple resource types in a single search. I'm wondering if we should introduce a limit there too so you can't nest more than 1 level.

jingtang10:
it seems this is more #723 than #292?

PR #787, opened 2021-09-15
Encrypt FHIR engine database via SQLCipher

### LINKED ISSUE #392: Support encryption of sqllite DB with SQLCipher
Due to the sensitive nature of medical data, we need to re-assure users of their data privacy. This issue seeks to address that by supporting encryption of data generated when using the SDK.

We have a related discussion [here](https://github.com/OpenSRP/fhircore/issues/35#issuecomment-801774240) around possibly integrating with SQLCipher.

None at the moment. Open to discussion.

**Additional context**
https://github.com/sqlcipher/android-database-sqlcipher#using-sqlcipher-for-android-with-room SQLCipher + Room integration

Open to it

stevenckngaa:

Fixes #392

**Description**
Here are the changes:
1. Introduce a new dependency: SQLCipher for database encryption
2. Generate a private key in Android keystore for signing messages using HMAC SHA256.
3. Sign an empty string with the private key in 2. Then use the signed message as the SQLCipher security key.

**Alternative(s) considered**

Database encryption: SQLCipher is the one sqlite3 encryption library that is available.
Key encryption:

1. Implement an encryption wrapper around Datastore
     Pros: 
     - Avoid shared preferences issue, such as blocking disk I/O on UI thread, lack of type safety, inefficient thread usage / disk usage, no guarantee of data written after `apply()` etc
   
   Cons: 
     - more work
     - we don't benefit from the goodies of data store that much because we are only storing one key.
2. EncryptedSharedPreferences
     Pros: 
     - less work. We don't need to maintain the encryption code.
   
    Cons: 
     - See Datastore pros above
     - Only support Android M or above
3. Use HMAC SHA256 signing algorithm to sign an empty string. Then use the signed byte array as database key (**chosen**)
     Pros:
     - Easy to implement
     - No need to generate key in the application layer. Avoid a potential attack vector.
     - No need to store the encrypted database key to the disk. No need to handle I/O error if the encrypted database key isn't available.
     
     Cons:
      - Since this solution is backed by Android keystore. To avoid known keystore security issues, we only support Android M or above.

**Type**
Feature

N/A

**Checklist**


engine/src/main/java/com/google/android/fhir/FhirServices.kt
```diff
@@ -30,12 +31,20 @@ internal data class FhirServices(
 ) {
   class Builder(private val context: Context) {
     private var inMemory: Boolean = false
+    private var enableEncryption: Boolean = false
 
     internal fun inMemory() = apply { inMemory = true }
 
+    internal fun enableEncryption() = apply { enableEncryption = true }
```
stevenckngaa, line 38: Should we check if Android API >= M here?

engine/src/androidTest/java/com/google/android/fhir/db/impl/DatabaseImplTest.kt
```diff
@@ -68,8 +71,11 @@ import org.junit.runner.RunWith
  * * Robolectric's SQLite implementation does not match Android, e.g.:
  * https://github.com/robolectric/robolectric/blob/master/shadows/framework/src/main/java/org/robolectric/shadows/ShadowSQLiteConnection.java#L97
  */
-@RunWith(AndroidJUnit4::class)
+@MediumTest
+@RunWith(Parameterized::class)
 class DatabaseImplTest {
+  @JvmField @Parameterized.Parameter(0) var encrypted: Boolean = false
```
jingtang10, line 77: add a little bit of explanation on what this does to the test please thanks!

engine/src/main/java/com/google/android/fhir/FhirServices.kt
```diff
@@ -30,12 +33,30 @@ internal data class FhirServices(
 ) {
   class Builder(private val context: Context) {
     private var inMemory: Boolean = false
+    private var enableEncryption: Boolean = false
+    private var databaseErrorStrategy = DatabaseErrorStrategy.UNSPECIFIED
 
     internal fun inMemory() = apply { inMemory = true }
 
+    internal fun enableEncryption() = apply {
```
jingtang10, line 41: is there any value in using the `apply` function here?

same with the function above `inMemory()`

engine/src/main/java/com/google/android/fhir/FhirEngineProvider.kt
```diff
@@ -29,8 +39,39 @@ object FhirEngineProvider {
...
+   * If unspecified, all database errors will be propagated to the call site. The caller shall
+   * handle the database error on a case-by-case basis.
+   */
+  UNSPECIFIED,
+
+  /**
+   * If a database error occurs at open, automatically recreate the database.
+   *
+   * This strategy is NOT respected when opening a previously unencrypted database with an encrypted
+   * configuration or vice versa.
```
jingtang10, line 74: can you add what would happen in those cases?

engine/src/main/java/com/google/android/fhir/FhirEngineProvider.kt
```diff
@@ -29,8 +39,39 @@ object FhirEngineProvider {
...
-      fhirEngine = FhirServices.builder(context.applicationContext).build().fhirEngine
+      if (!::fhirEngineConfiguration.isInitialized) {
+        fhirEngineConfiguration = FhirEngineConfiguration(enableEncryption = false, UNSPECIFIED)
+      }
+      fhirEngine =
+        FhirServices.builder(context.applicationContext)
+          .apply {
+            if (fhirEngineConfiguration.enableEncryption) enableEncryption()
+            setDatabaseErrorStrategy(fhirEngineConfiguration.databaseErrorStrategy)
+          }
```
jingtang10, line 50: the more i look at this block the more i feel FhirServices should just be inlined in this class.. not sure how you feel about that? no need to do that in this PR but i think this can be a follow-up.

@yigit

engine/src/main/java/com/google/android/fhir/db/DatabaseEncryptionException.kt
```diff
@@ -0,0 +1,158 @@
...
+            }
+          return encryptionException
+        }
+      }
+    }
+    return DatabaseEncryptionException(this, UNKNOWN)
+  }
+
+@Suppress("Unused")
+object KeyStoreExceptionErrorCode {
```
jingtang10, line 94: is there some code we can point to in our open source repo that is the origin of these errors? that we might need to keep these things up to date :/

reference/src/main/java/com/google/android/fhir/reference/FhirApplication.kt
```diff
@@ -33,6 +35,7 @@ class FhirApplication : Application() {
   }
 
   private fun constructFhirEngine(): FhirEngine {
+    FhirEngineProvider.init(FhirEngineConfiguration(enableEncryption = true, RECREATE_AT_OPEN))
```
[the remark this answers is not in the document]
jingtang10: i don't know if we should be more lenient here. if this code is called accidentally by developers with FhirEngineProvider still alive, there will be exceptions. is this what we want to happen?

engine/src/main/java/com/google/android/fhir/db/impl/SQLCipherSupportHelper.kt
```diff
@@ -0,0 +1,101 @@
...
+
+  override fun getWritableDatabase(): SupportSQLiteDatabase? {
+    val result =
+      try {
+        standardHelper.getWritableDatabase(passphraseFetcher())
+      } catch (ex: SQLiteException) {
+        if (databaseErrorStrategy == DatabaseErrorStrategy.RECREATE_AT_OPEN) {
+          Log.w(LOG_TAG, "Fail to open database. Recreating database.")
+          configuration.context.getDatabasePath(databaseName).delete()
+          standardHelper.getWritableDatabase(passphraseFetcher())
```
jingtang10, line 84: is it possible at all for this to throw for some other reason? or is it guaranteed to succeed?

engine/src/main/java/com/google/android/fhir/db/impl/DatabaseImpl.kt
```diff
@@ -38,97 +51,188 @@ import org.hl7.fhir.r4.model.ResourceType
...
+  databaseConfig: DatabaseConfig
+) : com.google.android.fhir.db.Database {
 
   val builder =
-    if (inMemory) {
-      Room.inMemoryDatabaseBuilder(context, ResourceDatabase::class.java)
-    } else {
-      Room.databaseBuilder(context, ResourceDatabase::class.java, DEFAULT_DATABASE_NAME)
+    when {
+      databaseConfig.inMemory -> Room.inMemoryDatabaseBuilder(context, ResourceDatabase::class.java)
```
jingtang10, line 62: can we make it not possible to set the db config with `inMemory` and `enableEncryption` both true?

jingtang10 (changes requested):

engine/src/main/java/com/google/android/fhir/db/impl/DatabaseImpl.kt
```diff
@@ -130,5 +149,45 @@ internal class DatabaseImpl(context: Context, private val iParser: IParser, inMe
...
   }
 }
+
+data class DatabaseConfig(
+  val inMemory: Boolean,
+  val enableEncryption: Boolean,
+  val databaseErrorStrategy: DatabaseErrorStrategy
+)
+
+suspend fun <R> RoomDatabase.withWrappedTransaction(
```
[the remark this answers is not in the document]
jmarkoff: You can test with an emulator as well, but I'm not sure of a good way to test the keystore without an emulator or real device unfortunately. You would need to build out some stubs. For Jetpack Security, we just require that you test on a device (real or virtual).

engine/src/main/java/com/google/android/fhir/db/DatabaseEncryptionException.kt
jmarkoff: They are only defined in the platform code sadly. Generally they shouldn't change to often though.

https://android.googlesource.com/platform/frameworks/base/+/master/core/java/android/security/keymaster/KeymasterDefs.java

engine/src/main/java/com/google/android/fhir/db/DatabaseEncryptionException.kt
```diff
@@ -0,0 +1,158 @@
...
+import com.google.android.fhir.db.KeyStoreExceptionErrorCode.ERROR_UNSUPPORTED_PADDING_MODE
+import com.google.android.fhir.db.KeyStoreExceptionErrorCode.ERROR_UNSUPPORTED_PURPOSE
+import com.google.android.fhir.db.KeyStoreExceptionErrorCode.ERROR_UNSUPPORTED_TAG
+import java.security.KeyStoreException
+
+/**
+ * An database encryption exception wrapper which maps comprehensive keystore errors to a limited
+ * set of actionable errors.
+ */
+class DatabaseEncryptionException(
```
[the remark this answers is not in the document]
jmarkoff: It's something that I had to write, the error codes are defined in the platform code only.
https://android.googlesource.com/platform/frameworks/base/+/master/core/java/android/security/keymaster/KeymasterDefs.java

The code that I shared with you wraps all keystore actions with this type of exception wrapper.

reference/src/main/java/com/google/android/fhir/reference/FhirApplication.kt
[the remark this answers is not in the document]
stevenckngaa: I don't think app developers should call the initialization anywhere apart from `Application#onCreate`. [`WorkManager#initialize`](https://developer.android.com/reference/kotlin/androidx/work/WorkManager#initialize) throw an exception if it is called more than once in the same process. I think we may also want to throw an exception if the app developer access database at direct boot, which is out of this PR scope.

@yigit FYI

engine/src/main/java/com/google/android/fhir/db/impl/DatabaseImpl.kt
yigit: do we care about that? I mean, it is not great that it wont work but also it has no negative impact on the actual use case (testing)

engine/src/main/java/com/google/android/fhir/db/impl/DatabaseImpl.kt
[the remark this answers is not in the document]
yigit: this is going to be easy to get wrong. as in you are not really convering all of the access to room database.

Any reason why we cannot do this when reading the key in open helper callback instead of here? I mean, do the retry in the open helper callback (and literally block by blocking the thread so that the rest of the database access do not get complicated?)

engine/src/main/java/com/google/android/fhir/db/impl/DatabaseImpl.kt
```diff
@@ -38,97 +51,188 @@ import org.hl7.fhir.r4.model.ResourceType
...
+  context: Context,
+  retryAttempt: Int = 0,
+  block: suspend () -> R
+): R {
+  require(retryAttempt >= 0) { "$LOG_TAG: Database retry attempt must not be a negative integer" }
+  return withContext(CoroutineScope(Dispatchers.IO).coroutineContext) {
```
yigit, line 198: unrelated nit, you can just do withContext(Dispatchers.IO). but as mentioned above, we should try to avoid leaking this complexity to here and instead handle it when we need to read the key from keystore.

engine/src/main/java/com/google/android/fhir/db/impl/SQLCipherSupportHelper.kt
```diff
@@ -0,0 +1,101 @@
...
+import net.sqlcipher.database.SQLiteDatabaseHook
+import net.sqlcipher.database.SQLiteException
+import net.sqlcipher.database.SQLiteOpenHelper
+
+/** A [SupportSQLiteOpenHelper] which initializes a [SQLiteDatabase] with a passphrase. */
+class SQLCipherSupportHelper(
+  private val configuration: SupportSQLiteOpenHelper.Configuration,
+  hook: SQLiteDatabaseHook? = null,
+  private val databaseErrorStrategy: DatabaseErrorStrategy,
+  private val passphraseFetcher: () -> ByteArray,
```
yigit, line 33: so this code could literally retry itself.
I know it will need to block a thread (not as great as doing a delay in coroutines) but the advantage is that you isolate it to the lower layer database and guaranteed that all access in room handles it.

reference/src/main/java/com/google/android/fhir/reference/FhirApplication.kt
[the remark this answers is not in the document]
yigit: yea i think following up what WM does is the right way to go for consistency

yigit (changes requested):

engine/src/main/java/com/google/android/fhir/db/impl/SQLCipherSupportHelper.kt
stevenckngaa: Right, the only reason that I didn't do it here was the thread blocking problem you mentioned. Sounds like we are okay with it. I am happy to move it back to this class.

engine/src/androidTest/java/com/google/android/fhir/db/impl/DatabaseImplTest.kt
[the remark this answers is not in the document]
stevenckngaa: Whether it is var or val, the compiler forces me to add a default value. I prefer `var` because `val`with default value false gives an impression that this value is always false.

I can't use `lateinit` because this is a JvmField. Also checked other implementations in our internal codebase. Everyone is using var with a default value.

jingtang10 (approved):
I ❤️ this. This is #lit.

Thanks @stevenckngaa for this exemplary work!

Left some comments but mostly cosmetic.

jingtang10 (approved):

buildSrc/src/main/kotlin/Dependencies.kt
```diff
@@ -78,50 +78,51 @@ object Dependencies {
...
   const val robolectric = "org.robolectric:robolectric:${Versions.robolectric}"
   const val truth = "com.google.truth:truth:${Versions.truth}"
-  const val flexBox = "com.google.android.flexbox:flexbox:${Versions.flexBox}"
 
   object Versions {
     object Androidx {
       const val activity = "1.2.1"
       const val appCompat = "1.1.0"
       const val constraintLayout = "1.1.3"
+      const val datastorePref = "1.0.0-rc02"
```
yigit, line 118: FYi this has 1.0.0 now.

yigit (approved):

buildSrc/src/main/kotlin/Dependencies.kt
stevenckngaa: This change doesn't introduce the datastorePref. Jing asked me to sort dependencies in the lexicographic order. I'll make a change in a separate PR. Cheers!

PR #801, opened 2021-09-28
Open Choice widget

### LINKED ISSUE #689: Add support for open-choice answerFormats in Questionnaire
Current SDC library does not support open-choice option

Able to provide open-choice as an item type. Specifying this will then implement the behaviour as per the specification here

**Additional context**
Here is a sample Questionnaire that uses open-choice options - https://openhie.github.io/case-reporting-hiv-vn/Questionnaire-hiv-case-report-questionnaire.json.html

There are two scenarios to support:

**1. No list to select from:**
There are no choices to select from and open-choice just provides a text box and the answer is stored as a String
```
{ 
"linkId" : "confirming_lab",
 "text" : "Confirming lab", 
"type" : "open-choice", 
"repeats" : false 
},
```

2. **When there is an answerValueSet (or answerOptions) specified**
open-choice renders a selection list (e.g. a drop-down) with an 'other' option which is selected then shows a text box to be populated
e.g here: there would be a list of occupations and an other option
```
{ "linkId" : "occupation", 
"text" : "Occupation", 
"type" : "open-choice", 
"repeats" : false, 
"answerValueSet" : "http://openhie.org/fhir/openhie.vn.case-reporting.hiv/ValueSet/vs-hiv-occupation"
 }
```

kevinmost:

Fixes #689

**Description**
This PR refactors out `OptionSelectDialogFragment` from the previous `MultiSelectDialogFragment`. `OptionSelectDialogFragment` now supports all permutations of the following options:

- `addOtherOptions`: If this is true, the dialog will append an additional option called "Other". When this option is selected, if `multiSelectEnabled = false`, the user will see a single EditText in which they can enter a free-form option. If `multiSelectEnabled = true`, the user will see an "Add another answer" button. When clicked, this button will show additional EditTexts (which will have delete buttons next to them).
- `multiSelectEnabled`: This is controlled by the `repeats` option on a questionnaire item. This decides whether checkboxes or radio buttons are used, and affects the behavior of `addOtherOptions` as detailed above.

**Alternative(s) considered**
Discussed offline. This approach gives us the most flexibility and conforms best to FHIR data standards.

**Type**
Choose one: (Feature)

**Checklist**


datacapture/src/main/java/com/google/android/fhir/datacapture/views/OptionSelectDialogFragment.kt
```diff
@@ -0,0 +1,360 @@
...
+    /** Whether the "other" option will be exposed for free-form text input. */
+    val addOtherOptions: Boolean,
+  )
+
+  private val viewModel: QuestionnaireItemDialogSelectViewModel by activityViewModels()
+
+  private val questionLinkId: String by lazy { arguments?.getString(KEY_QUESTION_LINK_ID)!! }
+
+  override fun onCreateDialog(savedInstanceState: Bundle?): Dialog {
+    isCancelable = false
```
jingtang10, line 60: why should this not be cancellable?

also is this duplicated with line 82?

datacapture/src/main/java/com/google/android/fhir/datacapture/views/OptionSelectDialogFragment.kt
```diff
@@ -0,0 +1,360 @@
...
+    return AlertDialog.Builder(requireContext())
+      .setTitle(title)
+      .setView(view)
+      .setPositiveButton(android.R.string.ok) { _, _ -> saveSelections(adapter.currentList) }
+      .setNegativeButton(android.R.string.cancel) { _, _ -> }
+      .setCancelable(false)
+      .create()
+  }
+
+  /** Saves the current RecyclerView adapter into the ViewModel. */
```
jingtang10, line 86:
```suggestion
  /** Saves the current selections in the RecyclerView into the ViewModel. */
```

datacapture/src/main/java/com/google/android/fhir/datacapture/views/OptionSelectDialogFragment.kt
```diff
@@ -0,0 +1,360 @@
...
+  val title: String,
+  val config: Config,
+) : DialogFragment() {
+
+  /** Configures this [OptionSelectDialogFragment]. */
+  data class Config(
+    /** Whether multi-select will be enabled or not. */
+    val multiSelect: Boolean,
+    /** Whether the "other" option will be exposed for free-form text input. */
+    val addOtherOptions: Boolean,
```
jingtang10, line 52:
```suggestion
    val canAddOtherOptions: Boolean,
```

or maybe `otherOptionsAllowed`?

datacapture/src/main/java/com/google/android/fhir/datacapture/views/OptionSelectDialogFragment.kt
```diff
@@ -0,0 +1,360 @@
...
+          options = currentList.filterIsInstance<OptionSelectRow.Option>().map { it.option },
+          otherOptions =
+            currentList.filterIsInstance<OptionSelectRow.OtherEditText>().map { it.currentText },
+        )
+      )
+    }
+  }
+
+  /** Converts the initial ViewModel state into rows to display in the RecyclerView adapter. */
+  private fun SelectedOptions.toOptionRows(): List<OptionSelectRow> {
```
jingtang10, line 101: to be a bit more defensive here -- do we need to handle cases where there might be multiple other options but the question is a single select? in that case shall we keep only one custom option and drop others? or in the radio button cases would the last one be selected using the current code?

jingtang10 (commented):
This is exemplary work. Thanks @kevinmost!

One design decision in the single select case: Should "Other" and the edit text box be on the same line? I think there might be space concern.. but if they're not on the same line, do we want to hide the edit text until other is selected? Can we get some UX input?

In the multi select case - do we even need to have an "other" checkbox? Can we just not have it?

Alternatively, we can have an edit text with an "add option" button which if you click will convert what's in the edit text to a checkbox that is automatically ticked (and can't be unticked, but can be removed)...

shelaghm:
UX perspective

1. I think we can do it on the same line if there is space (might work in English and larger phone sizes, but I worry about how this will scale across phones/languages). Attached screenshot below of what this can look like. Google forms does something similar to this and I think it works well.
2. Agree, don't show the enter text field until other is selected.
3. Re: removing multi-select "other" checkbox: I'm curious to hear more why you propose remove it? I'm not sure I'm fully following, do you mean (a) removing the whole option of "other" in the list of options and replacing it with a button that says "add your own" or just (b) removing the checkbox? My gut reaction is that I think having an other option in the list is maybe more of a common pattern, but not opposed to adding a button instead. 
4. What are your thoughts on the remove icon, do you think the (-) [https://fonts.google.com/icons?selected=Material%20Icons%20Outlined%3Aremove_circle_outline%3A](url) or [trash icon](https://fonts.google.com/icons?selected=Material%20Icons%20Outlined%3Adelete%3A)? 

@kevinmost @jingtang10 

Screen Shot 2021-09-29 at 3 45 58 PM

jingtang10:
Thanks @shelaghm!

I don't really think the "other" check box does much in the multi select case. Can the user untick the box and will the user input options still be there but just not selected as answers (discarded when you submit the questionnaire)?

To me, in the multi select case at least, the presence of the user input indicates the custom options should be selected. to remove them that's what the remove button is for.

jingtang10:
re 4 i like "-" because it's simple and clear. the trash icon looks very busy and seems highly contextual... (maybe to some people trash cans don't look exactly like that?)

jingtang10:
i guess i'm thinking about something like this below regarding removing the others check box in the multi select case. Basically just edit text fields for your custom options, and remove buttons next to them. saves the user from ticking the others box first. and we can add a "hint" in the text box that "enter other options here" or something similar.

Screen Shot 2021-09-30 at 18 45 38

shelaghm:
@jingtang10 

Thanks for elaborating. I hear you on that it seems more clear to use the remove '-' button than to uncheck the other box if you want to unselect/remove the other choice. The 'add other' button could also help with people typing in the text, rather than just selecting other without adding the descriptive text.  I mocked up what it could look like with an 'add other' button and without "other" listed in the list of options and I think the pattern works. see below

Is there a scenario where we want to enable selecting other, but not adding edit text?

Screen Shot 2021-10-04 at 9 27 52 AM

datacapture/src/main/java/com/google/android/fhir/datacapture/views/OptionSelectDialogFragment.kt
```diff
@@ -0,0 +1,370 @@
...
+
+  /** Configures this [OptionSelectDialogFragment]. */
+  data class Config(
+    /** Whether multi-select will be enabled or not. */
+    val multiSelect: Boolean,
+    /** Whether the "other" option will be exposed for free-form text input. */
+    val otherOptionsAllowed: Boolean,
+  )
+
+  private val viewModel: QuestionnaireItemDialogSelectViewModel by activityViewModels()
```
santosh-pingle, line 55: @kevinmost @jingtang10 
Is the reason we instantiated QuestionnaireItemDialogSelectViewModel using activityViewModels() because it is also shared by QuestionnaireItemDialogSelectViewHolderFactory.kt? The only drawback I see is that even though the QuestionnaireFragment (which is used to render the QuestionnaireViewItem) is not alive, the model is kept in memory, which raises issue #2055

Solution to fix the issue : 
Is it okay to change the replay count to 0? If the replay count is set to 0, I do not see the issue existing. The selected option is also stored as an answer, so if the same view is rendered, it can be shown as selected.

datacapture/src/main/java/com/google/android/fhir/datacapture/views/QuestionnaireItemDialogSelectViewHolderFactory.kt
```diff
@@ -0,0 +1,193 @@
...
+      override fun init(itemView: View) {
+        holder = DialogSelectViewHolder(itemView)
+      }
+
+      override fun bind(questionnaireItemViewItem: QuestionnaireItemViewItem) {
+        val activity =
+          requireNotNull(holder.question.context.tryUnwrapContext()) {
+            "Can only use dialog select in an AppCompatActivity context"
+          }
+        val viewModel: QuestionnaireItemDialogSelectViewModel by activity.viewModels()
```
santosh-pingle, line 63: please refer earlier comment.

PR #851, opened 2021-10-24
Add gradle task to generate code at runtime

### LINKED ISSUE #831: create new gradle task to run code-generation at build time
> > 1. I'm not very familiar with poet. But what is the workflow if we're changing the code? Do we need to regenerate the code and merge in the same PR?
> 
> As of now we don't need to. But I think if we add a gradle task (as @yigit suggested) we could generate these files at build time.

so at the moment you need to run the `main` function in the `codegen` model manually to generate the new file? can you create a follow-up issue to handle the creation of the new gradle task?

_Originally posted by @jingtang10 in https://github.com/google/android-fhir/issues/773#issuecomment-939864223_

epicadk:

Fixes #831 

**Description**

1.  Add gradle task to generate indexing code at runtime
2.  Ensure that `:engine:build` depends on the aforementioned task.

**Alternative(s) considered**
nil

**Type**
Choose one: Feature

**Checklist**
- [ ] I have read and acknowledged the [Code of conduct](https://github.com/google/android-fhir/blob/master/docs/CODE_OF_CONDUCT.md)
- [ ] I have read [How to Contribute](https://github.com/google/android-fhir/blob/master/docs/contributing.md)
- [ ] I have read the [Developer's guide](https://github.com/google/android-fhir/wiki/Developer's-Guide)
- [ ] I have signed the Google [Individual CLA](https://cla.developers.google.com/about/google-individual), or I am covered by my company's [Corporate CLA](https://cla.developers.google.com/about/google-corporate )
- [ ] I have run `./gradlew spotlessApply` and `./gradlew spotlessCheck` to check my code follows the style guide of this project
- [ ] I have run `./gradlew check` and `./gradlew connectedCheck` to test my changes locally
- [ ] I have built and run the reference app(s) to verify my change fixes the issue and/or does not break the reference app(s)


codegen/src/main/java/com/google/android/fhir/codegen/SearchParameterRepositoryGenerator.kt
```diff
@@ -126,9 +126,15 @@ object SearchParameterRepositoryGenerator {
...
-fun main() {
-  val sp = File(inputFilePath)
+/**
+ * @param args The command line arguments. args[0] should contain the input file location. args[1]
+ * should contains the output file location. If the arguments are absent the default values
+ * [inputFilePath] & [outputFilePath] will be used. Useful when the runtime classpath is not the
+ * root project directory.
+ */
+fun main(args: Array<String>) {
+  val sp = File(args.getOrElse(0) { inputFilePath })
```
jingtang10, line 136:
```suggestion
  val searchParamDef = File(args.getOrElse(0) { inputFilePath })
```

codegen/src/main/java/com/google/android/fhir/codegen/SearchParameterRepositoryGenerator.kt
```diff
@@ -126,9 +126,15 @@ object SearchParameterRepositoryGenerator {
 private const val inputFilePath = "codegen/src/main/res/search-parameters.json"
 private const val outputFilePath = "engine/src/main/java"
 
-fun main() {
-  val sp = File(inputFilePath)
+/**
+ * @param args The command line arguments. args[0] should contain the input file location. args[1]
+ * should contains the output file location. If the arguments are absent the default values
+ * [inputFilePath] & [outputFilePath] will be used. Useful when the runtime classpath is not the
+ * root project directory.
```
jingtang10, line 133: please add that this is used in `codegen/gradle.build.kts` file.

also did you add this because this causes issues on CI?

jingtang10 (approved):

codegen/src/main/java/com/google/android/fhir/codegen/SearchParameterRepositoryGenerator.kt
epicadk: Yes, the functions works fine ( with relative paths) when I run it from within the IDE, however in the gradle task the relative paths don't work. I think this is because of the classpath.

epicadk:
> > Do we also want github actions to commit the changed code to the branch? It is possible to do however, I not entirely sure if the CLA bot would allow this.
> 
> how should CI/review etc be handled? i think perhaps we don't need this now and new code changes should be manually commited with the PR that changes the search parameter definitions.

I agree. I'll create a follow up issue to fail the workflow if the generated file and the committed file do not match.

PR #900, opened 2021-10-29
Always canonicalize quantity values

### LINKED ISSUE #733: Refactoring quantity index table by removing the canonical columns
> > I don't mind keeping only canonicalized quantities in Index. Could we make it explicit about not supporting case 4 in @epicadk comment?
> 
> Another possible option would be to not support canonical match at all as discussed with @jingtang10 . I think that makes more sense than this.

My guess is that it's unlikely for users to search by a quantity using the equal operator. I would imagine quantity would be used for sorting. But also probably used for search within a range.

If this is true, I would argue for canonicalisation. Although to make sorting work perfectly with different units there will be much more work.

_Originally posted by @jingtang10 in https://github.com/google/android-fhir/issues/682#issuecomment-900587901_

jingtang10:

Fixes #733 

**Description**
Always canonicalize quantity values in the quantity index table.

**Alternative(s) considered**

**Type**
Choose one: Code health

**Checklist**

### DIFF
--- a/engine/src/main/java/com/google/android/fhir/search/MoreSearch.kt
+++ b/engine/src/main/java/com/google/android/fhir/search/MoreSearch.kt
@@ -361,48 +361,42 @@ internal fun getConditionParamPair(
   system: String?,
   unit: String?
 ): ConditionParam<Any> {
-  // value cannot be null -> the value condition will always be present
-  val valueConditionParam = getConditionParamPair(prefix, value)
-  val argList = mutableListOf<Any>()
+  var canonicalizedUnit = unit
+  var canonicalizedValue = value
 
-  val condition = StringBuilder()
-  val canonicalCondition = StringBuilder()
-  val nonCanonicalCondition = StringBuilder()
-
-  if (system != null) {
-    argList.add(system)
-    condition.append("index_system = ? AND ")
-  }
-
-  if (unit != null) {
-    argList.add(unit)
-    nonCanonicalCondition.append("index_code = ? AND ")
-  }
-
-  nonCanonicalCondition.append(valueConditionParam.condition)
-  argList.addAll(valueConditionParam.params)
-
+  // Canonicalize the unit if possible. For example, 1000 g will be canonilized to 1 kg
   if (system == ucumUrl && unit != null) {
     try {
-      val ucumUnit = UnitConverter.getCanonicalForm(UcumValue(unit, value))
-      val canonicalConditionParam = getConditionParamPair(prefix, ucumUnit.value)
-      argList.add(ucumUnit.code)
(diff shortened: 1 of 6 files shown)


engine/src/main/java/com/google/android/fhir/search/MoreSearch.kt
```diff
@@ -330,48 +330,42 @@ internal fun getConditionParamPair(
   system: String?,
   unit: String?
 ): ConditionParam<Any> {
-  // value cannot be null -> the value condition will always be present
-  val valueConditionParam = getConditionParamPair(prefix, value)
-  val argList = mutableListOf<Any>()
+  var canonicalizedUnit = unit
+  var canonicalizedValue = value
+
+  // Canonicalize the unit if possible. For example, 1000 g will be canonilized to 1 kg
```
aditya-07, line 336: nit
```suggestion
  // Canonicalize the unit if possible. For example, 1000 g will be canonicalized to 1 kg
```

engine/src/main/java/com/google/android/fhir/search/MoreSearch.kt
```diff
@@ -330,48 +330,42 @@ internal fun getConditionParamPair(
...
-  }
-
-  // Add OR only when canonical match is possible
-  if (canonicalCondition.isNotEmpty()) {
-    condition.append("($nonCanonicalCondition OR $canonicalCondition)")
-  } else {
-    condition.append(nonCanonicalCondition)
-  }
-  return ConditionParam(condition.toString(), argList)
+  return ConditionParam(queryBuilder.toString(), argList)
```
aditya-07, line 368: The only negative case I can think of is if because of some reason canonicalization fails during the Resource Indexing, then as long as canonicalization works fine here, the value won't be found.

aditya-07 (approved):
Added a nit. Otherwise, LGTM.

engine/src/main/java/com/google/android/fhir/search/MoreSearch.kt
epicadk: gram is the base unit in UCUM.

jingtang10: udpated this

engine/src/main/java/com/google/android/fhir/search/MoreSearch.kt
jingtang10: should we file an issue or add a todo?

Tarun-Bhardwaj:
@jingtang10 , considering that the reviewer has approved these changes, can this PR be merged?

Issue #836, opened 2021-10-13
Crash while using the questionnaire response without disabled items generated by the SDK to resume form filling.


joiskash:
**Describe the bug**
After the decision to remove questionnaireResponse items which are not enabled from the Questionnaire Response (more details in #755) , the `validateQuestionnaireResponseItems` function in the  QuestionnaireViewModel had to be updated because it was initially written assuming that disabled questionnaireResponse Items would also be there in the QuestionnaireReponse with empty answers.

**Component**
SDC library

**To Reproduce**
1. Use a questionnaire with an item that has enable when
2. Answer the questionnaire such that enableWhen is false
3. Use the questionnaire response that is output and the questionnaire as input to the fragment
4. App will crash.

**Expected behavior**
App should not crash since the spec clearly mentions that the disabled answers "may" be present

> When a QuestionnaireResponse references a Questionnaire, all the items in the questionnaire SHOULD be included in the QuestionnaireResponse if they are relevant to the interpretation of the answers, including items of type = display. This **might** also include items that were >not enabled when the Questionnaire was answered, though of course these will not have answers

**Screenshots**

**Smartphone (please complete the following information):**
 - Device: [e.g. Pixel4a emulator]
 - Android version: [e.g. Settings -> About phone -> Android version]
 - Build number: [e.g. Settings -> About phone -> Build number]

**Additional context**

**Would you like to work on the issue?**
Yes

Issue #910, opened 2021-11-03
Allow rendering of complex questions with repeated answers and nested items, repeated groups and nested items


joiskash:
**Is your feature request related to a problem? Please describe.**
Currently the SDC library supports repeated answers, but not items nested in answers that can repeat. Also full support for repeated answers is not yet available in the SDC library

**Describe the solution you'd like**
To support the spec completely we need to be able to handle all the questionnaireItem types that are shown in the diagram bellow.

![Questionnaire ResponseTree_V2](https://user-images.githubusercontent.com/48797364/140041819-a73dea3d-bc95-4f97-99f2-5b11a8941408.jpg)

| linkId | item.repeats | item.type == group | hasNestedItem | isSupported |
| ------------- | ------------- | ------------- |  ------------- | ------------- |
| item_1  | true  | false  | true  | false  |
| not in image  | true | false  | false  | true  |
| item_2  | false  | false  | false  | true  |
| not in image  | false  | false  | true  | true  |
| item_group_3  | false  | true  | true  | true  |
| item__group_4  | true  | true  | true  | false  |

**Additional context**
This also requires some thinking from the UX/UI point of view. eg: How do we handle a checkbox type with nested items? do we keep extra buttons to clear the nested answers/ edit nested items? What will the view look like?

**Would you like to work on the issue?**
I had started some initial work :

joiskash:
@RaaziaTarique I had created this diagram to get a hold of how questionnaireResponseItems can occur for various questionnaireItem. It maybe of use while you implement the refactoring in #1369

f-odhiambo:
Any update on this @kevinmost? What's the ETA for this?

jingtang10:
I think we should focus on rendering of nested questions under repeated answers in this ticket. Repeated groups are handled by kevin in a different issue #726

PR #926, opened 2021-11-17
Change validateQuestionnaireResponseItems to allow questionnaire response with missing items

### LINKED ISSUE #912: QuestionnaireResponse validator not consider missing items and answers
**Describe the bug**
QuestionnaireResponse validator assumes that questionnaire and response items has 1-1 mapping in [validator](https://github.com/google/android-fhir/blob/master/datacapture/src/main/java/com/google/android/fhir/datacapture/validation/QuestionnaireResponseValidator.kt#L42). This leads to validating question item X against response item Y

```
while (questionnaireItemListIterator.hasNext() &&
      questionnaireResponseItemListIterator.hasNext())
```

**Component**
Data Capture

**To Reproduce**
1. Create Questionnaire with multiple items in groups having skip logic and assign validation to all
2. Skip one group from answer
3. Validate
4. See error on item A although answer is correct due to the fact that the item above was not answered due to skip logic

joiskash:

Fixes #912 #836

**Description**
Re-wrote validateQuestionnaireResponseItems to take the input questionnaire response and validate it against the questionnaire response that is created by using the questionnaire.

**Alternative(s) considered**
Yes the alternative was to use the input questionnaire response as the main questionnaire response and add items if and when enable was activated. This seemed to be a more complicated solution

**Type**
Bug fix

**Checklist**


joiskash:
@jingtang10 please take a look at this. There are two questions:

joiskash:
@jingtang10 please take a look at this. These two questions are dealt as follows:

1. How do we deal with a questionnaire response that has more items than the questionnaire? Currently throwing an error No matching questionnaire item for questionnaire response item
2. How do we deal with a questionnaire response that has less items than the questionnaire? Not throwing any error because questionnaire response may not have disabled items.

joiskash:
> Should the validation code live in QuestionnaireResponseValidator.kt instead?
That class is responsible for validating answers that are typed in via UI. I dont think we should mix them up. Maybe we change the name of this function to checkQuestionnaireResponseAgainstQuestionnaire

This function is independent of creating the questionnaire response item . What I have done is to use the questionnaire response item that is created with the questionnaire and only fill up answers from the input questionnaire response after verifying that 

1. linkIDs match and order of questions is maintained
2. type matches for the primitive types
3. only questionnaire items which allow repeats allow multiple answers
4. There are no extra questionnaire response items

Since we are not using the input questionnaire response as the main questionnaire response to build view items, we will not be generating view items without response in the tuple. This is to make sure that irrespective of the items present in the input questionnaire response, the SDK will allow the user to answer the questionnaire completely.

PS: input questionnaire response is the one that is passed to the fragment via the bundle. The main questionnaire response is what we create by going through the questionnaire hierarchy.

joiskash:
I have made two changes as per @jingtang10 suggestion

1. Moved the validation function to QuestionnaireResponseValidator.kt 
2. Made some change in the getQuestionnaireState function to add missing questionnaireResponseItems.

I felt that instead of allowing for empty questionnaireResponseItems in the QuestionnaireItemViewItem, we could just use existing code to create them while fetching the list for the state flow. This change also helps us in the futures by avoiding zipping questionnaires and questionnaireResponse together because of the details in #910

jingtang10:
@kevinmost can you please also take a look at this? thanks!

datacapture/src/main/java/com/google/android/fhir/datacapture/QuestionnaireViewModel.kt
```diff
@@ -248,14 +249,22 @@ internal class QuestionnaireViewModel(application: Application, state: SavedStat
...
-        .flatMap { (questionnaireItemAndIndex, questionnaireResponseItem) ->
-          val (index, questionnaireItem) = questionnaireItemAndIndex
-
+        .flatMap { (index, questionnaireItem) ->
+          var questionnaireResponseItem = questionnaireItem.createQuestionnaireResponseItem()
+
+          // If there is an enabled questionnaire response available then we use that. Or else we
+          // just use an empty questionnaireResponse Item
+          if (responseIndex < questionnaireResponseItemList.size &&
+              questionnaireItem.linkId.equals(questionnaireResponseItem.linkId)
```
kevinmost, line 263: Nit: Prefer `==` in Kotlin

datacapture/src/main/java/com/google/android/fhir/datacapture/QuestionnaireViewModel.kt
```diff
@@ -248,14 +249,22 @@ internal class QuestionnaireViewModel(application: Application, state: SavedStat
     pagination: QuestionnairePagination?,
   ): QuestionnaireState {
     // TODO(kmost): validate pages before switching between next/prev pages
+    var responseIndex = 0
     val items: List<QuestionnaireItemViewItem> =
       questionnaireItemList
         .asSequence()
         .withIndex()
```
kevinmost, line 256: Nit: Would be a bit more performant to use

```
.flatMapIndexed { index, questionnaireItem ->
```

datacapture/src/main/java/com/google/android/fhir/datacapture/validation/QuestionnaireResponseValidator.kt
```diff
@@ -59,4 +59,85 @@ object QuestionnaireResponseValidator {
...
+   * Traverse (DFS) through the list of questionnaire items and the list of questionnaire response
+   * items and check if the linkId of the matching pairs of questionnaire item and questionnaire
+   * response item are equal.
+   */
+  fun validateQuestionnaireResponseItems(
+    questionnaireItemList: List<Questionnaire.QuestionnaireItemComponent>,
+    questionnaireResponseInputItemList:
+      List<QuestionnaireResponse.QuestionnaireResponseItemComponent>,
+  ) {
+    val questionnaireResponseInputItemListIterator = questionnaireResponseInputItemList.iterator()
```
kevinmost, line 73: If these two need to have the same length (which it looks like they do, as we throw an error if they don't on line 137), I would recommend using `zip` and doing something like this:

```kotlin
require(questionnaireItemList.size == questionnaireResponseInputItemList.size) {
  "Unequal item amounts: ${questionnaireItemList.size} items and ${questionnaireResponseInputItemList.size} responses"
}
questionnaireItemList.zip(questionnaireResponseInputItemList)
  .forEach { (questionnaireItem, questionnaireResponseInputItem) -> ... }
```

datacapture/src/main/java/com/google/android/fhir/datacapture/validation/QuestionnaireResponseValidator.kt
```diff
@@ -59,4 +59,85 @@ object QuestionnaireResponseValidator {
...
+    val questionnaireItemListIterator = questionnaireItemList.iterator()
+
+    while (questionnaireResponseInputItemListIterator.hasNext()) {
+      // TODO: Validate type and item nesting within answers for repeated answers
+      // https://github.com/google/android-fhir/issues/286
+      val questionnaireResponseInputItem = questionnaireResponseInputItemListIterator.next()
+      if (questionnaireItemListIterator.hasNext()) {
+        val questionnaireItem = questionnaireItemListIterator.next()
+        if (!questionnaireItem.linkId.equals(questionnaireResponseInputItem.linkId)) {
```
kevinmost, line 82: Nit: I'd also do this via `require`, and use `==`.

```kotlin
require(questionnaireItem.linkId == questionnaireResponseInputItem.linkId) { "Mismatching..." }
```

datacapture/src/test/java/com/google/android/fhir/datacapture/QuestionnaireViewModelTest.kt
```diff
@@ -313,10 +353,165 @@ class QuestionnaireViewModelTest(private val questionnaireSource: QuestionnaireS
...
       }
     val questionnaireResponse = QuestionnaireResponse().apply { id = "a-questionnaire-response" }
+    val questionnaireResponseWithMissingItem =
+      QuestionnaireResponse().apply {
+        id = "a-questionnaire-response"
+        addItem(
+          QuestionnaireResponse.QuestionnaireResponseItemComponent().apply {
+            linkId = "a-link-id"
+            answer =
+              mutableListOf(
```
kevinmost, line 368: Does this need to be mutable?

datacapture/src/main/java/com/google/android/fhir/datacapture/validation/QuestionnaireResponseValidator.kt
joiskash: They don't have to be of the same length.. The questionnaire response may not have some items . This can happen if the item was disabled / not answered. That's why i removed the zip from the previous implementation.

joiskash: The error we throw on line 137 is to make sure the questionnaire response does not have more items than the questionnaire. But the questionnaire response can have less number of items. Thats why I am iterating over questionnaireResponseList and using the index to fetch the corresponding questionnaire item. You will find more details regarding this in #836 . Please let me know if it does not make sense.

datacapture/src/main/java/com/google/android/fhir/datacapture/QuestionnaireViewModel.kt
joiskash: done

datacapture/src/main/java/com/google/android/fhir/datacapture/QuestionnaireViewModel.kt
joiskash: done

datacapture/src/main/java/com/google/android/fhir/datacapture/validation/QuestionnaireResponseValidator.kt
joiskash: done

datacapture/src/test/java/com/google/android/fhir/datacapture/QuestionnaireViewModelTest.kt
joiskash: nope , i made changed it to `listOf` :)

jingtang10 (changes requested):

datacapture/src/main/java/com/google/android/fhir/datacapture/validation/QuestionnaireResponseValidator.kt
```diff
@@ -59,4 +59,85 @@ object QuestionnaireResponseValidator {
     }
     return linkIdToValidationResultMap
   }
+
+  /**
+   * Traverse (DFS) through the list of questionnaire items and the list of questionnaire response
+   * items and check if the linkId of the matching pairs of questionnaire item and questionnaire
+   * response item are equal.
+   */
+  fun validateQuestionnaireResponseItems(
```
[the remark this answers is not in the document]
joiskash: done, I have made this function a private helper and changed the names of the validate functions.

datacapture/src/main/java/com/google/android/fhir/datacapture/validation/QuestionnaireResponseValidator.kt
```diff
@@ -53,10 +53,104 @@ object QuestionnaireResponseValidator {
...
+            type != Questionnaire.QuestionnaireItemType.GROUP
+        ) {
+          if (!questionnaireItem.repeats && questionnaireResponseInputItem.answer.size > 1) {
+            throw IllegalArgumentException(
+              "Multiple answers in ${questionnaireResponseInputItem.linkId} and repeats false in " +
+                "questionnaire item ${questionnaireItem.linkId}"
+            )
+          }
+          questionnaireResponseInputItem.answer.forEachIndexed {
+            index,
```
kevinmost, line 112: Nit: Looks like the index is not needed, could this become `forEach` instead?

kevinmost (approved):

jingtang10 (approved):
Thanks @joiskash!

I think with this change, we should move a lot of the test cases from view model to the validation package. But that can be done in a follow-up.

datacapture/src/main/java/com/google/android/fhir/datacapture/validation/QuestionnaireResponseValidator.kt
joiskash: Will take care of it in my next PR. Thanks!

joiskash:
> Thanks @joiskash!
> 
> I think with this change, we should move a lot of the test cases from view model to the validation package. But that can be done in a follow-up.
> 
> 

Yes absolutely! That shouldn't be too hard to do.

PR #954, opened 2021-12-06
Super class field mapping Resource Extraction

### LINKED ISSUE #943: Handle inherited properties of Resource into questionnaire when doing extraction via ResourceMapper
**Describe the bug**
Fhirpath mapping into questionnaire allow mapping definition like below
 "definition": "http://hl7.org/fhir/StructureDefinition/Patient#Patient.name.given"

ResourceMapper processes it an automatically maps the questionnaire-response answer to mapping value. But this work only for declared properties. For inherited properties (from DomainResource, and REsource like tag, extension) it does not work 

**Component**
SDC library

**To Reproduce**
SDK has an issue with the way it works for fhir-path
```
private fun Class<*>.getFieldOrNull(name: String): Field? {
  return try {
    getDeclaredField(name)
  } catch (ex: NoSuchFieldException) {
    return null
  }
}
```
Instead of picking declared-field we should also pick fields inherited. Patient or any other resource extends DomainResource and the Resource. It means that we can not set inherited fields like meta.tag/extension etc using fhir-path.

maimoonak:
Fixes #943 

**Description**
Resource Extraction by Field allows mapping for parent class fields like DomainResource field extension for Patient class

**Alternative(s) considered**
The class just needed additional handling to traverse super class when Field not found in current class mapped

**Type**
Choose one: Bug fix

**Checklist**


Tarun-Bhardwaj:
@aditya-07 , could you please review this PR?

maimoonak:
@Tarun-Bhardwaj @aditya-07 I have fixed the failing CI issue

datacapture/src/main/java/com/google/android/fhir/datacapture/mapping/ResourceMapper.kt
```diff
@@ -607,7 +619,7 @@ private fun Class<*>.getFieldOrNull(name: String): Field? {
   return try {
     getDeclaredField(name)
   } catch (ex: NoSuchFieldException) {
-    return null
+    return superclass.getFieldOrNull(name)
```
aditya-07, line 622: superclass is nullable and may cause `NullPointerException` incase the field is not found throughout the class hierarchy.
```suggestion
    return superclass?.getFieldOrNull(name)
```

aditya-07 (changes requested):

maimoonak:
@aditya-07 feedback resolved.

datacapture/src/main/java/com/google/android/fhir/datacapture/mapping/ResourceMapper.kt
```diff
@@ -607,7 +619,7 @@ private fun Class<*>.getFieldOrNull(name: String): Field? {
   return try {
     getDeclaredField(name)
   } catch (ex: NoSuchFieldException) {
-    return null
+    return superclass?.getFieldOrNull(name)
```
epicadk, line 622:
```suggestion
    superclass?.getFieldOrNull(name)
```

aditya-07 (approved):
Great PR @maimoonak! LGTM!

jingtang10:
@maimoonak @aditya-07: Looking through this code, I understand that the change you made is making sure a latter value is not overriding an earlier value if you're setting the same field (which is a list).

So I'm not sure the title of this PR is accurate - can you confirm? and if so can you update the title and description?

thanks

maimoonak:
@jingtang10 The main aim of PR as to allow mapping fields into superclass. For example with 
`Patient->DomainResource->Resource`
code did not allow to map tag OR extension fields into superlcass to be mapped into SDC-extraction. This PR fixes it by [line](https://github.com/google/android-fhir/pull/954/files#diff-4d2985aba5e8203b97cc5dc2eac7dbbd0c7832be34a27036d1ecfb8e4a3a3e50R622). 
Other issue which was noticed when mapping extension was that incase of multiple mapping it was always overriding first value of extension hence only last mapped extension persisted. 

The issue that was fixed into PR was 'Allow super class field mapping for SDC Resource Extraction'

jingtang10:
thanks @maimoonak - as long as this is documented in this issue it's fine. if you can just list the two things this PR does in the description that'd be great. thanks!

jingtang10:
But perhaps a good follow-up PR to write is to provide minimal test cases. I think your new test case is too big and testing both of the things this PR is doing.

maimoonak:
Ok. Sure would break it down into two separate cases for each functionality separately

PR #1089, opened 2022-01-28
Fix focus for the date picker question

### LINKED ISSUE #1083: Datacapture, date picker doesn't open after cancel date picker dialog on same field
**Describe the bug**
When clicking on the date field like birthday, the date picker opens but when canceling dialog and clicking again then didn't open the date picker.  

**Component**
SDC library, Demo app

**To Reproduce**
1. Open Demo app.
2. Click on the plus icon.
3. Click on Date of Birth.
4. Click on the cancel button of the dialog.
5. Again click on Date of Birth.

**Expected behavior**
Open date picker dialog.

**Screenshots**
https://user-images.githubusercontent.com/98520274/151371327-b704bb6b-7053-460a-9af5-cfd54e7cef24.mp4

 - Device: Moto G(9)
 - Android version: 11
 - Build number: RPXS31.Q2-58-17-4-1
 - Version 1.0

**Additional context**

ktarasenko:

Fixes #1083

**Description**
Fix the date picker not opening after it was cancelled. 

**Alternative(s) considered**

**Type**
Bug fix

**Checklist**
- [ ] I have signed the Google [Individual CLA](https://cla.developers.google.com/about/google-individual), or I am covered by my company's [Corporate CLA](https://cla.developers.google.com/about/google-corporate )

### DIFF
--- a/datacapture/src/main/java/com/google/android/fhir/datacapture/views/QuestionnaireItemDatePickerViewHolderFactory.kt
+++ b/datacapture/src/main/java/com/google/android/fhir/datacapture/views/QuestionnaireItemDatePickerViewHolderFactory.kt
@@ -81,8 +81,6 @@ internal object QuestionnaireItemDatePickerViewHolderFactory :
                 QuestionnaireResponse.QuestionnaireResponseItemAnswerComponent().apply {
                   value = date
                 }
-              // Clear focus so that the user can refocus to open the dialog
-              textInputEditText.clearFocus()
               onAnswerChanged(textInputEditText.context)
             }
           )
@@ -92,7 +90,8 @@ internal object QuestionnaireItemDatePickerViewHolderFactory :
             .apply { arguments = bundleOf(REQUEST_BUNDLE_KEY_DATE to selectedDate) }
             .show(context.supportFragmentManager, DatePickerFragment.TAG)
           // Clear focus so that the user can refocus to open the dialog
-          textDateQuestion.clearFocus()
+          // TODO(ktarasenko) add a test case when UI tests are set up
+          textInputEditText.clearFocus()
         }
       }
 
--- a/datacapture/src/main/java/com/google/android/fhir/datacapture/views/QuestionnaireItemDatePickerViewHolderFactory.kt
+++ b/datacapture/src/main/java/com/google/android/fhir/datacapture/views/QuestionnaireItemDatePickerViewHolderFactory.kt
@@ -54,10 +54,8 @@ internal object QuestionnaireItemDatePickerViewHolderFactory :
         textInputEditText = itemView.findViewById(R.id.text_input_edit_text)
         // Disable direct text input to only allow input from the date picker dialog
         textInputEditText.keyListener = null
-        textInputEditText.setOnFocusChangeListener { _, hasFocus: Boolean ->
-          // Do not show the date picker dialog when losing focus.
-          if (!hasFocus) return@setOnFocusChangeListener
-
+        textInputEditText.setOnClickListener {
+          textInputEditText.requestFocus()
           // The application is wrapped in a ContextThemeWrapper in QuestionnaireFragment
           // and again in TextInputEditText during layout inflation. As a result, it is
           // necessary to access the base context twice to retrieve the application object
@@ -89,9 +87,6 @@ internal object QuestionnaireItemDatePickerViewHolderFactory :
           DatePickerFragment()
             .apply { arguments = bundleOf(REQUEST_BUNDLE_KEY_DATE to selectedDate) }
             .show(context.supportFragmentManager, DatePickerFragment.TAG)
-          // Clear focus so that the user can refocus to open the dialog
-          // TODO(ktarasenko) add a test case when UI tests are set up
-          textInputEditText.clearFocus()
         }
       }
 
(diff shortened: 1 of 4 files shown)


joiskash (commented):
Please look at video 1: This is with the change suggested in the PR, you will notice that on clicking the widget, the focus is immediately removed so you are unaware of which question is being answered cause the textInputEditText is not highlighted. 

https://user-images.githubusercontent.com/48797364/151954939-7bb2b798-210e-4175-8ed8-5636de06bb36.mp4

In video 2: Even though the date picker is open, the textInputEditText has focus. The case that is not handled right now is that we have not overridden the onCancel function in the DatePickerDialogFragment. So we never ending up calling setFragmentResult which leaves the focus on the textInputEditText.

https://user-images.githubusercontent.com/48797364/151955296-07f6b7b2-47e2-4c97-9859-43e82400a0c5.mp4

jingtang10:
@joiskash i can't really see the focus in the second video you posted. but yes i agree we should probably maintain focus.

this also means we probably shouldn't use the onfocus listener. because if we maintain focus, then if the user tap the text field again the listener is not going to be triggered and the dialog is not going to pop up. we have had this problem before.

joiskash:
> @joiskash i can't really see the focus in the second video you posted. but yes i agree we should probably maintain focus.
> 
> this also means we probably shouldn't use the onfocus listener. because if we maintain focus, then if the user tap the text field again the listener is not going to be triggered and the dialog is not going to pop up. we have had this problem before.

It is very subtle. But thats probably cause of dark mode. In the pic below you can see the grayish blue color on the left side whereas on the right its just gray. 

datefocus

The other option is to set the onCancel in DatePickerDialogFragment to call setFragmentResult with an empty bundle and then handle that case in the fragment listener. With something like

```
if(result.isEmpty){
    textInputEditText.clearFocus()
    return@listner
  }
```

If you select the date and press OK then onDateSet is called in DatePickerDialogFragment which internally calls setFragmentResult which clears the focus.

jingtang10:
thanks @joiskash. i get your point.

shall we try to use a method that doesn't involve on focus change listener?

joiskash:
Yes lets do that. Makes perfect sense!

ktarasenko:
I investigated switching to a ClickListener: it's not as straightforward as I would expect. 

If the EditText kept focusable - user needs to click twice to trigger the dialog, if not - the EditText doesn't get focus when dialog is open.  
Another quirk with focus here: when navigating with the keyboard (similar to what Accessibility tools would do), I need to move focus twice to get to the next element after date EditText. The focus moves to the beginning of the text and to the end of the text, just like in the editable EditText.

I see two possible solutions here: 
1) Apply the style of the EditText to the TextView
2) Introduce a custom view, that would handle TouchEvents and FocusEvents as expected 
Will take another look, which is more feasible

jingtang10:
> Thanks for the detailed report @joiskash! 
> 
> I investigated switching to a ClickListener: it's not as straightforward as I would expect. 
> 
> If the EditText kept focusable - user needs to click twice to trigger the dialog, if not - the EditText doesn't get focus when dialog is open.  
> Another quirk with focus here: when navigating with the keyboard (similar to what Accessibility tools would do), I need to move focus twice to get to the next element after date EditText. The focus moves to the beginning of the text and to the end of the text, just like in the editable EditText.
> 
> I see two possible solutions here: 
> 1) Apply the style of the EditText to the TextView
> 2) Introduce a custom view, that would handle TouchEvents and FocusEvents as expected 
> 
> Will take another look to see, which one is more feasible.

Might be worth checking if material design 3 widgets behave any better (@santosh-pingle was going to work on moving to material 3)

ktarasenko:
Would it be worth to fix the initial bug, but to introduce another, more minor focus issues?

joiskash:
> Would it be worth to fix the initial bug, but to introduce another, more minor focus issues?

I think it would, because the other issues are less important compared to this one. Mainly cause people using the SDC will have to be given instructions on how to avoid this.

ktarasenko:
Updated. The initial bug is fixed, and date-related fields are not focusable in touch mode now.
This means, that when the user clicks on the field, the focus isn't maintained behind the dialog.

shelaghm:
@jingtang10 Yes, that's exactly what I've been thinking as well. Some other thoughts/suggestions below

- Text entry should be the default when clicking the text box. 
- Adding helper text indicating the date format is going to be very important and helpful
- Calendar icon should be on the right side (this is the recommended pattern from groups that have done accessibility tests on this) so that the touch target area between text entry and date picker is greater.

Example below
Screen Shot 2022-02-18 at 10 19 40

jingtang10:
@ktarasenko the scope seems to be bigger than fixing the original focus problem. But can you see if this can be done within the same ticket/PR by setting the listener on the icon? Hoping that this is still a relatively small change. Many thanks

PallaviGanorkar:
With this fix, After launching app user need to double click textInputEditText to open Date Picker Dialog

jingtang10:
Can we use this: https://developer.android.com/reference/com/google/android/material/textfield/TextInputLayout#setEndIconOnClickListener(android.view.View.OnClickListener)

PallaviGanorkar:
@jingtang10  As per Above Screen shot,  helper text indicating the date format will be displayed from Questionnaire json file or given format is generic and and to read from string.xml

PR #1177, opened 2022-02-17
update flag so that keyboards starts with an upper case letter

### LINKED ISSUE #1176: Data Capture - soft keyboard should start with an upper case character for 'string' widgets
When I'm filling out a string field in a questionnaire, since the soft keyboard is set to lower case characters by default, I always have to change the keyboard to use uppercase letters first before I start typing. For instance, when I'm filling out the first name, I would like the first letter to be uppercase and then the rest of the name can be lowercase.

I would like that when I focus on a string field, the soft keyboard should be set to upper case by default. Then once I've typed in the first letter, it changes to lower case characters.

The alternative is doing this manually every time I encounter a string field but this gets tiresome for long questionnaires.

**Additional context**
I've only tried this on an Android 11 device, I'm not sure if this is also happening on previous versions of Android.

I don't know who can look into this but I would appreciate if someone looked into this as soon as possible.

epicadk:

Fixes #1176

**Description**

**Alternative(s) considered**
The other options are we CAP_WORDS and CAP_CHARACTERS.

I think we can have CAP_WORDS for singleline and CAP_SENTENCES for Multiline : /.

**Type**
Choose one: Feature 

**Checklist**
- [x] I have read the [Contributing](https://github.com/google/android-fhir/wiki/Contributing) page


epicadk:
@jingtang10 apologies forgot to or the input class

PR #1185, opened 2022-02-21
Change areItemsTheSame to check if internal objects are the same with ===

### LINKED ISSUE #765: Reselecting the single answer option initially set in questionnaire.item.initial disables nested answer options at the second level
**Describe the bug**
When using a questionnaire with three levels of nesting, if the top most question has a single answer which is already selected, then re-selecting that answer disables the third level of nesting permanently.

**Component**SDC library

**To Reproduce**
1. Use the questionnaire that is in this issue
2. Select the first answer in the drop down for the question number 1
3. You will see that on selecting the Ward i.e. the next question, the third question does not get displayed

**Expected behavior**
The third question must be enabled and displayed based on the answer of the second question.

**Screenshots**
singleAnswerBug

 -  Pixel4a emulator
 - Android version: 10
 - 29

Yes

joiskash:

Fixes #765 

**Description**
If you have a questionnaire with multiple levels of nesting, questionnaire response items are nested within answers. When the parent answer is changed, we clear the questionnaireResponseItem.answer list and add the new answer to it. The nested questionnaireResponseItems also get cleared and new instances of them are created in the questionnaireResponseItemChangedCallback. Since questionnaireItemViewItem.questionnaireResponseItem was a value , the ViewHolderFactory would always hold the reference to the old nested questionnaireResponseItem. On making it a variable, the viewHolderFactory now gets the reference of the correct questionnaireResponseitem while setting the answer.  This allows enable when to work properly and thus displays the third nested answer too.

**Alternative(s) considered**
No

**Type**
Choose one: Bug fix 

**Checklist**
- [x] I have read the [Contributing](https://github.com/google/android-fhir/wiki/Contributing) page


jingtang10:
Thanks Kashyap for fixing this. First thought is that we should write some tests to cover this case. Upon second thought I'm not entirely sure why this would be the fix anyway - it seems like if `val` vs `var` can fix this issue there's probably an underlying problem which is the items are not being replaced/deleted/cached correctly.

joiskash:
> Thanks Kashyap for fixing this. First thought is that we should write some tests to cover this case. Upon second thought I'm not entirely sure why this would be the fix anyway - it seems like if `val` vs `var` can fix this issue there's probably an underlying problem which is the items are not being replaced/deleted/cached correctly.

Sure , ill write a UT first.

datacapture/src/main/java/com/google/android/fhir/datacapture/views/QuestionnaireItemViewItem.kt
```diff
@@ -107,4 +107,22 @@ data class QuestionnaireItemViewItem(
...
+   * [questionnaireResponseItem] that are bound to the view. Whenever there is a change made to the
+   * [questionnaireResponseItem] via UI, a new list of QuestionnaireItemViewItems is created that
+   * holds a reference to the updated [questionnaireResponseItem]. It is important to note that
+   * [questionnaireResponseItem]s are nested in the parent
+   * [QuestionnaireResponse.QuestionnaireResponseItemAnswerComponent], so when a parent answer is
+   * changed, the nested items are cleared and need to be answered again. This new updated list with
+   * references to new nested [questionnaireResponseItem]s is passed to the adapter, which goes onto
+   * perform a Diff before updating the view. While checking if two [QuestionnaireItemViewItem]s are
+   * equal or not, we need to check if they are referencing the same objects.
+   */
```
jingtang10, line 122:
```suggestion
  /**
   * [QuestionnaireItemViewItem] is a transient object for the UI only. Whenever the user makes any change
   * via the UI, a new list of [QuestionnaireItemViewItem]s will be created, each holding references to the
   * underlying [QuestionnaireItem] and [QuestionnaireResponseItem]. To avoid refreshing the UI 
   * unnecessarily with the same [QuestionnaireItem]s and [QuestionnaireResponseItem]s,  we consider two
   * [QuestionnaireItemViewItem]s to be the same if they have the same underlying [QuestionnaireItem] and
   * [QuestionnaireResponseItem]. See [QuestionnaireItemAdapter.DiffCallback].
   * 
   * On the other hand, under certain circumstances, the underlying [QuestionnaireResponseItem] might be 
   * recreated for the same question. For example, if a [QuestionnaireItem] is nested under another 
   * [QuestionnaireItem], the [QuestionnaireResponseItem](s) will be nested under the parent 
   * [QuestionnaireResponse.QuestionnaireResponseItemAnswerComponent]. and if the 
   * [QuestionnaireResponse.QuestionnaireResponseItemAnswerComponent] is changed, the nested 
   * [QuestionnaireResponseItem] will be recreated, too. In such cases, it would be incorrect
   * to simply check that the `linkId` of the underlying [QuestionnaireItem] and 
   * [QuestionnaireResponseItem] match.
   */
```

datacapture/src/main/java/com/google/android/fhir/datacapture/views/QuestionnaireItemViewItem.kt
```diff
@@ -107,4 +107,22 @@ data class QuestionnaireItemViewItem(
...
+  override fun equals(other: Any?): Boolean {
+    return this.questionnaireItem === (other as QuestionnaireItemViewItem).questionnaireItem &&
+      this.questionnaireResponseItem ===
+        (other as QuestionnaireItemViewItem).questionnaireResponseItem
+  }
```
jingtang10, line 127: add another function `equalsDeep` to handle the `areContentsTheSame` functino of the `DiffCallback`.

jingtang10 (changes requested):

datacapture/src/main/java/com/google/android/fhir/datacapture/views/QuestionnaireItemViewItem.kt
```diff
@@ -107,4 +107,22 @@ data class QuestionnaireItemViewItem(
...
+   * [questionnaireResponseItem] via UI, a new list of QuestionnaireItemViewItems is created that
+   * holds a reference to the updated [questionnaireResponseItem]. It is important to note that
+   * [questionnaireResponseItem]s are nested in the parent
+   * [QuestionnaireResponse.QuestionnaireResponseItemAnswerComponent], so when a parent answer is
+   * changed, the nested items are cleared and need to be answered again. This new updated list with
+   * references to new nested [questionnaireResponseItem]s is passed to the adapter, which goes onto
+   * perform a Diff before updating the view. While checking if two [QuestionnaireItemViewItem]s are
+   * equal or not, we need to check if they are referencing the same objects.
+   */
+  override fun equals(other: Any?): Boolean {
```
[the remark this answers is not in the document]
joiskash: done

datacapture/src/main/java/com/google/android/fhir/datacapture/views/QuestionnaireItemViewItem.kt
joiskash: done

datacapture/src/main/java/com/google/android/fhir/datacapture/views/QuestionnaireItemViewItem.kt
```diff
@@ -107,4 +107,43 @@ data class QuestionnaireItemViewItem(
...
+   * might be recreated for the same question. For example, if a [QuestionnaireItem] is nested under
+   * another [QuestionnaireItem], the [QuestionnaireResponseItem](s) will be nested under the parent
+   * [QuestionnaireResponse.QuestionnaireResponseItemAnswerComponent]. and if the
+   * [QuestionnaireResponse.QuestionnaireResponseItemAnswerComponent] is changed, the nested
+   * [QuestionnaireResponseItem] will be recreated, too. In such cases, it would be incorrect to
+   * simply check that the `linkId` of the underlying [QuestionnaireItem] and
+   * [QuestionnaireResponseItem] match.
+   */
+  override fun equals(other: Any?): Boolean {
+    return if (other is QuestionnaireItemViewItem) {
```
jingtang10, line 130: turn this if condition around and return early.

datacapture/src/main/java/com/google/android/fhir/datacapture/views/QuestionnaireItemViewItem.kt
```diff
@@ -107,4 +107,43 @@ data class QuestionnaireItemViewItem(
...
+   * references to the underlying [QuestionnaireItem] and [QuestionnaireResponseItem]. To avoid
+   * refreshing the UI unnecessarily with the same [QuestionnaireItem]s and
+   * [QuestionnaireResponseItem]s, we consider two [QuestionnaireItemViewItem]s to be the same if
+   * they have the same underlying [QuestionnaireItem] and [QuestionnaireResponseItem]. See
+   * [QuestionnaireItemAdapter.DiffCallback].
+   *
+   * On the other hand, under certain circumstances, the underlying [QuestionnaireResponseItem]
+   * might be recreated for the same question. For example, if a [QuestionnaireItem] is nested under
+   * another [QuestionnaireItem], the [QuestionnaireResponseItem](s) will be nested under the parent
+   * [QuestionnaireResponse.QuestionnaireResponseItemAnswerComponent]. and if the
```
jingtang10, line 123:
```suggestion
   * [QuestionnaireResponse.QuestionnaireResponseItemAnswerComponent], and if the
```

datacapture/src/main/java/com/google/android/fhir/datacapture/views/QuestionnaireItemViewItem.kt
```diff
@@ -107,4 +107,43 @@ data class QuestionnaireItemViewItem(
...
+      false
+    }
+  }
+
+  /**
+   * Comparing the contents of two [QuestionnaireItemViewItem]s,involves traversing the underlying
+   * [Questionnaire.QuestionnaireItemComponent] and
+   * [QuestionnaireResponse.QuestionnaireResponseItemComponent],and comparing values of all the
+   * properties. This is done by using the [Questionnaire.QuestionnaireItemComponent.equalsDeep] and
+   * [QuestionnaireResponse.QuestionnaireResponseItemComponent.equalsDeep] functions
```
jingtang10, line 143:
```suggestion
   * Comparing the contents of two [QuestionnaireItemViewItem]s by traversing the underlying
   * [Questionnaire.QuestionnaireItemComponent] and
   * [QuestionnaireResponse.QuestionnaireResponseItemComponent] and comparing values of all the
   * properties. This is done by using [Questionnaire.QuestionnaireItemComponent.equalsDeep] and
   * [QuestionnaireResponse.QuestionnaireResponseItemComponent.equalsDeep].
```

jingtang10 (approved):
Thanks @joiskash again for debugging fixing this very thorning issue!

Just a few nits! Please feel free to merge once these are addressed.

datacapture/src/main/java/com/google/android/fhir/datacapture/views/QuestionnaireItemViewItem.kt
joiskash: done

datacapture/src/main/java/com/google/android/fhir/datacapture/views/QuestionnaireItemViewItem.kt
joiskash: done

datacapture/src/main/java/com/google/android/fhir/datacapture/views/QuestionnaireItemViewItem.kt
joiskash: done

datacapture/src/main/java/com/google/android/fhir/datacapture/views/QuestionnaireItemViewItem.kt
```diff
@@ -107,4 +107,43 @@ data class QuestionnaireItemViewItem(
...
+   * simply check that the `linkId` of the underlying [QuestionnaireItem] and
+   * [QuestionnaireResponseItem] match.
+   */
+  override fun equals(other: Any?): Boolean {
+    return if (other !is QuestionnaireItemViewItem) {
+      false
+    } else {
+      this.questionnaireItem === other.questionnaireItem &&
+        this.questionnaireResponseItem === other.questionnaireResponseItem
+    }
```
jingtang10, line 135:
```suggestion
    if (other !is QuestionnaireItemViewItem) return false
    return this.questionnaireItem === other.questionnaireItem &&
        this.questionnaireResponseItem === other.questionnaireResponseItem
```

jingtang10 (changes requested):

jingtang10 (approved):

PR #1207, opened 2022-02-24
Replace "reference" with "demo" in PR Template

omarismail94:
Nit fix. Since folder is called `demo` now, replaced mention of `reference` with `demo`


vvikramraj (approved):
Now that you have triggered my OCD, can you please also add full stops to these sentences? :)

omarismail94:
@vvikramraj fixed!

PR #1235, opened 2022-03-15
Use reified generics for delete and read operations in FhirEngine

### LINKED ISSUE #784: Design and implementation for generation of UUID by FHIR Engine
How do we handle resource IDs by the FHIR SDK. This has implications for how we can do the Sync from the Android client to the FHIR Server and also resolve potential ID conflict issues

See [this deck](https://docs.google.com/presentation/d/1xx1mt8wjktVanrjQk3YQe0jmfyigTcQLM24kxv6v79A/edit#slide=id.gf01cad751a_0_5) for exploration of the different approaches

Use PUT to create new resources on the server side. The SDK does not enforce (but strongly recommends) UUID creation (option 3)

* The SDK should be able to specify UUIDs for resources created by the client application 
* Resources created on the SDK side should be able to be sync'd to the FHIR server using PUT (not POST)
* The FHIR SDK will not enforce, but strongly recommend UUID creation - **default behaviour will be to create the UUIDs providing implementers the ability to override**

See this deck for detailed alternatives. In summary these include:
1. Server sets the resource IDs on POSTing the new resources
2. FHIR SDK enforces the creation of UUIDs

**Additional context**
1. UUIDs will only be generated for newly created resources.
2. This conforms to the standard according to the [FHIR spec](https://www.hl7.org/fhir/http.html#upsert). The FHIR SDK is effectively an extension of the FHIR server 
3. There will be server side requirements that will need to be enabled to support this. This will be documented as part of an ImplementationGuide

jingtang10:

Fixes #784 

**Description**
Use reified generics in the delete and read APIs in FhirEngine.

For example,

```
fhirEngine.load(Patient::class.java, patientId)
```

becomes

```
fhirEngine.get<Patient>(patientId)
```

and

```
fhirEngine.remove(Patient::class.java, patientId)
```

becomes

```
fhirEngine.delete<Patient>(patientId)
```

The implementation uses `inline` functions with `reified` generics, and we use the `@Deprecated` annotation to suggest replacement if the caller accidentally calls the internal APIs.

**Alternative(s) considered**
NA

**Type**
Choose one: Feature

NA

**Checklist**

### DIFF
--- a/engine/src/main/java/com/google/android/fhir/FhirEngine.kt
+++ b/engine/src/main/java/com/google/android/fhir/FhirEngine.kt
@@ -79,7 +79,6 @@ interface FhirEngine {
    */
   @Deprecated("Deprecated", ReplaceWith("this.get<R>(id)"))
   suspend fun <R : Resource> load(clazz: Class<R>, id: String): R
-  
 
   /**
    * Removes a FHIR resource given the class and the logical ID.
--- a/engine/src/main/java/com/google/android/fhir/FhirEngine.kt
+++ b/engine/src/main/java/com/google/android/fhir/FhirEngine.kt
@@ -73,21 +73,19 @@ interface FhirEngine {
   suspend fun getLastSyncTimeStamp(): OffsetDateTime?
 
   /**
-   * DO NOT USE. Internal and test only function to load a FHIR resource given the class and the
-   * logical ID.
+   * Loads a FHIR resource given the class and the logical ID.
    *
-   * The `@Deprecated` annotation suggests replacement if this function is called.
+   * DEPRECATED. Use `get<R>(id)` instead.
    */
-  @Deprecated("Internal and test only function", ReplaceWith("this.get<R>(id)"))
+  @Deprecated("Deprecated", ReplaceWith("this.get<R>(id)"))
   suspend fun <R : Resource> load(clazz: Class<R>, id: String): R
 
   /**
-   * DO NOT USE. Internal and test only function to delete a FHIR resource given the class and the
-   * logical ID.
+   * Removes a FHIR resource given the class and the logical ID.
    *
-   * The `@Deprecated` annotation suggests replacement if this function is called.
+   * DEPRECATED. Use `delete<R>(id)` instead.
    */
-  @Deprecated("Internal and test only function", ReplaceWith("this.delete<R>(id)"))
+  @Deprecated("Deprecated", ReplaceWith("this.delete<R>(id)"))
   suspend fun <R : Resource> remove(clazz: Class<R>, id: String)
 }
 
--- a/engine/src/main/java/com/google/android/fhir/FhirEngine.kt
+++ b/engine/src/main/java/com/google/android/fhir/FhirEngine.kt
@@ -43,22 +43,6 @@ interface FhirEngine {
   suspend fun <R : Resource> update(resource: R)
 
   /**
-   * Returns a FHIR resource of type [clazz] with [id] from the local storage.
-   *
-   * @param <R> The resource type which should be a subtype of [Resource].
-   * @throws ResourceNotFoundException if the resource is not found
-   */
-  @Throws(ResourceNotFoundException::class)
-  suspend fun <R : Resource> load(clazz: Class<R>, id: String): R
-
-  /**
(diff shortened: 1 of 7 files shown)


omarismail94 (commented):
Can `getInternal` and `deleteInternal` not also be turned into an inline reified function?

This way we can move from doing`database.delete(clazz, id)` and do `database.delete<R>(id)`. Not sure if this is possible (or if it makes sense, but just an observation I had)

engine/src/main/java/com/google/android/fhir/FhirEngine.kt
```diff
@@ -42,22 +42,6 @@ interface FhirEngine {
...
   suspend fun <R : Resource> update(resource: R)
 
-  /**
-   * Returns a FHIR resource of type [clazz] with [id] from the local storage.
-   *
-   * @param <R> The resource type which should be a subtype of [Resource].
-   * @throws ResourceNotFoundException if the resource is not found
-   */
-  @Throws(ResourceNotFoundException::class)
-  suspend fun <R : Resource> load(clazz: Class<R>, id: String): R
```
stevenckngaa, line 52: Wouldn't this break any clients that have already been using `load` & `remove`? Is there any harm of using `load` & `remove` which take clazz as a param? The extension functions are syntax sugar in my opinion. Could we just keep the existing load and remove as overloaded deprecated functions.

jingtang10:
> Can `getInternal` and `deleteInternal` not also be turned into an inline reified function?
> 
> This way we can move from doing`database.delete(clazz, id)` and do `database.delete<R>(id)`. Not sure if this is possible (or if it makes sense, but just an observation I had)

It's not possible to make `getInternal` and `deleteInternal` inline because they are virtual functions. I don't really have a perfect solution in this case because we do want to keep the interface but in the meantime the reified functions cannot be virtual so we have to use this extension function "trick". The most important thing to me in this change is to create APIs that are dead easy for developers to use and that's why I went with this.

engine/src/main/java/com/google/android/fhir/FhirEngine.kt
jingtang10: It would yes. Looking in https://github.com/opensrp/fhircore I'm not seeing it getting or deleting resources using these APIs, I think any access is done through the search API. @f-odhiambo @ekigamba @maimoonak to confirm.

I don't think with the alpha releases we are making any promises about backward compatibility. But we can definitely keep the old names if it's important.

stevenckngaa: Having internal methods in interface is an anti-pattern. If there is no way to hide those, let's do remove internal from the naming. There should not be any harm of callers using functions that take clazz as params.

engine/src/main/java/com/google/android/fhir/FhirEngine.kt
```diff
@@ -87,6 +71,44 @@ interface FhirEngine {
...
   /** Returns the timestamp when data was last synchronized. */
   suspend fun getLastSyncTimeStamp(): OffsetDateTime?
+
+  /**
+   * DO NOT USE. Internal and test only function to load a FHIR resource given the class and the
+   * logical ID.
+   *
+   * The `@Deprecated` annotation suggests replacement if this function is called.
+   */
+  @Deprecated("Internal and test only function", ReplaceWith("this.get<R>(id)"))
```
stevenckngaa, line 81: No longer internal. Same for line 90.

stevenckngaa (approved):

engine/src/main/java/com/google/android/fhir/FhirEngine.kt
ekigamba: We use both `load` and `remove` a couple of times. Some of the usages are here https://github.com/opensrp/fhircore/blob/70066010df18d31c436cedbf0d2c8b5cda2eb9f4/android/engine/src/main/java/org/smartregister/fhircore/engine/data/local/DefaultRepository.kt#L111 and https://github.com/opensrp/fhircore/blob/70066010df18d31c436cedbf0d2c8b5cda2eb9f4/android/engine/src/main/java/org/smartregister/fhircore/engine/data/local/DefaultRepository.kt#L104

PR #1263, opened 2022-03-26
Optimize CI

omarismail94:
Simplify CI cache:

1. By caching the AVD, we avoid creating a new AVD for each of the instrumentation test steps. An example of this is  [here](https://github.com/ReactiveCircus/android-emulator-runner#usage)
2. Using Gradle Build Action avoids the need to use the `cache` step ourselves. See [here](https://github.com/gradle/gradle-build-action#why-use-the-gradle-build-action) for details


jingtang10:
adding @florina-muntenescu who authored the caching mechanism in case she has anything to add. Thanks @florina-muntenescu and @omarismail94.

jingtang10:
@omarismail94 can you please elaborate the intended effects of this optimization?

.github/workflows/ci.yml
```diff
@@ -99,6 +78,27 @@ jobs:
...
+      - name: AVD cache
+        uses: actions/cache@v2
+        id: avd-cache
+        with:
+          path: |
+            ~/.android/avd/*
+            ~/.android/adb*
+          key: avd-${{ matrix.api-level }}
+
+      - name: create AVD and generate snapshot for caching
```
jingtang10, line 90:
```suggestion
      - name: Create AVD and generate snapshot for caching if cache not hit
```

jingtang10 (approved):

PR #1326, opened 2022-04-23
Workflow Library: Implementing Library.evaluate and first androidTest suite

vitorpamplona:
**Description**

Implemented FhirOperator.evaluateLibrary function
Connected the LibraryProcessor to the FhirEngine's local DB, matching evaluateMeasure
Generalized FhirEngineRetrieveProvider to work with any FHIR Model
Added the first androidTests with a simple COVID Immunization check

Fixes #[issue number]

**Alternative(s) considered**

**Type**
Choose one: Feature

**Checklist**


workflow/src/androidTest/java/com/google/android/fhir/workflow/FhirOperatorLibraryEvaluateTest.kt
```diff
@@ -0,0 +1,82 @@
...
+import org.hl7.fhir.r4.model.Parameters
+import org.hl7.fhir.r4.model.Resource
+import org.hl7.fhir.r4.model.ResourceType
+import org.junit.Assert.assertEquals
+import org.junit.Before
+import org.junit.Test
+import org.junit.runner.RunWith
+
+@RunWith(AndroidJUnit4::class)
+class FhirOperatorLibraryEvaluateTest {
```
ktarasenko, line 36: Please move this test to /test root, since it's a robolectric test, not the one that runs on the instrumented device

workflow/src/androidTest/java/com/google/android/fhir/workflow/FhirOperatorLibraryEvaluateTest.kt
```diff
@@ -0,0 +1,82 @@
...
+  fun open(assetName: String): InputStream? {
+    return javaClass.classLoader?.getResourceAsStream(assetName)
+  }
+
+  @Before
+  fun setUp() = runBlocking {
+    val ids =
+      fhirEngine.create(json.parseResource(open("COVIDImmunizationHistory.json")) as Resource)
+    println(ids)
+    println("Composition " + fhirEngine.get(ResourceType.Composition, ids.first()))
```
ktarasenko, line 54: could it be a part of the test, instead of logs?

workflow/testdata/COVIDCheck.cql
```diff
@@ -0,0 +1,35 @@
+library DDCCPass version '0.0.1'
```
ktarasenko, line 1: Is cql file is here for the reference?  Some other files aren't seems to be used too

workflow/src/androidTest/java/com/google/android/fhir/workflow/FhirOperatorLibraryEvaluateTest.kt
```diff
@@ -0,0 +1,82 @@
...
+          "CompletedImmunization",
+          "GetFinalDose",
+          "GetSingleDose",
+          "ModernaProtocol",
+          "PfizerProtocol"
+        )
+      ) as
+        Parameters
+
+    assertEquals(true, results.getParameterBool("CompletedImmunization"))
```
ktarasenko, line 78: please use [Truth library](https://truth.dev/) for assertions

workflow/src/main/java/com/google/android/fhir/workflow/FhirEngineRetrieveProvider.kt
```diff
@@ -47,112 +42,35 @@ class FhirEngineRetrieveProvider(val fhirEngine: FhirEngine) : TerminologyAwareR
...
+          context is String &&
+          contextValue is String
+      ) {
+        val search = Search(ResourceType.fromCode(dataType))
+        search.filter(ReferenceClientParam(contextPath), { value = "$context/$contextValue" })
+        fhirEngine.search<Resource>(search).toMutableList()
+      } else {
+        val search = Search(ResourceType.fromCode(dataType))
+        if (hasField(dataType, "active")) {
+          // TODO: I am not sure why the default search is only for active entities
```
ktarasenko, line 60: This assumption (filter on active patients) didn't work. I removed it in the other PR that isn't merged yet.

workflow/src/androidTest/java/com/google/android/fhir/workflow/FhirOperatorLibraryEvaluateTest.kt
```diff
@@ -0,0 +1,82 @@
...
+
+  private val fhirEngine =
+    FhirEngineProvider.getInstance(ApplicationProvider.getApplicationContext())
+  private val fhirContext = FhirContext.forCached(FhirVersionEnum.R4)
+  private val fhirOperator = FhirOperator(fhirContext, fhirEngine)
+
+  private val json = fhirContext.newJsonParser()
+
+  fun open(assetName: String): InputStream? {
+    return javaClass.classLoader?.getResourceAsStream(assetName)
```
ktarasenko, line 46: could be just `javaClass..getResourceAsStream(assetName)`

workflow/src/main/java/com/google/android/fhir/workflow/FhirEngineRetrieveProvider.kt
```diff
@@ -47,112 +42,35 @@ class FhirEngineRetrieveProvider(val fhirEngine: FhirEngine) : TerminologyAwareR
...
-              fhirEngine.search<CarePlan> { filter(Patient.ACTIVE, { value = of(true) }) }
-            careplan.toMutableList()
-          }
-        }
-        else -> {
-          throw NotImplementedError("Not implemented yet")
+      if (contextPath == "id" && contextValue is String) {
+        mutableListOf(fhirEngine.get(ResourceType.fromCode(dataType), contextValue))
+      } else if (contextPath is String &&
+          contextPath != "id" &&
```
ktarasenko, line 50: contextPath != "id" is redundant

workflow/src/main/java/com/google/android/fhir/workflow/FhirEngineRetrieveProvider.kt
```diff
@@ -47,112 +42,35 @@ class FhirEngineRetrieveProvider(val fhirEngine: FhirEngine) : TerminologyAwareR
...
+        fhirEngine.search<Resource>(search).toMutableList()
       }
     }
   }
+  fun hasField(dataType: String?, field: String): Boolean {
+    if (dataType == null) return false
+    return try {
+      Class.forName("org.hl7.fhir.r4.model.$dataType").getDeclaredField(field)
+      true
+    } catch (e: NoSuchFieldException) {
```
ktarasenko, line 72: If we still need this method, would it be possible to check if the field exists 
`Class.forName("").declaredFields.contains(field)` instead of catching an exception

workflow/testdata/COVIDCheck.cql
ktarasenko: Could you please move the test data to a subfolder? It's slowly getting crowded

workflow/src/main/java/com/google/android/fhir/workflow/FhirEngineRetrieveProvider.kt
```diff
@@ -47,112 +42,35 @@ class FhirEngineRetrieveProvider(val fhirEngine: FhirEngine) : TerminologyAwareR
...
-            careplan.toMutableList()
-          } else {
-            val careplan =
-              fhirEngine.search<CarePlan> { filter(Patient.ACTIVE, { value = of(true) }) }
-            careplan.toMutableList()
-          }
-        }
-        else -> {
-          throw NotImplementedError("Not implemented yet")
+      if (contextPath == "id" && contextValue is String) {
```
ktarasenko, line 47: is there a constant for "id" somewhere?

ktarasenko (changes requested):
Amazing, thanks for adding this!

workflow/src/androidTest/java/com/google/android/fhir/workflow/FhirOperatorLibraryEvaluateTest.kt
vitorpamplona: My idea was to keep CQL tests running on the instrumented device to catch runtime exceptions of the environment (e.g. when XML-based or CQL/text libs are loaded and crash on Android only). Does it make sense?

workflow/src/main/java/com/google/android/fhir/workflow/FhirEngineRetrieveProvider.kt
vitorpamplona: We could use `fhirEngine.search` for everything instead of singling out the "id" request to the `fhirEngine.get`. I assume there is some performance benefit of using `get` instead of `search`, but if not, using search could remove the need to specify `id`

workflow/src/main/java/com/google/android/fhir/workflow/FhirEngineRetrieveProvider.kt
vitorpamplona: Can you point me out to the PR? I can incorporate those changes here.

workflow/src/androidTest/java/com/google/android/fhir/workflow/FhirOperatorLibraryEvaluateTest.kt
ktarasenko: Got it. Yeah, makes sense.

workflow/src/main/java/com/google/android/fhir/workflow/FhirEngineRetrieveProvider.kt
ktarasenko: keep if for now, I'll merge the change afterward

workflow/src/main/java/com/google/android/fhir/workflow/FhirEngineRetrieveProvider.kt
ktarasenko: I think keeping `get` is fine, i meant, do you know, is there a constant for the "id" string somewhere in `org.hl7.fhir.r4.model`?

ktarasenko (approved):

workflow/src/main/java/com/google/android/fhir/workflow/FhirEngineRetrieveProvider.kt
```diff
@@ -48,111 +43,29 @@ class FhirEngineRetrieveProvider(val fhirEngine: FhirEngine) : TerminologyAwareR
...
+        mutableListOf(fhirEngine.get(ResourceType.fromCode(dataType), contextValue))
+      } else if (contextPath is String && context is String && contextValue is String) {
+        val search = Search(ResourceType.fromCode(dataType))
+        search.filter(ReferenceClientParam(contextPath), { value = "$context/$contextValue" })
+        fhirEngine.search<Resource>(search).toMutableList()
+      } else {
+        val search = Search(ResourceType.fromCode(dataType))
+        if (hasField(dataType, "active")) {
+          // TODO: I am not sure why the default search is only for active entities
```
f-odhiambo, line 55: @vitorpamplona We had this check to basically filter records that are matched to active patients thus we do not run/evaluate this for inactive patients but this may not be true for all use cases cc @ktarasenko

vitorpamplona: I believe CQL rules should filter only active records themselves as opposed to the engine forcing all libraries to run only on active records.

workflow/src/main/java/com/google/android/fhir/workflow/FhirEngineRetrieveProvider.kt
vitorpamplona: There's no constant definition. The String comes from [here]( https://github.com/DBCG/cql_engine/blob/27d1321b92c83beb618b85181324b1c16a714b7a/engine.fhir/src/main/java/org/opencds/cqf/cql/engine/fhir/model/FhirModelResolver.java#L98)

workflow/src/androidTest/java/com/google/android/fhir/workflow/FhirOperatorLibraryEvaluateTest.kt
jingtang10: are we sure these things can be caught?

workflow/src/main/java/com/google/android/fhir/workflow/FhirOperator.kt
```diff
@@ -160,6 +177,40 @@ class FhirOperator(fhirContext: FhirContext, fhirEngine: FhirEngine) {
...
+      loadLib(entry.resource as Library)
+    }
+  }
+
+  /**
+   * The function evaluates a FHIR library by Id and returns a Parameters resource that contains the
+   * evaluation result Parameters: id - the Id of the Library to evaluate patientId - the patient Id
+   * to use for evaluation, if applicable expressions - names of expressions in the Library to
+   * evaluate. if omitted all expressions are evaluated. Returns: IBaseParameters
+   */
```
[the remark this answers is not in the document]
jingtang10: no need to include @param or @return if there's nothing interesting to say about the param or the return value

workflow/src/main/java/com/google/android/fhir/workflow/FhirEngineRetrieveProvider.kt
```diff
@@ -48,111 +43,29 @@ class FhirEngineRetrieveProvider(val fhirEngine: FhirEngine) : TerminologyAwareR
...
+          search.filter(TokenClientParam("active"), { value = of(true) })
         }
+        fhirEngine.search<Resource>(search).toMutableList()
       }
     }
   }
+  fun hasField(dataType: String?, field: String): Boolean {
```
jingtang10, line 62: add empty line before new function

workflow/testdata/covid-check/COVIDCheck.cql
```diff
@@ -0,0 +1,35 @@
+library DDCCPass version '0.0.1'
```
jingtang10, line 1: is this file used anywhere? i couldn't quite work out.

workflow/src/androidTest/java/com/google/android/fhir/workflow/FhirOperatorLibraryEvaluateTest.kt
```diff
@@ -0,0 +1,87 @@
...
+  }
+
+  @Before
+  fun setUp() = runBlocking {
+    fhirEngine = FhirEngineProvider.getInstance(ApplicationProvider.getApplicationContext())
+    fhirOperator = FhirOperator(fhirContext, fhirEngine)
+  }
+
+  @Test
+  fun evaluateCOVIDCheck() = runBlocking {
```
jingtang10, line 61: I would really love to see our test cases to be minimal. If for some reason, this test starts to fail, I will have no idea how to debug the cql libs. is there a way to simplify the test case so that it's basically testing the code path, and checking that the library is getting evaluated, without having complex logic such as covid?

from the point of view of the codebase, if the library does a complex calculation or if the library returns a simple true or false value doesn't make any diff right?

jingtang10 (changes requested):

workflow/testdata/covid-check/COVIDCheck.cql
vitorpamplona: This is kept just for reference since it has to be compiled into a Base64 representation and put into a library JSON file. The long-term goal is to compile the CQL before running these tests directly, removing the need to keep intermediary ELM representations.

vitorpamplona: I deleted this file for now. It might come back when we are ready to compile CQL into Libraries directly on the SDK.

workflow/src/main/java/com/google/android/fhir/workflow/FhirOperator.kt
```diff
@@ -160,6 +177,40 @@ class FhirOperator(fhirContext: FhirContext, fhirEngine: FhirEngine) {
...
+    for (entry in libBundle.entry) {
+      loadLib(entry.resource as Library)
+    }
+  }
+
+  /**
+   * Returns the evaluation result of a FHIR library
+   * @param libraryUrl ...
+   * @param patientId the Id of the patient to be evaluated
+   * @param expressions ...... if omitted all expressions are evaluated
```
jingtang10, line 190: please update the `...` to actual words, or delete otherwise

workflow/src/androidTest/java/com/google/android/fhir/workflow/FhirOperatorLibraryEvaluateTest.kt
jingtang10: good to know. thanks

jingtang10 (approved):
thanks @vitorpamplona - around tests what i'm really concerned about is just that if the tests start failing, will someone be able to step in and debug the test. if the tests are super complex with a lot of artifacts, it'll be super hard. that's why i'm always pushing for minimum tests. thanks for making that change. 

additionally, can you add a little more comment on where the cql lib came from and how that's generated etc in the code comment? i know you can't add comment in json, so adding explanatino in the test file is good enough.

otherwise looking good. just have another small nit.

happy to merge if these are addressed.

workflow/src/main/java/com/google/android/fhir/workflow/FhirOperator.kt
vitorpamplona: Fixed.

vitorpamplona:
> around tests what i'm really concerned about is just that if the tests start failing, will someone be able to step in and debug the test. if the tests are super complex with a lot of artifacts, it'll be super hard. that's why i'm always pushing for minimum tests. thanks for making that change.

I completely agree. Test structures for most of these CQL evaluation tasks are quite complex. I added PR #1395 to help understand this Test Case. It goes through the steps of Compiling CQL, generating the FHIR Library and making the bundle, which are necessary to call evaluate functions. As you can probably see, these are not really Unit tests. It's more of an integration test than anything else. 

I think moving forward, when the CQL compilation becomes available in the SDK, we can redesign these tests to go from reading CQL directly and running on it - masking all the intermediary processes and hopefully facilitating the understanding of what's happening. 

Once CQL compilation is available, we can also start breaking tests down to individual evaluation features. Right now it would be too much complexity and duplicated test resources to handle.

PR #1333, opened 2022-04-25
Cleanup FhirEngineProvider for tests

### LINKED ISSUE #1317: provide testing APIs
`FhirEngineProvider` is currently a singleton, which is not great for tests as it leaks the engine (and the database) between tests (also to the disk so even between test runs).

This bug is to track what to do to better isolate our tests:
* `FhirEngineProvider`'s init method (or the config object) should have a "TestMode" similar to the database:
https://github.com/google/android-fhir/blob/9cffffc28532c77461db3615ca1f57f5c87f6d77/engine/src/main/java/com/google/android/fhir/FhirServices.kt#L44
In test mode, FhirEngineProvider should create the FhirService in memory mode.

* `FhirEngineProvider` should have a `cleanup` method to clear the state (`fhirEngineConfiguration` and `fhirServices`). We will need to change these to `nullable` instead of `lateinit`. Calling `cleanup` outside test mode should should throw an exception.

* We should create a `@TestRule / TestWatcher` that we can add into test classes which will take care of init-deinit for FhirEngineProvider. (call init with test flag BeforeTest and call cleanup in AfterTest).

*Nice To Have*: We could have some initialization code in the test application class that will set some `alwaysRequireTestMode` flag in `FhirEngineProvider` that will make it throw if its configuration is not set for testing. That would allow us to catch cases where someone forgot to apply the rule.

---

As a second step, we should also consider providing an artifact for our clients to write integration tests using the real `FhirEngineProvider`. That API would depend on what we need to provide. WorkManager's testing API might be a good example to follow:
https://developer.android.com/topic/libraries/architecture/workmanager/how-to/integration-testing

cc: [ktarasenko](https://github.com/ktarasenko)

ktarasenko:

Fixes #1317

**Description**
Add reset mechanism to FhirEngineProvider.
Add FhirEngineProviderTestRule that cleans up FhirEngineProvider

**Alternative(s) considered**

**Type**
Code health | Testing

**Checklist**


build.gradle.kts
```diff
@@ -9,6 +9,7 @@ buildscript {
     classpath(Plugins.androidGradlePlugin)
     classpath(Plugins.kotlinGradlePlugin)
     classpath(Plugins.navSafeArgsGradlePlugin)
+    classpath("org.jetbrains.kotlin:kotlin-gradle-plugin:1.6.10")
```
yigit, line 12: why do we need this?

engine/build.gradle.kts
```diff
@@ -136,6 +136,7 @@ dependencies {
   testImplementation(Dependencies.AndroidxTest.core)
   testImplementation(Dependencies.AndroidxTest.workTestingRuntimeKtx)
   testImplementation(Dependencies.Kotlin.kotlinCoroutinesTest)
+  testImplementation(Dependencies.Kotlin.kotlinTestJunit)
```
yigit, line 139: we already have truth hence this seems unnecessary

engine/src/main/java/com/google/android/fhir/FhirEngineProvider.kt
```diff
@@ -51,30 +51,45 @@ object FhirEngineProvider {
...
+            setDatabaseErrorStrategy(configuration.databaseErrorStrategy)
+            configuration.serverConfiguration?.let { setServerConfiguration(it) }
+            if (configuration.testMode) {
+              inMemory()
+            }
           }
           .build()
     }
-    return fhirServices
+    return fhirServices as FhirServices
```
yigit, line 76: the as is a bit confusing, maybe just do checkNotNull(fhirServices) to clarify intend?

testing/src/main/java/com/google/android/fhir/testing/FhirEngineProviderTestRule.kt
```diff
@@ -0,0 +1,40 @@
...
+import com.google.android.fhir.FhirEngineConfiguration
+import com.google.android.fhir.FhirEngineProvider
+import org.junit.rules.TestRule
+import org.junit.runner.Description
+import org.junit.runners.model.Statement
+
+/** A [TestRule] that cleans up [FhirEngineProvider] instance after each test run. */
+class FhirEngineProviderTestRule : TestRule {
+
+  override fun apply(base: Statement, description: Description): Statement {
```
yigit, line 28: why don't we use it in our tests? a rule would be better than using `@After`

build.gradle.kts
ktarasenko: That wasn't intentional, reverted.

engine/build.gradle.kts
ktarasenko: I need it for `assertFailsWith`

engine/src/main/java/com/google/android/fhir/FhirEngineProvider.kt
ktarasenko: done

vitorpamplona:
Thank you! I was going crazy with my tests until I realized the structure wasn't cleaning up the database. This Fixes also #1340

engine/build.gradle.kts
vitorpamplona: Updating to 4.13 should give you `assertThrows`. See Issue #1341. PR coming.

vitorpamplona (approved):

testing/src/main/java/com/google/android/fhir/testing/FhirEngineProviderTestRule.kt
ktarasenko: > Also,I'm not sure why the different module makes the API more verbose, did you mean that about gradle dependencies or something else?
I meant I have to add methods that only needed for tests I'd rather not to expose (for example `cleanup` is visible for FhirEngineProvider consumers)

engine/build.gradle.kts
ktarasenko: Used `runCatching` and removed the dependency. 

For the sake of argument, I ended up having this:  
```
assertThat(runCatching { provider.cleanup() }.exceptionOrNull()).isInstanceOf(IllegalStateException::class.java)
```
I still think that this is more readable way to do it kotlin: 
```
assertFailsWith<IllegalStateException> { provider.cleanup() }
```

testing/src/main/java/com/google/android/fhir/testing/FhirEngineProviderTestRule.kt
ktarasenko: > as for cases where they want to configure the FhirEngineProvider further, the rule could receive a block for that configuration

That's not exactly the problem here. Test cases I added to the FhirEngineProviderTest verify that the provider is recreated properly after the cleanup and that the cleanup is not possible if the provider was created not in the test mode. Explicit initializations and cleanups help understand what the test is doing exactly instead of having half of the flow hidden by the rule.

engine/src/main/java/com/google/android/fhir/FhirEngineProvider.kt
```diff
@@ -51,30 +51,45 @@ object FhirEngineProvider {
...
         FhirServices.builder(context.applicationContext)
           .apply {
-            if (fhirEngineConfiguration.enableEncryptionIfSupported) enableEncryptionIfSupported()
-            setDatabaseErrorStrategy(fhirEngineConfiguration.databaseErrorStrategy)
-            fhirEngineConfiguration.serverConfiguration?.let { setServerConfiguration(it) }
+            if (configuration.enableEncryptionIfSupported) enableEncryptionIfSupported()
+            setDatabaseErrorStrategy(configuration.databaseErrorStrategy)
+            configuration.serverConfiguration?.let { setServerConfiguration(it) }
+            if (configuration.testMode) {
+              inMemory()
```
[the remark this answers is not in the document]
ktarasenko: Let's follow up on this when we face the problem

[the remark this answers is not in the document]
vitorpamplona: Agree, let's merge.

testing/build.gradle.kts
```diff
@@ -0,0 +1,34 @@
+plugins {
+  id(Plugins.BuildPlugins.androidLib)
+  id(Plugins.BuildPlugins.kotlinAndroid)
+  id(Plugins.BuildPlugins.mavenPublish)
+  jacoco
+}
+
+afterEvaluate {
+  // TODO: setup publishing
+}
```
jingtang10, line 10: remove this block - we're never gonna publish this module

jingtang10 (approved):
thanks @ktarasenko

engine/src/main/java/com/google/android/fhir/FhirEngineProvider.kt
```diff
@@ -51,30 +51,45 @@ object FhirEngineProvider {
...
 
   @Synchronized
   private fun getOrCreateFhirService(context: Context): FhirServices {
-    if (!::fhirServices.isInitialized) {
-      if (!::fhirEngineConfiguration.isInitialized) {
-        fhirEngineConfiguration = FhirEngineConfiguration()
-      }
+    if (fhirServices == null) {
+      if (fhirEngineConfiguration == null) fhirEngineConfiguration = FhirEngineConfiguration()
+      val configuration = checkNotNull(fhirEngineConfiguration)
```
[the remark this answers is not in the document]
ktarasenko: I need to keep the local variable to avoid checks few lines below in `apply` of `FhirServices.builder()`

yigit (approved):

PR #1368, opened 2022-05-06
Indexing for SearchParams defined in Resource class

### LINKED ISSUE #1305: Update search indexing logic to consider search params in full Class hierarchy.
1. It appears that engine is not traversing the full Class hierarchy of the model classes to figure out all the SearchParams that need to be indexed into the database. This is causing search by `Patient.RES_ID` to fail.

_Originally posted by @aditya-07 in https://github.com/google/android-fhir/issues/1283#issuecomment-1102336832_

aditya-07:

Fixes #1305 

**Description**
The search parameters defined in Resource class like id has path defined as `Resource.id`. With this, when a derived class like `Patient` is passed into the `FHIRPathEngine.evaluate` function, it doesn't return anything. So, we are adding the index values manually.

**Alternative(s) considered**

**Type**
Choose one: Bug fix

**Checklist**


demo/src/main/java/com/google/android/fhir/demo/FhirApplication.kt
```diff
@@ -38,7 +38,7 @@ class FhirApplication : Application() {
     }
     FhirEngineProvider.init(
       FhirEngineConfiguration(
-        enableEncryptionIfSupported = true,
+        enableEncryptionIfSupported = false,
```
jingtang10, line 41: did you need this change for testing? reverse please if so.

engine/src/main/java/com/google/android/fhir/index/ResourceIndexer.kt
```diff
@@ -98,6 +100,28 @@ internal object ResourceIndexer {
         }
       }
 
+    addIndexesFromResourceClass(resource, indexBuilder)
+    return indexBuilder.build()
+  }
+
+  /**
+   * Adds indexes for [SearchParameter] defined in [Resource] class. The reason they are added here
```
jingtang10, line 108:
```suggestion
   * Adds indexes for [SearchParameter]s defined in [Resource] class. The reason they are added here
```

engine/src/main/java/com/google/android/fhir/index/ResourceIndexer.kt
```diff
@@ -98,6 +100,28 @@ internal object ResourceIndexer {
...
+   * manually is because the path for these [SearchParameter]s is defined for [Resource] class and
```
jingtang10, line 109: "The reason they are added here manually is because" -> "This is because"

engine/src/main/java/com/google/android/fhir/index/ResourceIndexer.kt
```diff
@@ -98,6 +100,28 @@ internal object ResourceIndexer {
...
+   * [FHIRPathEngine.evaluate] doesn't return anything when run against a derived resource like
```
jingtang10, line 110: The actual reason is the code generator doesn't generate search parameters for the base resource class right?

jingtang10 (changes requested):

demo/src/main/java/com/google/android/fhir/demo/FhirApplication.kt
aditya-07: Yes, oversight on my end. Will revert it.

engine/src/main/java/com/google/android/fhir/index/ResourceIndexer.kt
aditya-07: Right and also because the expression path is defined for the resource class and hence when a derived class like `Patient` is evaluated for id using expression path `Resource.id`, `FHIRPathEngine` doesn't return anything. it expects the path to be "Patient.id".

demo/src/main/java/com/google/android/fhir/demo/FhirApplication.kt
aditya-07: Done

engine/src/main/java/com/google/android/fhir/index/ResourceIndexer.kt
aditya-07: Updated the comments to explain the reason.

engine/src/main/java/com/google/android/fhir/index/ResourceIndexer.kt
aditya-07: Done

engine/src/main/java/com/google/android/fhir/index/ResourceIndexer.kt
aditya-07: Done

jingtang10 (approved):

Issue #1173, opened 2022-02-15
FhirPathEngine should use existing ValidationSupport instead of init again


maimoonak:
**Describe the Issue**

The line where we are [generating FhirPathEngine ](https://github.com/google/android-fhir/blob/master/datacapture/src/main/java/com/google/android/fhir/datacapture/mapping/ResourceMapper.kt#L76) creates a new instance of DefaultProfileValidationSupport which occupies more than 17MB. If this is init into app anywhere again it keep multiplying and causes a performance lag. The method getValidation context into FhirContext does it automatically hence this should use validationSupport from FhirContext.

Impact on app. 
Around 500MB performance impact was noticed in profiling with doing this.

Before
![Screenshot from 2022-02-05 01-35-38](https://user-images.githubusercontent.com/4829880/154071570-88f3f3e2-a09e-4de7-b206-7373fd8d404d.png)

After
![Screenshot from 2022-02-14 21-32-39](https://user-images.githubusercontent.com/4829880/154071689-4994128a-204d-4a7b-b6af-4661cb647a1d.png)

**Would you like to work on the issue?**

yes

jingtang10:
@maimoonak do you have a proposed fix for this?

maimoonak:
Done in #1328

Tarun-Bhardwaj:
Closing the issue as per discussion with Maimoona.

Issue #1059, opened 2022-01-21
Add variable Extension support


fredhersch:
**Is your feature request related to a problem? Please describe.**
Desire to use variable names to perform calculations via SDC. 

For example, to calculate BMI: (weight_value/heigh_value*height_value)

* From the current Questionnaire, it is simpler to set a variable and then use that in the calculation. See [example here](http://build.fhir.org/ig/HL7/sdc/examples.html#calculatedexpression)

**Describe the solution you'd like**
Implement variable Extension support - [see specification](http://build.fhir.org/ig/HL7/sdc/expressions.html)

jingtang10:
my first question is that do you evaluate the expressions lazily or do you evaluate them before hand?

do you cache the values? and if you do how do you update them when the questionnaire answers change?

shoaibmushtaq25:
As per discussion with @jingtang10 over the call, First, I will explore whether FHIRPath supports the variables to pass them as input parameters to evaluate an expression. If I wouldn't be able to find it, I should ask a question over the Zulip FHIR chat to get an opinion from the community experts.

Let's say, if FHIR Path supports the variables to pass them as input parameters to evaluate an expression, then we need to evaluate the variable values before hand and then pass these values to the FHIR path to evaluate further expression (i.e, BMI calculation)

Also, I have to explore that does this variable used in the SDC population or not. 
@fredhersch Can you share some examples of usage of variables in the population if there is any possibility to use it in population.

@jingtang10 Can you please review my comment and let me know If I missed something here, Thanks.
cc: @jingtang10 @f-odhiambo

shoaibmushtaq25:
@jingtang10 , I posted [this query](https://chat.fhir.org/#narrow/stream/179255-questionnaire/topic/FHIRPATH.20SDC.20implementation/near/276058530) over Zulip FHIR Chat and got an answer [here](https://chat.fhir.org/#narrow/stream/179255-questionnaire/topic/FHIRPATH.20SDC.20implementation/near/276069745). As per the comment, It should work in FHIR Path, I am going to try this out how can we apply this using FHIR Path.

shoaibmushtaq25:
I tried to evaluate the fhirpath expressions like `%resource.repeat(item).where(linkId='3.3.1').item.answer.valueDecimal` as mentioned in the questionnaire [here](http://build.fhir.org/ig/HL7/sdc/Questionnaire-questionnaire-sdc-profile-example-form-behavior.json.html) and this worked well.

[...]

PR #1380, opened 2022-05-11
Implement calculated-expression extension

### LINKED ISSUE #971: Support calculatedExpression
See #803

See example of calculatedExpression and variable: http://build.fhir.org/ig/HL7/sdc/examples.html#using-calculatedexpression-and-variable

@maimoonak provided this example where we want to take age OR birthdate as input and calculate the other from one which was input: https://github.com/google/android-fhir/blob/mssng-quest-feat/datacapturegallery/src/main/assets/sample_family_member_registration_calculatedexpression.json

Probably use FHIRPath

NA

**Additional context**
NA

NA

maimoonak:
Fixes #971 #1173

**Description**
Implemented the calculated expression extension. The feature also calculated cyclic dependency, and prevents the infinite calls of callback

**Type**
Choose one: Feature

**Checklist**


f-odhiambo:
@RaaziaTarique kindly take a 1st pass review

RaaziaTarique (approved):
LGTM

Tarun-Bhardwaj:
Assigning to @aditya-07 for review.

datacapture/src/main/java/com/google/android/fhir/datacapture/QuestionnaireViewModel.kt
```diff
@@ -184,6 +200,69 @@ internal class QuestionnaireViewModel(application: Application, state: SavedStat
...
+                      .indexOfFirst { it.questionnaireItem.linkId == questionnaireItem.key }
+                      .let { if (it > -1) _questionnaireItemValueStateFlow.emit(it) }
+                  }
+              }
+            }
+          }
+        }
+      }
+    }
+  }
```
aditya-07, line 246:
```suggestion
  private fun runCalculatedExpressions() {
    linkIdToQuestionnaireItemMap.filter { it.value.calculatedExpression != null }
      .forEach { questionnaireItem ->
        linkIdToQuestionnaireResponseItemMap[questionnaireItem.key]?.let { questionnaireResponseItem
          ->
          questionnaireItem.value.calculatedExpression?.expression?.let { expression ->
            fhirPathEngine.evaluate(null, questionnaireResponse, null, null, expression).firstOrNull()?.let {
              val evaluatedAnswer = it.castToType(it)
              val currentAnswer = if (questionnaireResponseItem.hasAnswer()) questionnaireResponseItem.answerFirstRep.value else null
              if (!evaluatedAnswer.equalsDeep(currentAnswer)) {
                questionnaireResponseItem.answerFirstRep.value = evaluatedAnswer
              }
              // notify UI to update it value i.e. notify item changed to adapter
              viewModelScope.launch {
                if (modificationCount.value > 0) {
                  questionnaireStateFlow.collectLatest {
                    it.items
                      .indexOfFirst { it.questionnaireItem.linkId == questionnaireItem.key }
                      .let { if (it > -1) _questionnaireItemValueStateFlow.emit(it) }
                  }
                }
              }
            }
          }
        }
      }
  }
```

datacapture/src/main/java/com/google/android/fhir/datacapture/QuestionnaireViewModel.kt
```diff
@@ -366,6 +446,9 @@ internal class QuestionnaireViewModel(application: Application, state: SavedStat
    * when first opening this questionnaire. Otherwise, returns `null`.
    */
   private fun Questionnaire.getInitialPagination(): QuestionnairePagination? {
+    detectCalculatedExpressionCyclicDependency()
```
aditya-07, line 449: Should `detectCalculatedExpressionCyclicDependency` be called  right after the `linkIdToQuestionnaireItemMap` is generated to fail early.

datacapture/src/main/java/com/google/android/fhir/datacapture/QuestionnaireViewModel.kt
```diff
@@ -137,6 +150,9 @@ internal class QuestionnaireViewModel(application: Application, state: SavedStat
         }
       }
     }
+
+    runCalculatedExpressions()
```
aditya-07, line 154: This may cause expressions to be calculated again and again. Should we restrict the scope to calculate items that depend on the item being answered in this callback? cc: @jingtang10 

On another note, If a user is entering text in the answer, this callback gets called after each character is added/removed. Maybe we should add some sort of wait to let user complete typing before the callback is called (like Rx debounce). cc: @jingtang10 @stevenckngaa @yigit

catalog/src/main/assets/calculated_expression_questionnaire.json
```diff
@@ -0,0 +1,37 @@
...
+      "text": "Age years",
+      "hint": "Input age to automatically calculate birthdate in case exact birthdate is not known",
+      "type": "quantity",
+      "initial": [{
+        "valueQuantity": {
+          "unit": "months",
+          "system": "http://unitsofmeasure.org",
+          "code": "months"
+        }
+      }]
```
aditya-07, line 29: Do we use this information to let user know what to input?

catalog/src/main/java/com/google/android/fhir/catalog/ComponentListViewModel.kt
```diff
@@ -77,5 +77,10 @@ class ComponentListViewModel(application: Application, private val state: SavedS
     ),
     SLIDER(R.drawable.ic_slider, R.string.component_name_slider, "slider_questionnaire.json"),
     IMAGE(R.drawable.ic_image, R.string.component_name_image, ""),
+    CALCULATED_EXPRESSION(
+      R.drawable.ic_unitoptions,
+      R.string.component_name_calculated_expression,
+      "calculated_expression_questionnaire.json"
+    ),
```
aditya-07, line 84: Should the option to see Calculated expressions in the Catalog app be part of Layouts tab or maybe a new tab altogether ? cc: @santosh-pingle  @jingtang10

jingtang10 (changes requested):
please complete the description of the PR

catalog/src/main/assets/calculated_expression_questionnaire.json
maimoonak: That is a different extension which allows to give user an option to choose unit of quantity http://hl7.org/fhir/R4/extension-questionnaire-unitoption.html. This is initial https://build.fhir.org/ig/HL7/sdc/behavior.html#initial where we can make type quantity as initial without value for unit reference purpose. This makes it default to the unit type specified in initial

datacapture/src/main/java/com/google/android/fhir/datacapture/QuestionnaireViewModel.kt
maimoonak: Made the suggested changes, however, two important points
- We can not skip null evaluations, so even if answer is null or empty we do set the value to erase the  previous calculated value 
- To prevent loop back we do verify that the new answer is not same as previous one. A dependents value change may not necessarily update the value. For e.g. if we calculate age in years from birthdate changing dates may not update age unless age changes its digit, similarly if we are ignoring decimals for any calculated value we may have same value as previous one.

datacapture/src/main/java/com/google/android/fhir/datacapture/QuestionnaireViewModel.kt
maimoonak: Moved the calculation after state has been updated and ready to be rendered. This would prevent some of the extra calls.
For debounce the delay may not work each case. For e.g. automation script can change value in millis and can submit form, and then we would end up missing the calculations, a QA user submit forms very quickly and the actual user may take 1 sec pause for each character. I also checked the isFocused, isActivated, but nothing seems to be relevant. The code for calculations prevents the unnecessary updates to some extent. The dependency graph generation is a bit tricky as calculation-expression could be anything, sum of multiple widgets, referring other items by linkId or by their value types etc, or other variables. User should be aware that using too much calculations can impact performance.
Any suggestions!

maimoonak:
@jingtang10 @aditya-07 Please review the changes

Tarun-Bhardwaj:
@aditya-07 @jingtang10 could you please review the changes?

datacapture/src/main/java/com/google/android/fhir/datacapture/QuestionnaireViewModel.kt
```diff
@@ -182,8 +197,77 @@ internal class QuestionnaireViewModel(application: Application, state: SavedStat
...
+      val otherDependent =
+        calculableItems.values.find {
+          it.calculatedExpression!!.expression.contains("'${current.key}'")
+        }
+      // if any calculable expression depends on this item and this item is referring to the
+      // dependent item in its own expression then raise error
+      if (otherDependent != null && currentExpression.contains("'${otherDependent.linkId}'"))
+        throw IllegalStateException(
+          "${current.key} and ${otherDependent.linkId} have cyclic dependency in calculated-expression extension"
+        )
```
aditya-07, line 267: Can we use `kotlin.check` here see https://kotlinlang.org/api/latest/jvm/stdlib/kotlin/check.html.

aditya-07 (approved):
Small nit otherwise LGTM!

datacapture/src/main/java/com/google/android/fhir/datacapture/QuestionnaireViewModel.kt
```diff
@@ -237,6 +321,7 @@ internal class QuestionnaireViewModel(application: Application, state: SavedStat
         )
       }
     }
+
```
jingtang10, line 324: nit: please revert this change. this makes the history look weird.

datacapture/src/main/java/com/google/android/fhir/datacapture/QuestionnaireViewModel.kt
```diff
@@ -328,6 +416,7 @@ internal class QuestionnaireViewModel(application: Application, state: SavedStat
               .items
         }
         .toList()
+
```
jingtang10, line 419: nit: same here. please revert.

datacapture/src/main/java/com/google/android/fhir/datacapture/QuestionnaireViewModel.kt
```diff
@@ -138,6 +151,8 @@ internal class QuestionnaireViewModel(application: Application, state: SavedStat
       }
     }
     modificationCount.value += 1
+
+    runCalculatedExpressions()
```
jingtang10, line 155: can you move this before line 153?

datacapture/src/main/java/com/google/android/fhir/datacapture/QuestionnaireViewModel.kt
```diff
@@ -182,8 +197,77 @@ internal class QuestionnaireViewModel(application: Application, state: SavedStat
...
             pagination = questionnaire.getInitialPagination(),
           )
+            .also { runCalculatedExpressions() }
       )
 
+  /**
+   * Notify the updated value index to notify about the item update for UI (adapter item changed)
+   */
+  private val _questionnaireItemValueStateFlow = MutableSharedFlow<Int>()
+  internal val questionnaireItemValueStateFlow = _questionnaireItemValueStateFlow.asSharedFlow()
```
jingtang10, line 207: i kinda don't understand why these new stateflows are needed? can the modificationCount not be used in order to refresh the UI?

jingtang10 (changes requested):
Can you attach a screenshot please?

@shelaghm not sure if we want to show case this in the layout screen of the catalog app... i kind feel like this is better off in the components page... that is if we want to showcase this at all.

jingtang10:
@joiskash can you please leave a review? thanks!

shelaghm:
> Can you attach a screenshot please?
> 
> @shelaghm not sure if we want to show case this in the layout screen of the catalog app... i kind feel like this is better off in the components page... that is if we want to showcase this at all.

I agree that it's likely more useful to show in components page, not layout page. Having a screenshot that I can look at of this components will he helpful for me to provide more specific guidance.

datacapture/src/main/java/com/google/android/fhir/datacapture/QuestionnaireViewModel.kt
```diff
@@ -182,8 +197,77 @@ internal class QuestionnaireViewModel(application: Application, state: SavedStat
...
+                  }
+                }
+              }
+            }
+        }
+      }
+    }
+  }
+
+  fun detectCalculatedExpressionCyclicDependency(
```
joiskash, line 251: This seems like it could be moved to a function similar to checkQuestionnaireResponse in the QuestionnaireResponseValidator. Just a suggestion, does not have to be done.

joiskash (changes requested):
Nice work! Left a few comments

datacapture/src/main/java/com/google/android/fhir/datacapture/QuestionnaireViewModel.kt
maimoonak: yes. it is needed because 1- _questionnaireItemValueStateFlow does not update and notify list unless item count is changed (a question is hidden/shown)  2- item value update notification is not triggered in every case, rather it is updated only if value is changed from previous one. For ex: calculation of age from birthdate can lead to same value for all/most birthdates of a year, which means age input should on;y be notified iff age value is updated. (see L#235-240). So in short this calculation and its value event is notified only in specific case and not with every modificationCount++

datacapture/src/main/java/com/google/android/fhir/datacapture/QuestionnaireFragment.kt
```diff
@@ -103,6 +103,12 @@ open class QuestionnaireFragment : Fragment() {
         }
       }
     }
+
+    viewLifecycleOwner.lifecycleScope.launchWhenCreated {
+      viewModel.questionnaireItemValueStateFlow.collect { index ->
+        if (!recyclerView.isComputingLayout) adapter.notifyItemChanged(index)
+      }
+    }
```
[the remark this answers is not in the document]
maimoonak: yes, we need to do so. The list does not update the value unless a question is shown / hidden. This leads to notify item change which leads to item update in UI. Without this item value is not updated

datacapture/src/main/java/com/google/android/fhir/datacapture/QuestionnaireViewModel.kt
maimoonak: This method runs only on questionnaire items and does not have any link with questionnaireResponse. It runs only once for each questionnaireItem having the extension 'calculated-expression'

datacapture/src/main/java/com/google/android/fhir/datacapture/views/QuestionnaireItemDatePickerViewHolderFactory.kt
```diff
@@ -82,7 +82,11 @@ internal object QuestionnaireItemDatePickerViewHolderFactory :
         header.bind(questionnaireItemViewItem.questionnaireItem)
 
         textInputEditText.setText(
-          questionnaireItemViewItem.singleAnswerOrNull?.valueDateType?.localDate?.localizedString
+          questionnaireItemViewItem.singleAnswerOrNull
+            ?.takeIf { it.hasValue() }
+            ?.valueDateType
+            ?.localDate
+            ?.localizedString
```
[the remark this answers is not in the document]
maimoonak: yes. If not fixed it throws following exception on initialization of dateType value. 
```
java.lang.NullPointerException: year must not be null
        at com.google.android.fhir.datacapture.views.QuestionnaireItemDatePickerViewHolderFactoryKt.getLocalDate(QuestionnaireItemDatePickerViewHolderFactory.kt:148)
        at com.google.android.fhir.datacapture.views.QuestionnaireItemDatePickerViewHolderFactory$getQuestionnaireItemViewHolderDelegate$1.bind(QuestionnaireItemDatePickerViewHolderFactory.kt:85)
        at com.google.android.fhir.datacapture.views.QuestionnaireItemViewHolder.bind(QuestionnaireItemViewHolderFactory.kt:65)
        at com.google.android.fhir.datacapture.QuestionnaireItemAdapter.onBindViewHolder(QuestionnaireItemAdapter.kt:99)
```

datacapture/src/main/java/com/google/android/fhir/datacapture/QuestionnaireViewModel.kt
```diff
@@ -182,8 +197,77 @@ internal class QuestionnaireViewModel(application: Application, state: SavedStat
...
+   * Notify the updated value index to notify about the item update for UI (adapter item changed)
+   */
+  private val _questionnaireItemValueStateFlow = MutableSharedFlow<Int>()
+  internal val questionnaireItemValueStateFlow = _questionnaireItemValueStateFlow.asSharedFlow()
+
+  private fun runCalculatedExpressions() {
+    linkIdToQuestionnaireItemMap.filter { it.value.calculatedExpression != null }.forEach {
+      questionnaireItem ->
+      linkIdToQuestionnaireResponseItemMap[questionnaireItem.key]?.let { questionnaireResponseItem
+        ->
```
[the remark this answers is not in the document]
maimoonak: there is a variables PR for #1059 where @shoaibmushtaq25 is working for performance optimization of expressions. The calculated expression depends on variables as well, and any random valid expression as well. So building a tree is not straight forward. We are detecting cyclic dependency tree which would not allow endless loop, and we are also breaking the circuit by stopping the update notification if value is not updated. https://build.fhir.org/ig/HL7/sdc/expressions.html#calculatedExpression and https://build.fhir.org/ig/HL7/sdc/expressions.html#dependencies does not give any clear direction on this as well. We can work on this as separate ticket once variables PR is also merged and we are done evaluating different use cases.
Also we can have uses cases when we depend on multiple items for calculation without inclusion of linkId. For e.g. To calculate risk score based on 10 questions for covid health survey we may use expression like following

`%resource.descendants().where(linkId='group-id').select(answer.value).where(code='yes').count()`

The group for covid danger signs can have 10 danger signs and based on yes for each we may calculate risk score. Simialry I have seen a similar example in a metal health survey. 
This performance optimization would need a thorough analysis.

There is one more condition that needs to be implemented " If a user has edited a calculated value, it should no longer be changed when other answers in the questionnaire response change". Do we have a way to know if an input was updated by user or program?

Tarun-Bhardwaj:
@maimoonak could you please advise an ETA by when these changes can be ready for review?

maimoonak:
@Tarun-Bhardwaj I have resolved all conflicts with main, synced the code with new changes as well just now. 
The issue https://github.com/google/android-fhir/issues/1498 needs to be resolved to run the catalog (without extra needless edit-text). But this PR and functionality does not depend on this issue and is working as expected. (Below is a working example)

ezgif com-gif-maker

Tarun-Bhardwaj:
@maimoonak , could you please advise the status of these changes? Is there an ETA by when we can merge this PR?

maimoonak:
@Tarun-Bhardwaj @jingtang10 kindly review the PR.

datacapture/src/main/java/com/google/android/fhir/datacapture/fhirpath/ExpressionEvaluator.kt
```diff
@@ -64,6 +68,66 @@ object ExpressionEvaluator {
...
+            "${current.linkId} and ${dependent.linkId} have cyclic dependency in expression based extension"
+          }
+        }
+      }
+    }
+  }
+
+  /**
+   * Returns a pair of item and the calculated and evaluated value for all items with calculated
+   * expression extension, which is dependent on value of updated response
```
jingtang10, line 90: can you please rephrase this? not quite clear to me.

datacapture/src/main/java/com/google/android/fhir/datacapture/MoreQuestionnaireItemComponents.kt
```diff
@@ -295,6 +324,15 @@ fun QuestionnaireResponse.QuestionnaireResponseItemComponent.addNestedItemsToAns
   }
 }
 
+/**
+ * Flatten a nested list of [QuestionnaireResponse.QuestionnaireResponseItemAnswerComponent]
+ * recursively and returns a flat list of all items into list embedded at any level
+ */
+fun List<Questionnaire.QuestionnaireItemComponent>.flattened():
+  List<Questionnaire.QuestionnaireItemComponent> {
+  return this + this.flatMap { if (it.hasItem()) it.item.flattened() else it.item }
```
jingtang10, line 333: i don't think you need the if else here. if the condition in the`if` clause is false, `it.item` will always evaluate to empty list.

datacapture/src/main/java/com/google/android/fhir/datacapture/fhirpath/ExpressionEvaluator.kt
```diff
@@ -64,6 +68,66 @@ object ExpressionEvaluator {
...
+    questionnaireItemParentMap:
+      Map<Questionnaire.QuestionnaireItemComponent, Questionnaire.QuestionnaireItemComponent>
+  ): List<Pair<Questionnaire.QuestionnaireItemComponent, List<Type>>> {
+    return questionnaire
+      .item
+      .flattened()
+      .filter { item ->
+        // item is calculable and not modified yet and depends on the updated answer
+        item.calculatedExpression != null &&
+          modifiedResponses.none { it.linkId == item.linkId } &&
```
jingtang10, line 106: add a comment to say why this condition is needed.

datacapture/src/main/java/com/google/android/fhir/datacapture/QuestionnaireViewModel.kt
```diff
@@ -366,8 +370,46 @@ internal class QuestionnaireViewModel(application: Application, state: SavedStat
...
+        val currentAnswer = updatedCalculableResponse?.answer?.map { it.value }
+
+        // update and notify only if answer has changed to prevent any event loop
+        // if current and previous both are not empty and the answer is changed in count or
+        // content
+        if ((evaluatedAnswer + currentAnswer).isNotEmpty() &&
+            (evaluatedAnswer != currentAnswer ||
+              evaluatedAnswer
+                .filterIndexed { i, v -> currentAnswer[i].equalsDeep(v).not() }
+                .isEmpty())
```
jingtang10, line 399: can you extract this into a small private helper function in this class.

also feels like you should be able to use `zip` instead of `filterIndexed`.

jingtang10 (changes requested):

datacapture/src/main/java/com/google/android/fhir/datacapture/QuestionnaireViewModel.kt
maimoonak: simplified the logic using zip

jingtang10:
can you please make sure to reply to the review comments and mark the conversations as resolved if the comment is addressed in the code. once that's done please reassign back to me. thanks

jingtang10 (changes requested):

jingtang10:
@maimoonak please reassign to me once all comments are addressed

datacapture/src/main/java/com/google/android/fhir/datacapture/MoreQuestionnaireItemComponents.kt
```diff
@@ -87,8 +93,29 @@ internal fun Questionnaire.QuestionnaireItemComponent.findVariableExpression(
...
+    this.getExtensionByUrl(EXTENSION_CALCULATED_EXPRESSION_URL)?.let {
+      it.castToExpression(it.value)
+    }
+
+/** Returns list of extensions whose value is of type [Expression] */
+internal val Questionnaire.QuestionnaireItemComponent.expressionBasedExtensions
+  get() = this.extension.filter { it.value is Expression }
+
+/**
+ * Check if given [item] has calculable expression based extension and is referencing current item
```
jingtang10, line 108:
```suggestion
 * Whether [item] has any expression directly referencing the current questionnaire item by link ID (e.g. if [item] has an expression `%resource.item.where(linkId='this-question')` where `this-question` is the link ID of the current questionnaire item).
```

datacapture/src/main/java/com/google/android/fhir/datacapture/fhirpath/ExpressionEvaluator.kt
```diff
@@ -61,6 +65,74 @@ object ExpressionEvaluator {
...
+      Map<Questionnaire.QuestionnaireItemComponent, Questionnaire.QuestionnaireItemComponent>
+  ): List<Pair<Questionnaire.QuestionnaireItemComponent, List<Type>>> {
+    return questionnaire
+      .item
+      .flattened()
+      .filter { item ->
+        // 1- item is calculable
+        // 2- item answer is not modified and touched by user;
+        // https://build.fhir.org/ig/HL7/sdc/StructureDefinition-sdc-questionnaire-calculatedExpression.html
+        // 3- item answer depends on the updated item answer OR has a variable dependency
```
jingtang10, line 104: why do you also recalculate everything that has a variable dependency?

datacapture/src/main/java/com/google/android/fhir/datacapture/fhirpath/ExpressionEvaluator.kt
```diff
@@ -61,6 +65,74 @@ object ExpressionEvaluator {
...
+        item.calculatedExpression != null &&
+          modifiedResponses.none { it.linkId == item.linkId } &&
+          (updatedQuestionnaireItem.isReferencedBy(item) ||
+            findDependentVariables(item.calculatedExpression!!).isNotEmpty())
+      }
+      .map { calculable ->
```
jingtang10, line 110: i would prefer it if we refrain from introducing new terminology without defining them properly.

can we just use `questionnaireItem`?

datacapture/src/main/java/com/google/android/fhir/datacapture/QuestionnaireViewModel.kt
```diff
@@ -366,8 +370,41 @@ internal class QuestionnaireViewModel(application: Application, state: SavedStat
             currentPageIndex = getInitialPageIndex(),
             reviewMode = enableReviewPage
           )
+            .also { detectExpressionCyclicDependency(questionnaire.item) }
+            .also { questionnaire.item.flattened().forEach { runCalculations(it) } }
       )
 
+  fun runCalculations(updatedQuestionnaireItem: Questionnaire.QuestionnaireItemComponent) {
```
jingtang10, line 377: I think a much better name for this function is `updateDependentQuestionnaireResponseItems`. and please add documentation to explain.

datacapture/src/main/java/com/google/android/fhir/datacapture/fhirpath/ExpressionEvaluator.kt
```diff
@@ -61,6 +65,74 @@ object ExpressionEvaluator {
...
+    return questionnaire
+      .item
+      .flattened()
+      .filter { item ->
+        // 1- item is calculable
+        // 2- item answer is not modified and touched by user;
+        // https://build.fhir.org/ig/HL7/sdc/StructureDefinition-sdc-questionnaire-calculatedExpression.html
+        // 3- item answer depends on the updated item answer OR has a variable dependency
+        item.calculatedExpression != null &&
+          modifiedResponses.none { it.linkId == item.linkId } &&
```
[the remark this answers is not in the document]
jingtang10: actually this won't work. please remove `modifiedResponses` from the function signature of this function, and do this check in the `QuestionnareViewModel` (around line 387).

this is because the modified responses actually include response items... it's not a set of questionnaire items.

datacapture/src/main/java/com/google/android/fhir/datacapture/QuestionnaireViewModel.kt
```diff
@@ -366,8 +370,41 @@ internal class QuestionnaireViewModel(application: Application, state: SavedStat
...
+    evaluateCalculatedExpressions(
+      updatedQuestionnaireItem,
+      questionnaire,
+      questionnaireResponse,
+      modifiedQuestionnaireResponseItemSet,
+      questionnaireItemParentMap
+    )
+      .forEach { updatedCalculable ->
+        val updatedCalculableResponse =
+          questionnaireResponseItemPreOrderList.find { it.linkId == updatedCalculable.first.linkId }
```
jingtang10, line 387: you probably should use `findAll` instead of `find`, because you want to update all questionnaire response items with the same link id (meaning they originate from the same questionnaire item with the link id).

jingtang10 (changes requested):
i quite strong feel at this point we should create a dependency graph for expressions. the logic in your evaluator is getting very complex with part of it handling finding dependency and part of it handling the actual evaluation of variables.

i think we could possibly merge this without this change. but it will need to be an immediate refactoring i think.

datacapture/src/main/java/com/google/android/fhir/datacapture/QuestionnaireViewModel.kt
jingtang10: please still do this

datacapture/src/main/java/com/google/android/fhir/datacapture/fhirpath/ExpressionEvaluator.kt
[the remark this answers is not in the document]
jingtang10: discussed with @maimoonak, what should happen here is:

drop a questionnaire item, if all the questionnaire response items with the same link id have been modified (i.e. present in the modified set).

jingtang10 (changes requested):

datacapture/src/main/java/com/google/android/fhir/datacapture/fhirpath/ExpressionEvaluator.kt
maimoonak: as discussed in call, the variable dependency is not predictable. We want to evaluate all variables to ensure nothing is missed out

maimoonak:
@jingtang10 I have found few libraries for graph generation. I am still exploring these and if its possible to use any for our expression dependency graph
https://github.com/systek/dataflow
https://github.com/jgrapht/jgrapht/
https://github.com/google/guava/wiki/GraphsExplained
https://github.com/nidi3/graphviz-java#user-content-api

jingtang10 (approved):

PR #1384, opened 2022-05-11
Fix FHIR Engine search sort for DateTime properties

### LINKED ISSUE #1363: FHIR Engine sort by DateTime property not working
**Describe the bug**
Performing a FHIR Engine search with sort on a DateTime property does not work

**Component**
Core library

**To Reproduce**

Run the following query inside code and you notice that the result is neither sorted nor filtered by the `Patient.meta.lastUpdated` date time
```
fhirEngine.search<Patient> {

        sort(DateClientParam("_lastUpdated"), Order.ASCENDING)
        filter(DateClientParam("_lastUpdated"), {
          value = of(DateTimeType(Date(lastRecordUpdatedAt)))
          prefix = ParamPrefixEnum.GREATERTHAN_OR_EQUALS
        })

        //sort(DateClientParam("_lastUpdated"), Order.ASCENDING)
        count = batchSize
      }
```

**Expected behavior**
The result should respect the filter and sort by `_lastUpdated`. The debugged query should use `DateTimeIndexEntity` instead of `DateIndexEntity` because the `_lastUpdated` field is stored inside the `DateTimeIndexEntity` table

**Screenshots**
N/A

N/A

**Additional context**

Yes but I need help understanding how the current search API logic works and how this was supposed to work

ekigamba:
Fixes #1363

**Description**
Implements a double-join to the tables `DateTimeIndexEntity` and `DateIndexEntity` whenever a DateClientParam sort is requested

**Alternative(s) considered**
Yes, a number of alternatives were considered and discussed [here](https://docs.google.com/document/d/1ALYulBGNwSg-Agqbx4qcgOaAVtUg70GXL_mOTnN5v1s/edit#) namely:
- Merging the two tables into 1
- Generating code to hint/guide on the right table depending the the sort property
- Add a DateTimeClientParam

**Type**
Choose one: Bug fix

N/A

**Checklist**


Rkareko (approved):

Tarun-Bhardwaj:
@ekigamba , when can we expect this PR to get completed?

ekigamba:
@Tarun-Bhardwaj This PR is complete

engine/src/main/java/com/google/android/fhir/search/DateTimeClientParam.kt
```diff
@@ -0,0 +1,21 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.google.android.fhir.search
+
+import ca.uhn.fhir.rest.gclient.DateClientParam
+
+class DateTimeClientParam(theParamName: String) : DateClientParam(theParamName)
```
aditya-07, line 21: This might not be intuitive to the app developers that they need to pass DateTimeClientParam for date & time in the search sort function . Also they will have to re-define the `DateClientParam` define in hapi lib as a `DateTimeClientParam` now.

 ```
fun sort(parameter: DateClientParam, order: Order) {
    sort = parameter
    this.order = order
  }
```
One way to handle date-time could be to pass a boolean param in the `sort` indicating the type of search i.e Date or DateTime.
Any thoughts @jingtang10 ?

Comment (ekigamba) [reply engine/src/main/java/com/google/android/fhir/search/DateTimeClientParam.kt:21]: Thanks for bringing it up:

1. We can have below function to make it more intuitive
```
  fun sort(parameter: DateTimeClientParam, order: Order) {
    sort = parameter
    this.order = order
  }
```

2. Do we really need to define `DateTimeClientParam` in the hapi lib since this is only used for sorting on this lib? Having extended `DateClientParam` means it can be reused for other places where `DateClientParam` is supposed to be used without changing the code to accomodate the new class. It's only role will be better readability when passed as a param in other `DateClientParam` functions

Alternatively, we can use a boolean as you stated. This will require us to track the boolean and have it as a class property of `Search` that will only be used whenever the sort param is a `DateClientParam`

**Comment by jingtang10** [comment]
discussed this on thursday (@f-odhiambo @ekigamba @aditya-07 @maimoonak). the root problem is the fact that there is only 1 type of search param for both date and datetime indices, but at run time we don't know which index table to join.

we talked about 2 options:

1. merge date index table with datetime index table.
2. update the generated code so that there's a hint as to which params are for date index and which are for datetime index.

@ekigamba can you investigate these 2 options a little bit and discuss the pros and cons so we can make a decision? thanks

**Comment by Tarun-Bhardwaj** [comment]
> discussed this on thursday (@f-odhiambo @ekigamba @aditya-07 @maimoonak). the root problem is the fact that there is only 1 type of search param for both date and datetime indices, but at run time we don't know which index table to join.
> 
> we talked about 2 options:
> 
> 1. merge date index table with datetime index table.
> 2. update the generated code so that there's a hint as to which params are for date index and which are for datetime index.
> 
> @ekigamba can you investigate these 2 options a little bit and discuss the pros and cons so we can make a decision? thanks

@ekigamba , did you get a chance to work on the comments provided by @jingtang10 ?

**Comment by ekigamba** [comment]
> > discussed this on thursday (@f-odhiambo @ekigamba @aditya-07 @maimoonak). the root problem is the fact that there is only 1 type of search param for both date and datetime indices, but at run time we don't know which index table to join.
> > we talked about 2 options:
> > 
> > 1. merge date index table with datetime index table.
> > 2. update the generated code so that there's a hint as to which params are for date index and which are for datetime index.
> > 
> > @ekigamba can you investigate these 2 options a little bit and discuss the pros and cons so we can make a decision? thanks
> 
> @ekigamba , did you get a chance to work on the comments provided by @jingtang10 ?

Yea, in progress. Investigating both and how Hapi FHIR server handles this currently

**Comment by jingtang10** [comment]
https://github.com/google/android-fhir/issues/1363#issuecomment-1175075572

**Review by Rkareko — APPROVED** [review]

File: engine/src/main/java/com/google/android/fhir/search/MoreSearch.kt [hunk 2]
```diff
@@ -73,11 +73,30 @@ internal fun Search.getQuery(
...
       """
       ORDER BY b.${sortTableName.columnName} ${order.sqlString}
       """.trimIndent()
-    sortArgs += sort.paramName
+
+    // Fixes issue https://github.com/google/android-fhir/issues/1363
+    if (sort is DateClientParam) {
+      sortOrderStatement +=
+        ", bDateTime.${SortTableInfo.DATE_TIME_SORT_TABLE_INFO.columnName} ${order.sqlString}"
+    }
```
jingtang10, line 99: this code is very procedural and difficult to read - i think what should happen is for the `when` clause on line 65, you sould return a list of `SortTableInfo`s. And for each of them, you add the join statement and the sort order statement. that way it's much clearer.

so in this case or the `DateClientParam`, you just need to add both date and datetime tables.

jingtang10 (changes requested):

engine/src/test/java/com/google/android/fhir/search/SearchTest.kt
```diff
@@ -2119,6 +2124,19 @@ class SearchTest {
...
+    val search =
+      Search(ResourceType.Patient).apply { sort(DateClientParam("_lastUpdated"), Order.DESCENDING) }
+
+    val searchQuery = search.getQuery()
+
+    assertThat(searchQuery.query).contains("LEFT JOIN DateIndexEntity b")
+    assertThat(searchQuery.query).contains("LEFT JOIN DateTimeIndexEntity c")
+    assertThat(searchQuery.query).contains("ORDER BY b.index_from DESC, c.index_from DESC")
+  }
+
```
jingtang10, line 2139: is this test case superfulous because you already had the two test cases before?
```suggestion

```

jingtang10 (changes requested):
just one comment on test - otherwise good to go

engine/src/main/java/com/google/android/fhir/search/MoreSearch.kt
```diff
@@ -61,23 +61,39 @@ internal fun Search.getQuery(
...
   sort?.let { sort ->
-    val sortTableName =
+    val sortTableNames =
       when (sort) {
-        is StringClientParam -> SortTableInfo.STRING_SORT_TABLE_INFO
-        is NumberClientParam -> SortTableInfo.NUMBER_SORT_TABLE_INFO
-        is DateClientParam -> SortTableInfo.DATE_SORT_TABLE_INFO
+        is StringClientParam -> listOf(SortTableInfo.STRING_SORT_TABLE_INFO)
+        is NumberClientParam -> listOf(SortTableInfo.NUMBER_SORT_TABLE_INFO)
+        is DateClientParam ->
```
jingtang10, line 68: actually add a comment here:
```suggestion
        // The DateClientParam maps to two index tables (Date without timezone info and DateTime with timezone info). Any data field in any resource will only have index records in one of the two tables. So we simply sort by both in the SQL query.
        is DateClientParam ->
```

jingtang10 (changes requested):

engine/src/main/java/com/google/android/fhir/search/MoreSearch.kt
ekigamba: I have made the changes as requested. However, I think that the `forEachIndexed` can be confusing since it makes it seem as if the engine allows multiple sort params. It is also not clear that the are implemented for an exception. The for loops feel like generic solutions while they help fix the issue for 1 exception.

engine/src/test/java/com/google/android/fhir/search/SearchTest.kt
ekigamba: yes

engine/src/main/java/com/google/android/fhir/search/MoreSearch.kt
ekigamba: Yes

jingtang10 (approved):
thanks @ekigamba

engine/src/main/java/com/google/android/fhir/search/MoreSearch.kt
ekigamba: @jingtang10

PR #1393, opened 2022-05-13
Change database insert to insertOrUpdate, change insert onConflictStrategy from REPLACE to IGNORE

### LINKED ISSUE #1172: Stop using OnConflictStrategy.REPLACE in the database
I would actually recommend we go with a local UUID instead as the key (can even ignore resource type). We can have a server id in addition to that  to be used for server communications. FYI room 2.4 adds out of the box support for UUID (and saves it as a byte array). And for conflcits, we can simply ignore, it is super super unlikely to happen :)

Also, keep in mind, sqlite ID (which i assume it referes to the auto-generated primary key), is a bit tricky to use. More specifically, if we ever do "insert(onConflict=REPLace)", sqlite will "Delete" the existing item and insert the new one, with a new id. this also means all fkey relationships will be deleted.  (not different for UUID except it wouldn't be changing so with defferred fkey checks, we can handle that)
Hence, we should also make sure to never use `onConflict=REPLACE` (instead implement a custom upsert in a transaction).

_Originally posted by @yigit in https://github.com/google/android-fhir/issues/98#issuecomment-1023685036_

wantingzhang77:

Fixes #1172

**Description**
Change the database insert onConflictStrategy from "REPLACE" to "IGNORE"

**Alternative(s) considered**

**Type**
Choose one: Code health 

**Checklist**


engine/src/main/java/com/google/android/fhir/db/impl/dao/ResourceDao.kt
```diff
@@ -81,35 +82,126 @@ internal abstract class ResourceDao {
     return resources.map { resource -> insertResource(resource) }
   }
 
-  @Insert(onConflict = OnConflictStrategy.REPLACE)
-  abstract suspend fun insertResource(resource: ResourceEntity)
+  // @Insert(onConflict = OnConflictStrategy.REPLACE)
+  @Insert(onConflict = OnConflictStrategy.IGNORE)
+  abstract suspend fun insertResourceEntity(resourceEntity: ResourceEntity): Long
```
yigit, line 87: i'm guessing the goal here is to never call these methods directly but instead call the new methods that delegate.
if so, lets make these protected so they cannot be called from outside.

yigit (changes requested):
can we also add some tests that would fail w/o this change so that if in the future someone reverts this, tests will catch?

wantingzhang77:
Hi @yigit 
I am seeing testing error from the test of createAll_shouldCreateResource(). The error message is:
```
com.almworks.sqlite4java.SQLiteException: [787] DB[1] reset [FOREIGN KEY constraint failed]
	at com.almworks.sqlite4java.SQLiteConnection.throwResult(SQLiteConnection.java:1436)
	at com.almworks.sqlite4java.SQLiteConnection.throwResult(SQLiteConnection.java:1409)
	at com.almworks.sqlite4java.SQLiteConnection.cacheStatementHandle(SQLiteConnection.java:1338)
	at com.almworks.sqlite4java.SQLiteConnection.access$900(SQLiteConnection.java:51)
	at com.almworks.sqlite4java.SQLiteConnection$CachedController.dispose(SQLiteConnection.java:1733)
	at com.almworks.sqlite4java.SQLiteStatement.dispose(SQLiteStatement.java:187)
	at org.robolectric.shadows.ShadowLegacySQLiteConnection$Connections$5.call(ShadowLegacySQLiteConnection.java:551)
	at org.robolectric.shadows.ShadowLegacySQLiteConnection$Connections$5.call(ShadowLegacySQLiteConnection.java:548)
	at java.base/java.util.concurrent.FutureTask.run(FutureTask.java:264)
	at java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1128)
	at java.base/java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:628)
	at java.base/java.lang.Thread.run(Thread.java:829)

```
I don't quit understand it. Is this some concurrent related issue? Do you think I missed anything in the code changing the "insert" to "insertOrUpdate".

engine/src/main/java/com/google/android/fhir/db/impl/dao/ResourceDao.kt
wantingzhang77: > can we also add some tests that would fail w/o this change so that if in the future someone reverts this, tests will catch?

Room use `onConflictStrategy` to deal with conflict, and conflict happens only when 2 rows has the same "PrimaryKey". Now the PrimaryKey is auto generated using `@PrimaryKey(autoGenerate = true) val id: Long,` and it is hard to create test cases of 2 entities with same PrimaryKey.

In my opinion, the conflict won't happen in current FHIR Library with auto-gen PrimaryKey unless if there's any weird glitch in the system. But change "REPLACE" to "IGNORE" is a good practice for database manipulation. So maybe we don't need extra test for this? I can add comment in the code saying REPLACE should be avoided in room incase of any further change.

aditya-07:
> Hi @yigit I am seeing testing error from the test of createAll_shouldCreateResource(). The error message is:
> 
> ```
> com.almworks.sqlite4java.SQLiteException: [787] DB[1] reset [FOREIGN KEY constraint failed]
> 	at com.almworks.sqlite4java.SQLiteConnection.throwResult(SQLiteConnection.java:1436)
> 	at com.almworks.sqlite4java.SQLiteConnection.throwResult(SQLiteConnection.java:1409)
> 	at com.almworks.sqlite4java.SQLiteConnection.cacheStatementHandle(SQLiteConnection.java:1338)
> 	at com.almworks.sqlite4java.SQLiteConnection.access$900(SQLiteConnection.java:51)
> 	at com.almworks.sqlite4java.SQLiteConnection$CachedController.dispose(SQLiteConnection.java:1733)
> 	at com.almworks.sqlite4java.SQLiteStatement.dispose(SQLiteStatement.java:187)
> 	at org.robolectric.shadows.ShadowLegacySQLiteConnection$Connections$5.call(ShadowLegacySQLiteConnection.java:551)
> 	at org.robolectric.shadows.ShadowLegacySQLiteConnection$Connections$5.call(ShadowLegacySQLiteConnection.java:548)
> 	at java.base/java.util.concurrent.FutureTask.run(FutureTask.java:264)
> 	at java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1128)
> 	at java.base/java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:628)
> 	at java.base/java.lang.Thread.run(Thread.java:829)
> ```
> 
> I don't quit understand it. Is this some concurrent related issue? Do you think I missed anything in the code changing the "insert" to "insertOrUpdate".

Docs in the `DatabaseImplTest` suggest that the Robolectric's SQLite implementation does not match Android's and hence it should be run as Integration test. Maybe moving `FhirEngineImplTest` to instrumentation tests may help with this issue.

omarismail94:
Closing this PR:
https://github.com/google/android-fhir/issues/1172#issuecomment-1337082049

Issue #1427, opened 2022-05-27
questionnaireStateFlow not emitting as expected


joiskash:
**Describe the bug**
The questionnaireStateFlow is a combination of two flows 

1. Modification count
2. pageFlow

The expectation is that every time modificationCount changes i.e. every time an answer changes due to user action, the combination of the flow should trigger resulting in getQuestionnaireState emitting a new state that is continuously being collected by the QuestionnaireFragment.
 
We have noticed that the state is collected only when pageFlow changes and not when modification count changes.

**Component**
SDC library

**To Reproduce**
1. Open a questionnaire in debugging mode
4. Add a break point where `adpater.submitList(list)` is called.
5. Make a change in the any answer
6. The break point will not get triggered
7. Add a break point in getQuestionnaireState
8. Make a change in any answer
9. The Break point will trigger

**Expected behavior**
Combination of two flows should result in an emit operation when either of the flows change.

This issue is linked https://github.com/google/android-fhir/issues/227

joiskash:
According to  **Strong equality-based conflation** section in https://kotlinlang.org/api/kotlinx.coroutines/kotlinx-coroutines-core/kotlinx.coroutines.flow/-state-flow/

> Values in state flow are conflated using [Any.equals](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin/-any/equals.html) comparison in a similar way to [distinctUntilChanged](https://kotlinlang.org/api/kotlinx.coroutines/kotlinx-coroutines-core/kotlinx.coroutines.flow/distinct-until-changed.html) operator. It is used to conflate incoming updates to [value](https://kotlinlang.org/api/kotlinx.coroutines/kotlinx-coroutines-core/kotlinx.coroutines.flow/-mutable-state-flow/value.html) in [MutableStateFlow](https://kotlinlang.org/api/kotlinx.coroutines/kotlinx-coroutines-core/kotlinx.coroutines.flow/-mutable-state-flow/index.html) and to suppress emission of the values to collectors when new value is equal to the previously emitted one. State flow behavior with classes that violate the contract for [Any.equals](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin/-any/equals.html) is unspecified.

Since QuestionnaireItemViewItem is just a wrapper around the questionnaireResponse items, the new state is always equal to the old one , even after an answer is changed. Unless there is a new enabled question or page change, the questionnairestateFlow will not emit. 

cc: @jingtang10 @aurangzaibumer

PR #1301, opened 2022-04-18
Implement the `entryMode` extension

aurangzaibumer:
**IMPORTANT: All PRs must be linked to an issue (except for extremely trivial and straightforward changes).**

Issue# [1272](https://github.com/google/android-fhir/issues/1272)

**Description**
Clear and concise code change description. 

**Alternative(s) considered**
Have you considered any alternatives? And if so, why have you chosen the approach in this PR?

**Type**
Choose one: Feature

**Screenshots/Video**

https://user-images.githubusercontent.com/35099184/170076662-ca522aa3-6ce5-48ed-96bc-85a65ec0938e.mp4

**Checklist**
- [x] I have read and acknowledged the [Code of conduct](https://github.com/google/android-fhir/blob/master/docs/CODE_OF_CONDUCT.md).
- [x] I have read the [Contributing](https://github.com/google/android-fhir/wiki/Contributing) page.
- [x] I have signed the Google [Individual CLA](https://cla.developers.google.com/about/google-individual), or I am covered by my company's [Corporate CLA](https://cla.developers.google.com/about/google-corporate ).
- [x] I have discussed my proposed solution with code owners in the linked issue(s) and we have agreed upon the general approach.
- [x] I have run `./gradlew spotlessApply` and `./gradlew spotlessCheck` to check my code follows the style guide of this project.
- [x] I have run `./gradlew check` and `./gradlew connectedCheck` to test my changes locally.
- [x] I have built and run the demo app(s) to verify my change fixes the issue and/or does not break the demo app(s).

PR #1468, opened 2022-06-22
Migrating Validation Result logic from View factory to QuestionnaireViewModel

aurangzaibumer:

**Description**

**Alternative(s) considered**

**Type**
Choose one: Improvement 

**Checklist**


aurangzaibumer:
@jingtang10 can you please review. As discussed last week I had created a separate PR for the validationResult changes, removed validation being instantiated on View Factory.

joiskash (changes requested):
Looks good to me! Just a small comment, Is there a unit test you can write to verify that validation is takes place in getQuestionnaireState function?

aurangzaibumer:
I tried to fix some of the unit tests by triggering validation for the first time in the viewmodel (in bind method), the problem is we need to fix https://github.com/google/android-fhir/issues/1427 first so on each interaction (when onAnswerChanged calls) there will be a new list of questionnaireItemViewItem along with the validationResult that gets generated by the View Model (in the getQuestionnaireState method) and we need to refresh the list so new validationResult can be reflected on the views. @jingtang10 @f-odhiambo

datacapture/src/test/java/com/google/android/fhir/datacapture/views/QuestionnaireItemAutoCompleteViewHolderFactoryTest.kt
```diff
@@ -194,39 +194,6 @@ class QuestionnaireItemAutoCompleteViewHolderFactoryInstrumentedTest {
       .isNull()
   }
 
-  @Test
```
aurangzaibumer, line 197: Removed just for checking the PR gateways, will re-add

aurangzaibumer:
> With the patch in #1486 collect gets triggered as expected. The two issues raised by @aurangzaibumer are , please add any details that I missed:
> 
> 1. How do we trigger the validation result to be displayed? Currently since the QuestionnaireItemViewItem is refers to the questionnaireResponse, the diffUtil will return a true for areItemsSame and areContentsSame.
> 2. Since we create new questionnaireItemViewItems the modified field gets overwritten.
> 
> cc : @jingtang10

Thanks @joiskash, you have covered everything. So these are the issues i'm currently facing which is kind of a blocker to https://github.com/google/android-fhir/pull/1301 as well because that PR needs a validation in order to work for a PRIOR_EDIT Pagination (one of the entry mode). So we need to have keep the state of the **modified** field and to reflect the ValidationResult (that was captured from the QuestionnaireViewModel) to the UI level without refreshing the whole screen

aurangzaibumer:
1468postfix

Tried adding a dedicated answer property to questionnaireitemviewitem and compared the new answer on QuestionnaireItemViewItem equal deeps method as mentioned below

`fun equalsDeep(other: QuestionnaireItemViewItem): Boolean {
    return this.questionnaireItem.equalsDeep(other.questionnaireItem) &&
      this.questionnaireResponseItem.equalsDeep(other.questionnaireResponseItem)
      this.questionnaireResponseItem.equalsDeep(other.questionnaireResponseItem) && this.answer == other.answer
  }`
  
  as you can see on the GIf, we faced UI rendering issue because it was rendering the whole UI element (which is currently visible to user), also the cursor reset to index 0. This is what we had tried in order to call the bind method if any change adapter can capture.

jingtang10 (changes requested):
rebase/merge to update your pr and then revert all the license header changes.

jingtang10:
This turned out to be major rewrite of the library.

In order for questionnaire item comparison to actually work, the way we create `QuestionnaireItemViewItem`s has been changed.

See documentation in the `QuestionnaireItemViewItem` class. The `questionnaireResponseItem` is marked as private so that views should not hav access to it. Instead, the views should use the APIs provided to update the answers.

The ansers are then used for comparison with what's already in the questionnaire to decide how the recycler view UI should be updated.

This results in changes in the `DiffCallback` implementation in the `QuestionnaireItemAdapter.kt` file. We consider two items to be the same if they're pointing to the same questionnaire item and questionnaire response item. But we will consider the contents to be the same only if the answers and validation results are also the same. These are the actual dynamic data during user interaction with the UI.

jingtang10 (approved):

omarismail94 (commented):
Initial feedback

datacapture/src/main/java/com/google/android/fhir/datacapture/views/QuestionnaireItemViewItem.kt
```diff
@@ -17,50 +17,68 @@
...
+ * render the data item in the UI. The view SHOULD NOT mutate the data using these properties.
  *
- * @param questionnaireResponseItemChangedCallback function that should be called whenever the
- * `questionnaireResponseItemBuilder` is changed to inform the rest of the questionnaire to be
- * updated
+ * The view should use the following answer APIs to update the answer(s):
+ * - setAnswer (for single answer only)
+ * - addAnswer (for repeated answers only)
+ * - removeAnswer (for repeated answers only)
+ * - clearAnswer
```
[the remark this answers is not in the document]
omarismail94: no, this is fine!

datacapture/src/test/java/com/google/android/fhir/datacapture/views/QuestionnaireItemAutoCompleteViewHolderFactoryTest.kt
jingtang10: these test cases are no longer valid after the validation code is moved to the view model.

datacapture/src/main/java/com/google/android/fhir/datacapture/views/QuestionnaireItemViewItem.kt
```diff
@@ -17,50 +17,68 @@
...
-    get() = questionnaireResponseItem.answer.singleOrNull()
-    set(value) {
-      questionnaireResponseItem.answer.clear()
-      value?.let { questionnaireResponseItem.addAnswer(it) }
+  private var _answers:
+    MutableList<QuestionnaireResponse.QuestionnaireResponseItemAnswerComponent> =
+    questionnaireResponseItem.answer.toMutableList()
+
+  val answers
+    get() = _answers.map { it.copy() }
```
yigit, line 70: you can use a CopyOnWriteArrayList instead of copy

datacapture/src/main/java/com/google/android/fhir/datacapture/views/QuestionnaireItemViewItem.kt
```diff
@@ -17,50 +17,68 @@
...
-  /**
-   * The single answer to the [QuestionnaireResponse.QuestionnaireResponseItemComponent], or `null`
-   * if there is none or more than one answer.
-   */
-  var singleAnswerOrNull
-    get() = questionnaireResponseItem.answer.singleOrNull()
-    set(value) {
-      questionnaireResponseItem.answer.clear()
-      value?.let { questionnaireResponseItem.addAnswer(it) }
+  private var _answers:
```
yigit, line 65: val answers : List<...> = ...
    private set // so it cannot be set from outside but can be read

jingtang10: Screenshot 2022-07-14 at 18 02 36

not sure if this works :/

jingtang10: oh sorry - i tried that but wasn't sure that's what you meant - all good!

jingtang10: this is done now.

datacapture/src/main/java/com/google/android/fhir/datacapture/views/QuestionnaireItemViewItem.kt
jingtang10: done

datacapture/src/test/java/com/google/android/fhir/datacapture/QuestionnaireViewModelTest.kt
```diff
@@ -868,24 +863,72 @@ class QuestionnaireViewModelTest(
...
+      }
+    val viewModel = createQuestionnaireViewModel(questionnaire)
+    val questionnaireItemViewItemList = viewModel.getQuestionnaireItemViewItemList()
+    val questionnaireItemViewItem = questionnaireItemViewItemList[0]
+    assertThat(questionnaireItemViewItem.validationResult)
+      .isEqualTo(ValidationResult(true, listOf()))
+  }
+
+  @Test
+  fun questionnaireItemViewItemList_shouldValidateModifiedQuestions() = runBlocking {
```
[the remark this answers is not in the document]
jingtang10: this is fixed now.

omarismail94 (approved):
This change is   👌👌👌👌👌👌👌👌👌

datacapture/src/main/java/com/google/android/fhir/datacapture/QuestionnaireViewModel.kt
```diff
@@ -393,6 +391,34 @@ internal class QuestionnaireViewModel(application: Application, state: SavedStat
...
+  /**
+   * Returns a [QuestionnaireResponse.QuestionnaireResponseItemComponent] with the given `linkId`.
+   */
+  private fun getQuestionnaireResponseItem(
+    linkId: String
+  ): QuestionnaireResponse.QuestionnaireResponseItemComponent? {
+    return questionnaireResponse
+      .item
+      .map { it.getQuestionnaireResponseItemComponent(linkId) }
+      .firstOrNull()
```
maimoonak, line 404: @jingtang10 I could not figure out the recursion here when finding the item by linkId. Could you please confirm that this method would find the item even at 10th level inside hierarchy.
```
item: {
  linkId: 1
  answer.value -> value
  answer.item -> list of another nested 'item' with answer and items
}
```

So lets say inside linkId:1 there is item linkId=1.1.1.1 at fourth level .. would it find it with this method?

jingtang10: all this function does is to look for the first item (DFS) in the questionnaire response with a certain link id. it doesn't concat the string or anything.

datacapture/src/main/java/com/google/android/fhir/datacapture/QuestionnaireViewModel.kt
```diff
@@ -393,6 +391,34 @@ internal class QuestionnaireViewModel(application: Application, state: SavedStat
...
+
+  /** Returns the current item or any descendant with the given `linkId`. */
+  private fun QuestionnaireResponse.QuestionnaireResponseItemComponent.getQuestionnaireResponseItemComponent(
+    linkId: String
+  ): QuestionnaireResponse.QuestionnaireResponseItemComponent? {
+    if (this.linkId == linkId) {
+      return this
+    }
+
+    val child = item.firstOrNull { it.linkId == linkId }
```
shoaibmushtaq25, line 415: @jingtang10 , I think @maimoonak wanted to add a comment on this method.
Here we are checking only one level down in the descendants but there is no recursion to check for all the levels down the hierarchy (Let's say if we have 10 levels of nesting). What do you think about it?

shoaibmushtaq25: okay, Thanks.

datacapture/src/main/java/com/google/android/fhir/datacapture/QuestionnaireViewModel.kt
jingtang10: see https://github.com/google/android-fhir/pull/1496

datacapture/src/main/java/com/google/android/fhir/datacapture/QuestionnaireViewModel.kt
jingtang10: see https://github.com/google/android-fhir/pull/1496

Issue #1039, opened 2022-01-14
Support for answerExpression extension


fredhersch:
**Is your feature request related to a problem? Please describe.**
Issue: #979 

When building a data collection questionnaire, there is often a need to filter the options in a drop down list based on the selected answer to an earlier list. For example: Country > State > City - see #979 

**Describe the solution you'd like**
Enable dynamic filtering of answerOptions using answerExpression with FHIRPath - see [here](http://build.fhir.org/ig/HL7/sdc/expressions.html)

In the scenario in issue #979, where there are Country > State > City, each of which the answerOptions comes from a valueSet, the following should happen:

- If country is USA
- Populate the State lists with ValueSet of states from USA
- When changing the value of the State, the City list is updated

**Describe alternatives you've considered**
Using nested set of enableWhen components

**Other information**
For Ux example of dynamically updating lists see issue #979

f-odhiambo:
Update on this ticket 
We are unable to find any example questionnaires as this expression is still in draft from FHIR side.
http://build.fhir.org/ig/HL7/sdc/StructureDefinition-sdc-questionnaire-answerExpression.html ,
http://hl7.org/fhir/R4/versions.html#std-process

Tarun-Bhardwaj:
@fredhersch @jingtang10 is there sample questionnaire that can be used in order to unblock this issue?

jingtang10:
@fredhersch i think the example with country/state/city in the original description of this issue doesn't quite match the exampel given in here http://build.fhir.org/ig/HL7/sdc/expressions.html#answerExpression for answerExpression.

From what we can see (discussed with @f-odhiambo @RaaziaTarique @maimoonak) the way this works is very similar to candidateExpression (see #1038), but answerExpression has very low maturity level. So we should definitely seek some clarity on fhir chat. 1) what's the diff between answerExp and canddiateExp and 2) with the maturity level being 0 is it worth implementing this.

RaaziaTarique:
@jingtang10 I posted the query to zulip chat and got this reply https://chat.fhir.org/#narrow/stream/179255-questionnaire/topic/expression.20library/near/282944010
<img width="869" alt="Screenshot 2022-05-19 at 9 55 25 PM" src="https://user-images.githubusercontent.com/89384079/169393817-171af148-81c2-4d6e-9f36-8f55c62c0c9a.png">
CC: @f-odhiambo @maimoonak

fredhersch:
[...]

PR #1479, opened 2022-06-28
Add X-FHIR-Query support in the FHIR Engine library

maimoonak:
Fixes part of #1039 

**Description**
Fhir engine supporting basic xfhir-query i.e. https://build.fhir.org/ig/HL7/sdc/expressions.html#fhirquery for [answer expression](https://build.fhir.org/ig/HL7/sdc/StructureDefinition-sdc-questionnaire-answerExpression.html)

**Type**
Choose one: Feature

**Checklist**


engine/src/main/java/com/google/android/fhir/search/MoreSearch.kt
```diff
@@ -45,6 +53,77 @@ internal suspend fun <R : Resource> Search.execute(database: Database): List<R>
...
+      // https://hl7.org/fhir/search.html#quantity
+      // quantity=5.4|http://unitsofmeasure.org|mg
+      // value OR value|unit OR value|system|unit
+      filterValue.split("|").let {
+        this.filter(
+          QuantityClientParam(param.name),
+          {
+            value = it.first().toBigDecimal()
+            system = if (it.size == 3) it.elementAt(1) else null
+            unit = it.lastOrNull()
```
aditya-07, line 76: We should use size to figure out unit as well since first & last value can be same incase list has a single item.

engine/src/main/java/com/google/android/fhir/FhirEngineProvider.kt
```diff
@@ -26,6 +26,8 @@ object FhirEngineProvider {
   private var fhirEngineConfiguration: FhirEngineConfiguration? = null
   private var fhirServices: FhirServices? = null
 
+  fun isInitialized() = fhirServices != null
```
aditya-07, line 29: Please remove if its not used anywhere.

engine/src/test/java/com/google/android/fhir/search/MoreSearchTest.kt
```diff
@@ -0,0 +1,238 @@
...
+      SearchParamDefinition(
+        "address-country",
+        Enumerations.SearchParamType.STRING,
+        "Patient.address.country"
+      )
+    )
+
+    val captor = argumentCaptor<StringClientParam>()
+    verify(search).sort(captor.capture(), any())
+    Truth.assertThat(captor.firstValue.paramName).isEqualTo("address-country")
```
aditya-07, line 55: Import Truth.assertThat.

engine/src/test/java/com/google/android/fhir/search/MoreSearchTest.kt
```diff
@@ -0,0 +1,238 @@
...
+    )
+
+    val captor = argumentCaptor<DateClientParam>()
+    verify(search).sort(captor.capture(), any())
+    Truth.assertThat(captor.firstValue.paramName).isEqualTo("birthdate")
+  }
+
+  @Test(expected = UnsupportedOperationException::class)
+  fun searchSort_shouldThrowUnsupportedOperationException_forUnsupportedParam() = runBlocking {
+    val search = spy(Search(ResourceType.Patient))
```
aditya-07, line 90: Do we need a spy object here? If not please remove.

engine/src/test/java/com/google/android/fhir/search/MoreSearchTest.kt
```diff
@@ -0,0 +1,238 @@
...
+        "probability",
+        Enumerations.SearchParamType.NUMBER,
+        "RiskAssessment.prediction.probability"
+      ),
+      "12"
+    )
+
+    val searchFilter = search.numberFilterCriteria.first().filters.first()
+    Truth.assertThat(searchFilter.parameter.paramName).isEqualTo("probability")
+    Truth.assertThat(searchFilter.value).isEqualTo(12.toBigDecimal())
```
aditya-07, line 116: Same comments for rest of the tests, please remove spy.

aditya-07 (changes requested):

engine/src/main/java/com/google/android/fhir/FhirEngineProvider.kt
maimoonak: this would be used in data capture in next [PR](https://github.com/google/android-fhir/pull/1480/files#diff-8cd45b205a44d10724e80221e479dbb6a10c58e7b110479f8ea2abdda793cf60R72) which is supposed to be merged after this is released

engine/src/test/java/com/google/android/fhir/search/MoreSearchTest.kt
maimoonak: yes. it is needed. because 1- we do not want to run fhirEngine -> resourceDao -> database... 2- We want to capture the arg that is being passed to fhirEngine for search and validate if it has right parameters. Otherwise we would not be able to capture the argument and flow would run database search unnecessarily.

engine/src/test/java/com/google/android/fhir/search/MoreSearchTest.kt
maimoonak: yes. it is needed. because We want to capture the arg that is being passed to fhirEngine for search and validate if it has right parameters. Otherwise we would not be able to capture the argument

maimoonak:
@aditya-07 kindly check the changes and my comments. The PR was passing codcov check but now all of a sudden it has started failing it.

aditya-07 (commented):
A small comment, otherwise the PR looks good. Should be able to approve it after this comment is discussed/addressed.

aditya-07 (approved):
LGTM !

jingtang10 (changes requested):

engine/src/main/java/com/google/android/fhir/FhirEngineProvider.kt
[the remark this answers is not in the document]
maimoonak: in tests for data capture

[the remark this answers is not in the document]
aditya-07: This is supposed to be used by the Data Capture code. https://github.com/google/android-fhir/pull/1479#discussion_r911977031

engine/src/main/java/com/google/android/fhir/search/XFhirQuery.kt
```diff
@@ -0,0 +1,29 @@
...
+package com.google.android.fhir.search
+
+import org.hl7.fhir.r4.model.ResourceType
+
+const val XFHIR_QUERY_SORT_PARAM = "_sort"
+const val XFHIR_QUERY_COUNT_PARAM = "_count"
+
+data class XFhirQuery(
+  var type: ResourceType,
+  var search: Map<String, String?>,
```
jingtang10, line 26: seems to me the value of the filter shouldn't be nullable?

```suggestion
  var search: Map<String, String>,
```

jingtang10 (changes requested):

engine/src/main/java/com/google/android/fhir/search/MoreSearch.kt
```diff
@@ -45,6 +53,77 @@ internal suspend fun <R : Resource> Search.execute(database: Database): List<R>
   return database.search(getQuery())
 }
 
+internal fun Search.filter(param: SearchParamDefinition, filterValue: String) =
```
jingtang10, line 56: can we create a new file called:

`FhirQueryTranslator.kt`

and have this API:

```
object FhirQueryTranslator {
  fun translate(fhirQuery: String): Search
}
```

and then in the `Search.kt` file, your code would be much simpler"

```
fun Search.search(fhirQuery: String): {
  return search(FhirQueryTranslator.translate(fhirQuery))
}
```

And then your test file `FhirQueryTranslatorTest.kt` will also have very clear test cases.

jingtang10 (changes requested):

engine/src/main/java/com/google/android/fhir/FhirEngineProvider.kt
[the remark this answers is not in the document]
jingtang10: Use this test rule instead: https://github.com/google/android-fhir/blob/master/testing/src/main/java/com/google/android/fhir/testing/FhirEngineProviderTestRule.kt

jingtang10 (changes requested):

engine/src/main/java/com/google/android/fhir/MoreDates.kt
```diff
@@ -17,8 +17,13 @@
...
 
 import java.time.LocalDate
+import java.time.format.DateTimeFormatter
 import java.util.Date
 
 @Suppress("DEPRECATION") // java.util.Date API used by HAPI
 internal val Date.epochDay
   get() = LocalDate.of(year + 1900, month + 1, date).toEpochDay()
+
+/** Validates if given string matches ISO_DATE format i.e. yyyy-MM-dd or yyyy-MM-dd+00:00 */
```
jingtang10, line 27: nit: this is fine. but i think we normally prefer more explicit comment.

e.g.

```suggestion
/** Returns true if given string matches ISO_DATE format i.e. "yyyy-MM-dd" or "yyyy-MM-dd+00:00", false otherwise. */
```

I also don't necessarily think this should be an extension function to String. I think extension functions should in general be related to the class it's extending. isValidDateOnly isn't a generic string utility. so i think it's better off if we left this just as a general utility.

engine/src/main/java/com/google/android/fhir/search/XFhirQueryTranslator.kt
```diff
@@ -0,0 +1,206 @@
...
+
+      sort?.forEach { sortParam ->
+        sortParam.first.toSearchParamDefinition(type).let { sort ->
+          this.applySortParam(sort, sortParam.second)
+        }
+      }
+    }
+  }
+
+  internal fun Search.applyFilterParam(param: SearchParamDefinition, filterValue: String) =
```
jingtang10, line 91: OK I konw this is a lot to ask... but...

I kinda like the idea of keeping these functions private... I really want to just expose a single API in this class which is `translate`.

if we do that, do you think it's easy to modify the tests so that they're still testing all the different edge cases sufficiently? let me know if not.

same with `applySortParams`.

engine/src/main/java/com/google/android/fhir/search/XFhirQueryTranslator.kt
```diff
@@ -0,0 +1,206 @@
+/*
```
jingtang10, line 1: move this class to a sub package called `query`. same with tests.

keep the `MoreSearch.kt` and `Search.kt` where they are and intact.

jingtang10 (changes requested):

jingtang10:
i really like the API change btw - i think the translator class is really clear

engine/src/main/java/com/google/android/fhir/search/XFhirQueryTranslator.kt
```diff
@@ -0,0 +1,206 @@
...
+  private const val XFHIR_QUERY_SORT_PARAM = "_sort"
+  private const val XFHIR_QUERY_COUNT_PARAM = "_count"
+
+  /**
+   * Translates the basic x-fhir-query defined in
+   * http://build.fhir.org/ig/HL7/sdc/expressions.html#x-fhir-query-enhancements and
+   * http://hl7.org/fhir/R4/search.html
+   *
+   * Example: Patient?active=true&gender=male&_sort=-name,gender&_count=11
+   */
```
jingtang10, line 50: why do you say "basic"? is there anything this function doesn't support right now? add comments to explain.

also, translates into what? please explain.

engine/src/main/java/com/google/android/fhir/search/XFhirQueryTranslator.kt
```diff
@@ -0,0 +1,206 @@
...
+import com.google.android.fhir.search.filter.TokenParamFilterValueInstance
+import org.hl7.fhir.r4.model.Coding
+import org.hl7.fhir.r4.model.DateTimeType
+import org.hl7.fhir.r4.model.DateType
+import org.hl7.fhir.r4.model.Enumerations
+import org.hl7.fhir.r4.model.Quantity
+import org.hl7.fhir.r4.model.Resource
+import org.hl7.fhir.r4.model.ResourceType
+
+internal object XFhirQueryTranslator {
```
jingtang10, line 40: add a brief kdoc.

don't have to repeat the documentatino for the `translate` function. but briefly explain what this class does.

jingtang10 (changes requested):

engine/src/main/java/com/google/android/fhir/search/XFhirQueryTranslator.kt
maimoonak: The change might take some effort. We can sort it out in other ticket. I have seen that private methods with @VisibleForTesting allows access to private methods in Test classes, but in sdk there are 3/4 annotation for this in different apis, and none of them worked when I tried. We can figure out to make that work as a separate issue and that way we would be able to test private methods across the project.

engine/src/main/java/com/google/android/fhir/search/XFhirQueryTranslator.kt
maimoonak: The search API in fhir is an extensive API with alot of other features as well including fhirpath support in query params, lt,gt,avg like functions into search, search over common params, search with complex params etc. Thus everything that a server has to offer. This PR addresses basic search including key-val literals, sorting (asc/desc), limit etc.

maimoonak:
@jingtang10 plz review it.

engine/src/test/java/com/google/android/fhir/impl/FhirEngineImplTest.kt
```diff
@@ -168,6 +170,43 @@ class FhirEngineImplTest {
...
+        buildPatient("2", "B", Enumerations.AdministrativeGender.FEMALE),
+        buildPatient("1", "A", Enumerations.AdministrativeGender.MALE)
+      )
+
+    fhirEngine.create(*patients.toTypedArray())
+
+    val result = fhirEngine.search("Patient?active=true&gender=male&_sort=name&_count=2")
+
+    testingUtils.assertResourceEquals(patients.elementAt(3), result.first())
+    assertThat(result.size).isEqualTo(2)
```
jingtang10, line 188: why only checking the first patient not the second?

engine/src/test/java/com/google/android/fhir/impl/FhirEngineImplTest.kt
```diff
@@ -168,6 +170,43 @@ class FhirEngineImplTest {
     testingUtils.assertResourceEquals(TEST_PATIENT_1, fhirEngine.get<Patient>(TEST_PATIENT_1_ID))
   }
 
+  @Test
+  fun search_byXFhirQueryString_shouldReturnResourceList() = runBlocking {
```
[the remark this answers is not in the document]
jingtang10: i don't really like this test case. i think the query is too complex. you're testing active, gender, sorting by name, and count...

can you split this into a test on sort, filter, and on count?

engine/src/test/java/com/google/android/fhir/search/query/XFhirQueryTranslatorTest.kt
```diff
@@ -0,0 +1,301 @@
...
+    val search = Search(ResourceType.Patient)
+
+    search.applySortParam(
+      SearchParamDefinition("birthdate", Enumerations.SearchParamType.DATE, "Patient.birthDate")
+    )
+
+    assertThat(search.sort!!.paramName).isEqualTo("birthdate")
+  }
+
+  @Test(expected = UnsupportedOperationException::class)
```
jingtang10, line 151: same - don't use this annotation for the rest of this PR

jingtang10 (changes requested):
big fan of this PR - thanks so much @maimoonak

engine/src/test/java/com/google/android/fhir/impl/FhirEngineImplTest.kt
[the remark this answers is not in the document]
maimoonak: Created 3 tests. All tests in current class are following similar naming convention. Thats why it was used

engine/src/main/java/com/google/android/fhir/search/query/XFhirQueryTranslator.kt
```diff
@@ -0,0 +1,216 @@
...
+   */
+  internal fun translate(xFhirQuery: String): Search {
+    val (type, queryStringPairs) =
+      xFhirQuery.split("?").let {
+        ResourceType.fromCode(it.first()) to it.elementAtOrNull(1)?.split("&")
+      }
+    val queryParams =
+      queryStringPairs?.mapNotNull {
+        // skip missing values like active=[missing]
+        it.split("=").takeIf { it.size > 1 }?.let { it.first() to it.elementAt(1) }
```
PallaviGanorkar, line 69: Can we add test case for missing value like "Patient?**active=&gender=**&_sort=-name,gender&_count=11"

engine/src/main/java/com/google/android/fhir/search/query/XFhirQueryTranslator.kt
```diff
@@ -0,0 +1,216 @@
...
+      }
+
+    val sort =
+      queryParams?.find { it.first == XFHIR_QUERY_SORT_PARAM }?.second?.split(",")?.map {
+        // sortParam with prefix '-' means descending order
+        it.removePrefix("-") to if (it.startsWith("-")) Order.DESCENDING else Order.ASCENDING
+      }
+    val count = queryParams?.find { it.first == XFHIR_QUERY_COUNT_PARAM }?.second?.toInt()
```
PallaviGanorkar, line 77: second?.toInt() can throw NumberFormatException if its not valid number. we may need to handle it

engine/src/test/java/com/google/android/fhir/search/query/XFhirQueryTranslatorTest.kt
```diff
@@ -0,0 +1,314 @@
...
+
+  @Test
+  fun translate_shouldAddLimit_ForQueryString() {
+    val search = translate("Patient?_count=10")
+
+    assertThat(search.count).isEqualTo(10)
+  }
+
+  @Test
+  fun translate_shouldNotAddLimit_WhenMissing() {
```
PallaviGanorkar, line 78: Can we add test case where count value is missing like "Patient?active=true&gender=male&_sort=-name,gender&_count="

engine/src/main/java/com/google/android/fhir/search/Search.kt
```diff
@@ -30,3 +31,7 @@ suspend inline fun <reified R : Resource> FhirEngine.count(init: Search.() -> Un
   search.init()
   return this.count(search)
 }
+
+suspend fun FhirEngine.search(xFhirQuery: String): List<Resource> {
```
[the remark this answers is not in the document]
maimoonak: Not possible. We do not know resource type beforehand. It has to be handled dynamically

engine/src/test/java/com/google/android/fhir/search/query/XFhirQueryTranslatorTest.kt
maimoonak: I just removed a method for a broader search as per this comment https://github.com/google/android-fhir/pull/1479#discussion_r927881928. Can add one only Patient?_count=

engine/src/main/java/com/google/android/fhir/search/query/XFhirQueryTranslator.kt
maimoonak: It should throw exception. Shall I add a test? OR throw a checked-exception?

maimoonak: resolved

engine/src/main/java/com/google/android/fhir/search/Search.kt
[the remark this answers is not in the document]
maimoonak: https://build.fhir.org/ig/HL7/sdc/examples.html#using-answerexpression-referenceresource-referenceprofile-and-lo here is the usage of this

maimoonak:
@PallaviGanorkar @jingtang10 kindly check comments and changes

engine/src/main/java/com/google/android/fhir/search/query/XFhirQueryTranslator.kt
PallaviGanorkar: We can catch it and skip count from the search param. @jingtang10   please provide you inputes as well

engine/src/test/java/com/google/android/fhir/impl/FhirEngineImplTest.kt
```diff
@@ -168,6 +170,84 @@ class FhirEngineImplTest {
...
+
+  @Test
+  fun search_byXFhirQueryString_shouldThrowFHIRException_ForUnrecognizedResourceType() {
+    val exception =
+      assertThrows(FHIRException::class.java) {
+        runBlocking {
+          fhirEngine.search("CustomResource?active=true&gender=male&_sort=name&_count=2")
+        }
+      }
+    assertThat(exception.message).isEqualTo("Unknown resource typeCustomResource")
```
PallaviGanorkar, line 239: "typeCustomResource" this can be like "type CustomResource". not sure if this testcase assert for message is success

PallaviGanorkar: @maimoonak thanks!
added few comments. Other things LGTM. Code owner can have look

engine/src/test/java/com/google/android/fhir/search/query/XFhirQueryTranslatorTest.kt
```diff
@@ -0,0 +1,340 @@
...
+import org.junit.runner.RunWith
+import org.robolectric.RobolectricTestRunner
+import org.robolectric.annotation.Config
+
+@RunWith(RobolectricTestRunner::class)
+@Config(sdk = [Build.VERSION_CODES.P])
+class XFhirQueryTranslatorTest {
+
+  @Test
+  fun translate_shouldAddDescendingSort_ForQueryString() {
```
PallaviGanorkar, line 40: can we use the new kotlin test case names in reverse single quotes: https://kotlinlang.org/docs/coding-conventions.html#names-for-test-methods

PallaviGanorkar: Thanks @maimoonak , added few comments, other things LGTM.

engine/src/test/java/com/google/android/fhir/impl/FhirEngineImplTest.kt
maimoonak: Error comes from FHIR ResourceType which sends message without space. 
Screenshot from 2022-08-02 16-56-23

maimoonak:
done @PallaviGanorkar @jingtang10

PallaviGanorkar:
> done @PallaviGanorkar @jingtang10

LGTM!

jingtang10 (approved):
This is great work! thank you @maimoonak

PR #1480, opened 2022-06-28
Answer expression extension implementation

maimoonak:
Fixes part of #1039 | Depends on #1479 

**Description**
Fhir data capture and catalog supporting basic xfhir-query i.e. https://build.fhir.org/ig/HL7/sdc/expressions.html#fhirquery for [answer expression](https://build.fhir.org/ig/HL7/sdc/StructureDefinition-sdc-questionnaire-answerExpression.html). 
Implement reference data type
Implement choice column support

**Type**
Choose one: (Feature

```
{
          "text": "Choose a Patient as Caregiver?",
          "type": "reference",
          "linkId": "1.1.0", "repeats": true,
          "extension": [
            {
              "url": "http://hl7.org/fhir/StructureDefinition/questionnaire-itemControl",
              "valueCodeableConcept": {
                "coding": [ {
                    "system": "http://hl7.org/fhir/questionnaire-item-control",
                    "code": "check-box", "display": "Checkbox"
                } ], "text": "Checkbox"
            } },
            {
              "url": "http://hl7.org/fhir/uv/sdc/StructureDefinition/sdc-questionnaire-answerExpression",
              "valueExpression": {
                "expression": "Patient?active=true",
                "language": "application/x-fhir-query"
              }
            },
            {
              "url": "http://hl7.org/fhir/uv/sdc/StructureDefinition/sdc-questionnaire-choiceColumn",
              "extension": [
                {
                  "url": "path",
                  "valueString": "name.given.first() + ' ' + name.family + ' - ' + gender + ' - ' + birthDate.toString() "
                },
                {
                  "url": "label",
                  "valueString": "name"
                },
                {
                  "url": "forDisplay",
                  "valueBoolean": true
                }
              ]
            }
          ]
        }
```

**Checklist**


datacapture/src/main/java/com/google/android/fhir/datacapture/QuestionnaireItemAdapter.kt
```diff
@@ -136,6 +137,7 @@ internal class QuestionnaireItemAdapter(
       QuestionnaireItemType.CHOICE -> getChoiceViewHolderType(questionnaireItemViewItem)
       QuestionnaireItemType.DISPLAY -> QuestionnaireItemViewHolderType.DISPLAY
       QuestionnaireItemType.QUANTITY -> QuestionnaireItemViewHolderType.QUANTITY
+      QuestionnaireItemType.REFERENCE -> QuestionnaireItemViewHolderType.REFERENCE
```
jingtang10, line 140: Check the function below `getChoiceViewHolderType`, we could probably reuse the function to decide how to present the options that are references to other FHIR resources.

The `reference` here is about what are the answer options, not necessarily how they should be presented.

datacapture/src/main/java/com/google/android/fhir/datacapture/MoreQuestionnaireItemComponents.kt
```diff
@@ -275,6 +288,83 @@ fun QuestionnaireResponse.QuestionnaireResponseItemComponent.addNestedItemsToAns
...
+        .also { this.answerOption.addAll(it) }
+    }
+}
+
+/** Populates answer options from given [Expression] with [QuestionnaireResponse] as base. */
+internal fun Questionnaire.QuestionnaireItemComponent.populateAnswerOptions(
+  expression: Expression,
+  questionnaireResponse: QuestionnaireResponse
+) =
+  fhirPathEngine
```
jingtang10, line 363: think about moving the fhir path engine evaluate logic to the caller. and keep this function strictly "util".

datacapture/src/main/java/com/google/android/fhir/datacapture/MoreQuestionnaireItemComponents.kt
```diff
@@ -275,6 +288,83 @@ fun QuestionnaireResponse.QuestionnaireResponseItemComponent.addNestedItemsToAns
...
+ * Apply and add each choice-column mapping to answer options
+ * https://build.fhir.org/ig/HL7/sdc/StructureDefinition-sdc-questionnaire-choiceColumn.html
+ *
+ * Control the information displayed in list. With reference it allows selection of fields from the
+ * resource for display and reference
+ */
+internal fun Questionnaire.QuestionnaireItemComponent.populateAnswerOptions(
+  dataList: List<Resource>
+) {
+  this.choiceColumn
```
jingtang10, line 336: can you just early return?

```
if (this.choiceColumn == null) {
  return
}
```

datacapture/src/main/java/com/google/android/fhir/datacapture/MoreQuestionnaireItemComponents.kt
```diff
@@ -275,6 +288,83 @@ fun QuestionnaireResponse.QuestionnaireResponseItemComponent.addNestedItemsToAns
...
+      }
+    }
+    ?.also { choiceColumns ->
+      dataList
+        .map { data ->
+          Reference().apply {
+            reference = "${data.resourceType}/${data.logicalId}"
+            display =
+              choiceColumns.filter { it.forDisplay }.joinToString(" ") {
+                fhirPathEngine.evaluateToString(data, it.path)
```
jingtang10, line 349: make this a separate function?

datacapture/src/main/java/com/google/android/fhir/datacapture/QuestionnaireViewModel.kt
```diff
@@ -364,15 +384,17 @@ internal class QuestionnaireViewModel(application: Application, state: SavedStat
     return questionnaireItemList
       .asSequence()
       .flatMap { questionnaireItem ->
-        var questionnaireResponseItem = questionnaireItem.createQuestionnaireResponseItem()
+        questionnaireItem.answerExpression?.let {
+          runBlocking { loadAnswerExpressionOptions(questionnaireItem, it) }
+        }
+
         // If there is an enabled questionnaire response available then we use that. Or else we
         // just use an empty questionnaireResponse Item
```
jingtang10, line 392:
```suggestion
        // just use an empty questionnaire response item.
```

datacapture/src/main/java/com/google/android/fhir/datacapture/enablement/EnablementEvaluator.kt
```diff
@@ -16,16 +16,12 @@
 
 package com.google.android.fhir.datacapture.enablement
 
-import ca.uhn.fhir.context.FhirContext
-import ca.uhn.fhir.context.FhirVersionEnum
-import ca.uhn.fhir.context.support.DefaultProfileValidationSupport
 import com.google.android.fhir.compareTo
 import com.google.android.fhir.datacapture.enableWhenExpression
+import com.google.android.fhir.datacapture.mapping.ResourceMapper.fhirPathEngine
```
jingtang10, line 21: can we have a separate fhir path engine provider please

datacapture/src/main/java/com/google/android/fhir/datacapture/views/QuestionnaireItemDropDownViewHolderFactory.kt
```diff
@@ -57,7 +57,11 @@ internal object QuestionnaireItemDropDownViewHolderFactory :
         val adapter =
           ArrayAdapter(context, R.layout.questionnaire_item_drop_down_list, answerOptionString)
         autoCompleteTextView.setText(
-          questionnaireItemViewItem.answers.singleOrNull()?.valueCoding?.display ?: ""
+          questionnaireItemViewItem.answers.singleOrNull()?.let {
+            if (it.hasValueReference()) it.valueReference.display
```
jingtang10, line 61: use `when`

jingtang10 (changes requested):

datacapture/src/main/java/com/google/android/fhir/datacapture/MoreQuestionnaireItemComponents.kt
maimoonak: done for base type. however, for resource type not possible to move this out as expression is specified with in choice column

maimoonak:
@jingtang10 kindly review the changes

jingtang10 (changes requested):

catalog/src/main/java/com/google/android/fhir/catalog/ComponentListViewModel.kt
```diff
@@ -71,6 +71,11 @@ class ComponentListViewModel(application: Application, private val state: SavedS
       "modal-questionnaire.json",
       "modal_with_validation_questionnaire.json"
     ),
+    REFERENCE(
+      R.drawable.ic_group_1278,
+      R.string.component_name_reference,
+      "dropdown-reference-questionnaire.json"
+    ),
```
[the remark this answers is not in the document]
jingtang10: i feel this feature "answer expression" isn't super easy to demonstrate in the catalog app.  without context, this doesn't seem that different from the dropdown widget.

the interesting part of this feature really lies within the api integration with fhir engine. and therefore the best way to showcase this is through extensive and clear documentation (fyi @williamito) rather than catalog app.

datacapture/src/main/java/com/google/android/fhir/datacapture/MoreQuestionnaireItemComponents.kt
```diff
@@ -275,6 +288,79 @@ fun QuestionnaireResponse.QuestionnaireResponseItemComponent.addNestedItemsToAns
...
+          choiceColumns
+            .filter { it.forDisplay }
+            .map { it.path }
+            .let { evaluateToDisplay(it, data) }
+            .also { display = it }
+        }
+      } else data.castToType(data)
+    }
+    .map { Questionnaire.QuestionnaireItemAnswerOptionComponent(it) }
+    .also { this.answerOption.also { it.clear() }.addAll(it) }
```
jingtang10, line 361: i sort of see the `questionnaire` in `QuestionnaireViewModel.kt` as an immutable object - meaning that we should not modify it. in the `QuestionnaireItemViewItem`s, we reference the `QuestionnnaireItem`s from that `questionnaire` object in `QuestionnaireViewModel.kt`. So what you're doing here, is directly modifying the `questionnaire` object.

we need to make sure there's no unintended consequences of modifying this `answerOption` field. for example, will the questionnaire be reused? or when the screen refreshes, will these values be stale etc?

but i feel like a safer way to do this is just not modify the `answerOption`.

jingtang10 (changes requested):
the test failure for modle `demo` is likely caused by the fact in this PR a new dependency from the `datacapture` module to the public maven release of fhir engine is added.

the `demo` module already has a local dependency on the `engine` module. and now it has another dependency to the same set of classes through the `datacapture` module. this causes the build error.

to properly fix this, i think we probably should create 2 build flavours: local build that uses local modules, and release build that uses everything on maven as dependency.

but that can probably be done in a followup pr.

thanks!

maimoonak:
@jingtang10 changes done. Kindly review

datacapture/build.gradle.kts
```diff
@@ -61,6 +68,10 @@ dependencies {
   coreLibraryDesugaring(Dependencies.desugarJdkLibs)
 
   implementation(Dependencies.androidFhirCommon)
+
+  releaseImplementation(Dependencies.androidFhirEngine)
+  debugImplementation(project(":engine"))
```
jingtang10, line 73: order these alphabetically. i.e. break them into separate blocks.

datacapture/build.gradle.kts
```diff
@@ -85,6 +96,7 @@ dependencies {
   testImplementation(Dependencies.mockitoKotlin)
   testImplementation(Dependencies.robolectric)
   testImplementation(Dependencies.truth)
+  testImplementation(project(":testing"))
   androidTestImplementation(Dependencies.Espresso.espressoCore)
```
jingtang10, line 100: move this line up to the androidTestImplementation block

datacapture/src/main/java/com/google/android/fhir/datacapture/MoreQuestionnaireItemComponents.kt
```diff
@@ -333,6 +346,77 @@ fun QuestionnaireResponse.QuestionnaireResponseItemComponent.addNestedItemsToAns
...
+        ChoiceColumn(
+          path = it.find { it.url == "path" }!!.value.asStringValue(),
+          label = it.find { it.url == "label" }?.value?.asStringValue(),
+          forDisplay =
+            it.any { it.url == "forDisplay" && it.castToBoolean(it.value).booleanValue() }
+        )
+      }
+    }
+
+data class ChoiceColumn(val path: String, val label: String?, val forDisplay: Boolean)
```
jingtang10, line 380: can you try to mark this as internal?

datacapture/src/main/java/com/google/android/fhir/datacapture/MoreQuestionnaireItemComponents.kt
```diff
@@ -333,6 +346,77 @@ fun QuestionnaireResponse.QuestionnaireResponseItemComponent.addNestedItemsToAns
...
+ * Control the information displayed in list.
+ * - With reference it allows selection of fields from the resource for display and reference
+ * - With other types it adds the options as is
+ */
+internal fun Questionnaire.QuestionnaireItemComponent.extractAnswerOptions(
+  dataList: List<Base>
+): List<Questionnaire.QuestionnaireItemAnswerOptionComponent> {
+  if (dataList.isEmpty()) return emptyList()
+
+  if (this.type != Questionnaire.QuestionnaireItemType.REFERENCE && dataList.first().isResource) {
```
jingtang10, line 397:
```suggestion
  if (this.type != Questionnaire.QuestionnaireItemType.REFERENCE && dataList.all { it.isResource } ) {
```

datacapture/src/main/java/com/google/android/fhir/datacapture/MoreQuestionnaireItemComponents.kt
```diff
@@ -333,6 +346,77 @@ fun QuestionnaireResponse.QuestionnaireResponseItemComponent.addNestedItemsToAns
...
+          this@extractAnswerOptions.choiceColumn
+            ?.filter { it.forDisplay }
+            ?.map { it.path }
+            ?.let { evaluateToDisplay(it, data) }
+            ?.also { display = it }
+        }
+      } else data.castToType(data)
+    }
+    .map { Questionnaire.QuestionnaireItemAnswerOptionComponent(it) }
+}
```
jingtang10, line 418:
```
when {
  (question type is reference) -> {
    check(data list must be resources)
    return map...
  }
  else -> {
    check(data list can't be resources)
    return map...
  }
```

jingtang10 (changes requested):

datacapture/src/main/java/com/google/android/fhir/datacapture/QuestionnaireViewModel.kt
```diff
@@ -411,6 +419,42 @@ internal class QuestionnaireViewModel(application: Application, state: SavedStat
...
   }
 
+  // TODO persist previous answers incase options are changing and new list does not have selected
+  // answer
+  @PublishedApi
+  internal suspend fun resolveAnswerExpression(
+    item: Questionnaire.QuestionnaireItemComponent
+  ): List<Questionnaire.QuestionnaireItemAnswerOptionComponent> {
+    // Check cache first for database queries
+    val answerExpression = item.answerExpression!!
```
jingtang10, line 429: i feel this can be softened a bit - it wouldn't be idea if someone else unknowningly calls this function and gets a crash.

maybe `item.answerExpression?.let { ... }` and then change the return type to nullable as well.

at the call site, you'll add !! at the end of the return value.

jingtang10 (approved):

datacapture/src/main/java/com/google/android/fhir/datacapture/QuestionnaireViewModel.kt
```diff
@@ -277,6 +282,9 @@ internal class QuestionnaireViewModel(application: Application, state: SavedStat
   private val answerValueSetMap =
     mutableMapOf<String, List<Questionnaire.QuestionnaireItemAnswerOptionComponent>>()
 
+  private val answerExpressionMap =
```
jingtang10, line 285: needs documentation. is this always up-to-date? when do you update this? add these to the documentation.

datacapture/src/main/java/com/google/android/fhir/datacapture/QuestionnaireViewModel.kt
```diff
@@ -411,6 +419,42 @@ internal class QuestionnaireViewModel(application: Application, state: SavedStat
...
 
+  // TODO persist previous answers incase options are changing and new list does not have selected
+  // answer
+  @PublishedApi
+  internal suspend fun resolveAnswerExpression(
+    item: Questionnaire.QuestionnaireItemComponent
+  ): List<Questionnaire.QuestionnaireItemAnswerOptionComponent> {
+    // Check cache first for database queries
+    val answerExpression = item.answerExpression!!
+    if (answerExpression.isXFhirQuery && answerExpressionMap.contains(answerExpression.expression)
```
jingtang10, line 430: ok so the assumption is that if the expression is x fhir query the value will never change?

is this always true? can you have an x fhir query that refers to an id that might be changed by another answer?

datacapture/src/test/java/com/google/android/fhir/datacapture/MoreQuestionnaireItemComponentsTest.kt
```diff
@@ -1277,6 +1279,220 @@ class MoreQuestionnaireItemComponentsTest {
...
+                  addExtension(Extension("forDisplay", BooleanType(true)))
+                }
+              )
+          }
+        )
+
+    assertThat(questionItem.itemFirstRep.choiceColumn!!)
+      .contains(ChoiceColumn("code", "CODE", false))
+    assertThat(questionItem.itemFirstRep.choiceColumn!!)
+      .contains(ChoiceColumn("display", "DESCRIPTION", true))
```
jingtang10, line 1358: make this stronger by using `containsExactly`

datacapture/src/test/java/com/google/android/fhir/datacapture/MoreQuestionnaireItemComponentsTest.kt
```diff
@@ -1277,6 +1279,220 @@ class MoreQuestionnaireItemComponentsTest {
...
+              nameFirstRep.addGiven("Jane")
+            }
+          )
+        )
+        .map { it.valueReference }
+
+    assertThat(answers.first().display).isEqualTo("John Doe")
+    assertThat(answers.first().reference).isEqualTo("Patient/1234")
+    assertThat(answers.last().display).isEqualTo("Jane Doe")
+    assertThat(answers.last().reference).isEqualTo("Patient/5678")
```
jingtang10, line 1463: use containsExactly. or at least you'll need to assert the size of this list

datacapture/src/main/java/com/google/android/fhir/datacapture/views/QuestionnaireItemViewItem.kt
```diff
@@ -61,12 +62,19 @@ data class QuestionnaireItemViewItem(
...
+    ) -> Unit,
   private val resolveAnswerValueSet:
     suspend (String) -> List<Questionnaire.QuestionnaireItemAnswerOptionComponent> =
-      {
-    emptyList()
-  },
+    {
+      emptyList()
+    },
+  private val resolveAnswerExpression:
```
jingtang10, line 72: we have done a huge amount of refactoring to make the data flow better, but i've just realised these callbacks do not seem to be consistent with what we've done.

the source of the truth should always be generated by the view model. is it possible to not have this callback but instead use a list which is populated by the view model?

jingtang10 (changes requested):

datacapture/src/main/java/com/google/android/fhir/datacapture/MoreQuestionnaireItemComponents.kt
```diff
@@ -368,6 +382,85 @@ fun QuestionnaireResponse.QuestionnaireResponseItemComponent.addNestedItemsToAns
...
+internal fun Questionnaire.QuestionnaireItemComponent.extractAnswerOptions(
+  dataList: List<Base>
+): List<Questionnaire.QuestionnaireItemAnswerOptionComponent> {
+  if (dataList.isEmpty()) return emptyList()
+
+  val options: List<Type> =
+    dataList.map { data ->
+      when (this.type) {
+        Questionnaire.QuestionnaireItemType.REFERENCE -> {
+          require(dataList.all { it.isResource }) {
```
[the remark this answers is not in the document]
maimoonak: this was discussed in last review. the require needs to assert that each value is fulfilling the criteria. This assertion is dependent on the type of widget. This does not hold true for other branches hence can be only inside this block

datacapture/src/main/java/com/google/android/fhir/datacapture/MoreQuestionnaireItemComponents.kt
```diff
@@ -368,6 +382,85 @@ fun QuestionnaireResponse.QuestionnaireResponseItemComponent.addNestedItemsToAns
...
+/**
+ * Apply and add each choice-column mapping to answer options
+ * https://build.fhir.org/ig/HL7/sdc/StructureDefinition-sdc-questionnaire-choiceColumn.html
+ *
+ * Control the information displayed in list.
+ * - With reference it allows selection of fields from the resource for display and reference
+ * - With other types it adds the options as is
+ */
+internal fun Questionnaire.QuestionnaireItemComponent.extractAnswerOptions(
+  dataList: List<Base>
```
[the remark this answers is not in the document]
maimoonak: this can be Resource or Coding as well. The choice column can handle complex types https://build.fhir.org/ig/HL7/sdc/StructureDefinition-sdc-questionnaire-choiceColumn.html

datacapture/src/main/java/com/google/android/fhir/datacapture/utilities/FhirPathUtil.kt
```diff
@@ -0,0 +1,31 @@
...
+import org.hl7.fhir.r4.hapi.ctx.HapiWorkerContext
+import org.hl7.fhir.r4.model.Resource
+import org.hl7.fhir.r4.utils.FHIRPathEngine
+
+internal val fhirPathEngine: FHIRPathEngine =
+  with(FhirContext.forCached(FhirVersionEnum.R4)) {
+    FHIRPathEngine(HapiWorkerContext(this, this.validationSupport))
+  }
+
+internal fun evaluateToDisplay(expressions: List<String>, data: Resource) =
```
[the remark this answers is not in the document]
maimoonak: For Reference items, if multiple columns are marked with 'forDisplay', the display value used will be a space-separated concatenation of all columns so-marked.
https://build.fhir.org/ig/HL7/sdc/StructureDefinition-sdc-questionnaire-choiceColumn.html

datacapture/src/main/java/com/google/android/fhir/datacapture/QuestionnaireViewModel.kt
maimoonak: Current implementation does not allow it to change value. However, the advance implement suggests to evaluate the fhirpath before any evaluation https://build.fhir.org/ig/HL7/sdc/expressions.html#x-fhir-query-enhancements.

datacapture/src/test/java/com/google/android/fhir/datacapture/MoreQuestionnaireItemComponentsTest.kt
maimoonak: containsExactly does not work with fhir Type. Added list size assertion

datacapture/src/main/java/com/google/android/fhir/datacapture/views/QuestionnaireItemViewItem.kt
maimoonak: Current implementation that is handling answer options using valueset also uses similar pattern. We would have to refactor both to accommodate this. Also this callback load the options in right way with using runBlocking. We need to implement a mechanism to load options in background with a progress bar and from the right source (answerOptions, valueSet, OR from answerOption etc). We can create a new ticket and add it as discussion item as this would need an overhaul.

jingtang10: please create a ticket

datacapture/src/main/java/com/google/android/fhir/datacapture/utilities/FhirPathUtil.kt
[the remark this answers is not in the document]
jingtang10: nice one - can you add that to the documentation?

datacapture/src/main/java/com/google/android/fhir/datacapture/MoreQuestionnaireItemComponents.kt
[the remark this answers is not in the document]
jingtang10: wait but you still map the list... just think how many times this require is run?

i think the `when` might need to be outside of the `dataList.map`

jingtang10 (changes requested):

datacapture/src/main/java/com/google/android/fhir/datacapture/QuestionnaireViewModel.kt
maimoonak: https://github.com/google/android-fhir/issues/1658

jingtang10 (approved):

PR #1481, opened 2022-06-28
Try Kokoro as CI Tool

omarismail94:
Use Kokoro as a CI Tool. 

It is a script-based, meaning that developers can run the `build.sh` script locally to see if there PRs will build and pass when run against Kokoro.

The tool completes its run in 30 minutes, halving the time from GitHub

Also added a `--record_screen` flag to `build.sh` so that users can record their screen locally when testing.

When Kokoro runs, it always adds this flag and uploads the screen capture as part of the artefacts it uploads. The artefacts also include the build logs, the `ci-repo` directory, the `build.zip` file, and any `xml` files from testing

I have kept the GitHub Actions config for now, and think we should use both CI tools for the time being, just for people to become familiar with Kokoro. We can migrate to using only Kokoro in a few weeks.

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -40,7 +40,13 @@ subprojects {
 
 // Create a CI repository and also change versions to include the build number
 afterEvaluate {
-  val buildNumber = System.getenv("KOKORO_BUILD_ID")
+  // Check to see which CI tool is being used (Kokoro or GitHub Actions)
+  val buildNumber =
+    if (project.providers.environmentVariable("KOKORO_BUILD_ID").isPresent) {
+      System.getenv("KOKORO_BUILD_ID")
+    } else {
+      System.getenv("GITHUB_RUN_ID")
+    }
   if (buildNumber != null) {
     subprojects {
       apply(plugin = Plugins.BuildPlugins.mavenPublish)
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -40,7 +40,7 @@ subprojects {
 
 // Create a CI repository and also change versions to include the build number
 afterEvaluate {
-  val buildNumber = System.getenv("GITHUB_RUN_ID")
+  val buildNumber = System.getenv("KOKORO_BUILD_ID")
   if (buildNumber != null) {
     subprojects {
       apply(plugin = Plugins.BuildPlugins.mavenPublish)
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -40,13 +40,7 @@ subprojects {
 
 // Create a CI repository and also change versions to include the build number
 afterEvaluate {
-  // Check to see which CI tool is being used (Kokoro or GitHub Actions)
-  val buildNumber =
-    if (project.providers.environmentVariable("KOKORO_BUILD_ID").isPresent) {
-      System.getenv("KOKORO_BUILD_ID")
-    } else {
-      System.getenv("GITHUB_RUN_ID")
-    }
+  val buildNumber = System.getenv("GITHUB_RUN_ID")
   if (buildNumber != null) {
     subprojects {
       apply(plugin = Plugins.BuildPlugins.mavenPublish)
--- a/datacapture/src/main/res/layout/questionnaire_item_multi_select_dialog.xml
+++ b/datacapture/src/main/res/layout/questionnaire_item_multi_select_dialog.xml
@@ -60,7 +60,7 @@
             android:layout_width="wrap_content"
             android:layout_height="wrap_content"
             android:padding="@dimen/padding_half_size"
-            android:text="@android:string/cancel"
+            android:text="@string/cancel"
         />
 
         <Space
--- a/kokoro/README.md
+++ b/kokoro/README.md
@@ -1,0 +1,5 @@
+## Kokoro Infrastructure
+
+The files in this directory serve as plumbing for running tests under Kokoro,
+our internal CI. If there are any changes required to these config files, please
+file an issue.
--- a/kokoro/README.md
+++ b/kokoro/README.md
@@ -1,5 +1,17 @@
 ## Kokoro Infrastructure
 
 The files in this directory serve as plumbing for running tests under Kokoro,
-our internal CI. If there are any changes required to these config files, please
-file an issue.
+our internal CI tool. If there are any changes required to these config files,
+please file an issue.
+
+The [build script](./gcp_ubuntu/kokoro_build.sh) runs each time a PR is opened,
+or when a commit is pushed to an open PR, or when an a PR is merged into the
+master branch. The script runs on a Google Compute Engine machine hosted in our
+own Google Cloud project, with public access to the logs.
+
+The script downloads the dependencies it needs, compiles, builds, and unit tests
+the code, and then uses Firebase Test Lab to run instrumentation tests. Code
+coverage reports are then uploaded to Codecov, where they are displayed on the
+GitHub Pull Request/Repo Homepage.
+
+**WARNING**:Do NOT run the script on your local machine.
--- a/kokoro/gcp_ubuntu/kokoro_build.sh
+++ b/kokoro/gcp_ubuntu/kokoro_build.sh
@@ -17,12 +17,10 @@
 
 # This script should NOT be run locally.
 
-# Script to setup Kokoro to run CI tests for the Android FHIR SDK. The script
-# downloads dependencies, and for each API_LEVEL specified below, creates a new
-# emulator, then runs the ./build script.
-
-# A full build is done for API Level 30. For subsequent API levels, Gradle
-# only runs the instrumentation tests.
+# Script to run CI pipeline for the Android FHIR SDK. The script downloads the
+# dependencies it needs, compiles, builds, and unit tests the code, and then
+# uses Firebase Test Lab to run instrumentation tests. Code coverage reports are
+# then uploaded to Codecov, where they are displayed on the GitHub Pull Request.
 
 # Fail on any error.
 set -e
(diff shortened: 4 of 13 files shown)


build.gradle.kts
```diff
@@ -33,7 +33,7 @@ subprojects {
 
 // Create a CI repository and also change versions to include the build number
 afterEvaluate {
-  val buildNumber = System.getenv("GITHUB_RUN_ID")
+  val buildNumber = System.getenv("KOKORO_BUILD_ID")
```
jingtang10, line 36: add document for this

build.sh
```diff
@@ -0,0 +1,91 @@
...
+  if [[ -n "${RECORD_SCREEN}"  ]]; then
+    echo "INFO: Stopping screen recording and pulling video from device"
+    $ADB shell ls "${RECORD_DIR}"
+    $ADB pull "${RECORD_DIR}" || echo "No videos on phone"
+  fi
+}
+
+function validate_args() {
+  if [[ $# -ge 3 ]]; then
+    echo "ERROR: Invalid number of args passed."
```
jingtang10, line 40: add number of args to the error message. or the args themselves.

build.sh
```diff
@@ -0,0 +1,91 @@
...
+#   ./build.sh 30 --record_screen
+
+# Fail on any error.
+set -e
+
+function cleanup() {
+  if [[ -n "${RECORD_SCREEN}"  ]]; then
+    echo "INFO: Stopping screen recording and pulling video from device"
+    $ADB shell ls "${RECORD_DIR}"
+    $ADB pull "${RECORD_DIR}" || echo "No videos on phone"
```
jingtang10, line 34: should this be a warning or info if no video was recorded?

build.sh
```diff
@@ -0,0 +1,91 @@
...
+}
+
+function validate_args() {
+  if [[ $# -ge 3 ]]; then
+    echo "ERROR: Invalid number of args passed."
+    exit 1
+  fi
+
+  RECORD_SCREEN=""
+  if [[ $1 == "--record_screen" || $2 == "--record_screen" ]]; then
```
jingtang10, line 45: i kinda feel like this is going to generate issues when the number of args grows. how about:
1) create a function to properly handle all args, and set some variable in the script before proceeding, or
2) use `$*`?

i'm not really familiar with the idiomatic ways of writing scripts... so feel free to correct me/ignore this. but this is what i found: https://superuser.com/questions/186272/check-if-any-of-the-parameters-to-a-bash-script-match-a-string

kokoro/README.md
```diff
@@ -0,0 +1,5 @@
+## Kokoro Infrastructure
+
+The files in this directory serve as plumbing for running tests under Kokoro,
```
jingtang10, line 3: add link to public documentation of kokoro if any

kokoro/gcp_ubuntu/kokoro_build.sh
```diff
@@ -0,0 +1,136 @@
...
+  adb devices | grep emulator | cut -f1 | while read line; do adb -s $line emu kill; sleep 5; done
+
+  # Install Android image and install on emulator, then turn on emulator
+  local system_image="system-images;android-$1;google_apis;x86"
+  sdkmanager  --install ${system_image} > /dev/null
+  echo "no" | avdmanager create avd  -f --name test_avd \
+      --tag google_apis --package ${system_image}
+  sudo chown ${USER} /dev/kvm
+
+  # TODO(omarismail): remove when https://github.com/google/android-fhir/issues/1482 fixed
```
jingtang10, line 93: remove which lines?

kokoro/gcp_ubuntu/kokoro_build.sh
```diff
@@ -0,0 +1,136 @@
...
+  # Set Java version to 11
+  sudo update-java-alternatives --set "/usr/lib/jvm/java-1.11.0-openjdk-amd64"
+
+  # Install npm for spotlessApply
+  sudo npm cache clean -f
+  sudo npm install -g n
+  sudo n stable
+
+  # Setup Android CLI tools
+  wget https://dl.google.com/android/repository/commandlinetools-linux-8512546_latest.zip \
```
jingtang10, line 70: would this url change in the future? and when/if it doesn it'll break our pipeline?

jingtang10 (changes requested):

kokoro/gcp_ubuntu/kokoro_build.sh
```diff
@@ -0,0 +1,136 @@
...
+  for api_level in "${API_LEVELS[@]}"; do
+    create_emulator "${api_level}"
+    ./build.sh "${api_level}" "--record_screen"
+  done
+}
+
+function run_codecov() {
+  # Don't write secrets to the logs
+  set +x
+  curl -Os https://uploader.codecov.io/latest/linux/codecov
```
bashir2, line 121: Interesting; it seems their approach [is not bash based](https://about.codecov.io/blog/introducing-codecovs-new-uploader/) anymore.
FYI, I was just curious and checked our analytics repo; then realized that we are not uploading coverage info since last September. I think [here](https://github.com/GoogleCloudPlatform/openmrs-fhir-analytics/commit/0137c66997897ec1c5d2300a1ae1010283ef7b60) is where we stopped it (unintentionally?).

bashir2 (commented):
Thanks @omarismail94 for this, I can imagine how much work it has been to put everything together (video for screen capture is great).
I only looked at `.sh` changes and made a few comments. Nothing is blocking and all are minor.

kokoro/gcp_ubuntu/kokoro_build.sh
```diff
@@ -0,0 +1,136 @@
...
+# in the job configuration.
+export ANDROID_HOME=${HOME}/android_sdk
+export PATH=$PATH:${ANDROID_HOME}/cmdline-tools/latest/bin
+export PATH=$PATH:${ANDROID_HOME}/platform-tools
+export PATH=$PATH:${ANDROID_HOME}/emulator
+
+function zip_artifacts() {
+  mkdir -p test-results
+  zip test-results/build.zip ./*/build -r -q
+  find . -type f -regex ".*[t|androidT]est-results/.*xml" \
```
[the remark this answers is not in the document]
omarismail94: We do need to search in both folders for all XML files. In Android, each library can have two types of tests: AndroidTests, and Tests ([example](https://github.com/google/android-fhir/tree/master/engine/src). The former is for tests that run on android devices, and the other are for those that do not. When gradle runs  tests, it stores the results of the tests in the folder the test classes are part of

kokoro/gcp_ubuntu/kokoro_build.sh
omarismail94: Ah, it must have been unintentionally when we migrated to Cloud Build. I do not have access to:
https://codecov.io/github/GoogleCloudPlatform/openmrs-fhir-analytics

to see the token, but if you can give me write access on the repo, I can set this up on Cloud Build for OpenMRS

testing/build.gradle.kts
```diff
@@ -15,6 +15,11 @@ android {
     sourceCompatibility = JavaVersion.VERSION_1_8
     targetCompatibility = JavaVersion.VERSION_1_8
   }
+
+  packagingOptions {
+    resources.excludes.addAll(listOf("META-INF/ASL-2.0.txt", "META-INF/LGPL-3.0.txt"))
+  }
```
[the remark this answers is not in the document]
omarismail94: I was getting an error about duplicates of this file existing. We also have the same function in our other libs:

* https://github.com/google/android-fhir/blob/master/engine/build.gradle.kts#L76
* https://github.com/google/android-fhir/blob/master/workflow/build.gradle.kts#L57

kokoro/gcp_ubuntu/kokoro_build.sh
omarismail94: I dont think so, I looked through old URLs and they still are valid with _latest appended:

```
wget https://dl.google.com/android/repository/commandlinetools-linux-8092744_latest.zip
```

still works

build.sh
```diff
@@ -0,0 +1,91 @@
...
+    RECORD_DIR="/sdcard/Recording/"
+  fi
+
+  regex_for_api_level='^[0-9]+$'
+  if [[ $1 =~ $regex_for_api_level ]] ; then
+     API_LEVEL=$1
+  fi
+}
+
+function check_adb_exists() {
```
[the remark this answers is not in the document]
omarismail94: Resolving as this file has been removed

datacapture/src/main/res/layout/questionnaire_item_multi_select_dialog.xml
```diff
@@ -60,7 +60,7 @@
             android:layout_width="wrap_content"
             android:layout_height="wrap_content"
             android:padding="@dimen/padding_half_size"
-            android:text="@android:string/cancel"
+            android:text="@string/cancel"
```
jingtang10, line 63: `@android:string/cancel` is the correct value: https://developer.android.com/reference/android/R.string#cancel

this value should be provided by the framework not by us... we don't want to have to translate cancel to different languages (thought we're not doing that now for any of our strings).

why do you think this is failing?

jingtang10 (approved):

datacapture/src/main/res/layout/questionnaire_item_multi_select_dialog.xml
omarismail94: There are other strings that are Android-defined as well that we do not use. For example, `yes` and `no`. We use our own definition as opposed to the android one.

build.gradle.kts
```diff
@@ -40,7 +40,13 @@ subprojects {
 
 // Create a CI repository and also change versions to include the build number
 afterEvaluate {
-  val buildNumber = System.getenv("GITHUB_RUN_ID")
+  // Check to see which CI tool is being used (Kokoro or GitHub Actions)
```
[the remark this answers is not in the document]
omarismail94: I reverted it back to GitHub only. My thinking was if we decide to only use Kokoro to run the instrumentation test, then we don't even need to change anything here

PR #1486, opened 2022-07-04
Add modification count in the state flow and make it thread-safe

### LINKED ISSUE #1427: questionnaireStateFlow not emitting as expected

jingtang10:

Fixes #1427

**Description**
Add modification count in the state flow in order for the flow to trigger an update

Also make it thread-safe by calling the `update` function instead of directly incrementing the value

cc @joiskash @aurangzaibumer 

**Alternative(s) considered**
NA

**Type**
Bug fix

NA

**Checklist**

### DIFF
--- a/buildSrc/src/main/kotlin/Dependencies.kt
+++ b/buildSrc/src/main/kotlin/Dependencies.kt
@@ -1,5 +1,5 @@
 /*
- * Copyright 2021 Google LLC
+ * Copyright 2022 Google LLC
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -169,7 +169,7 @@ object Dependencies {
     }
 
     object Kotlin {
-      const val kotlinCoroutinesCore = "1.4.2"
+      const val kotlinCoroutinesCore = "1.6.2"
       const val stdlib = "1.6.10"
     }
 
--- a/datacapture/build.gradle.kts
+++ b/datacapture/build.gradle.kts
@@ -71,6 +71,7 @@ dependencies {
     exclude(module = "httpclient")
     exclude(group = "net.sf.saxon", module = "Saxon-HE")
   }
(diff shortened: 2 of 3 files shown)


joiskash (approved):
Looks good to me! @maimoonak i think this change will also solve the problem that you have in the expressions PR. Please have a look. 

cc : @nsabale7 check with your review summary pr also. I know we created another flow there as a work around

joiskash:
There seems to be a slight problem that @aurangzaibumer has raised. This change fixed the issue related to making the flow emit the new state when modification count is changed i.e. when an answer is edited. 
But on submitList, the diffUtil returns true for both areItemsSame and areContentsSame resulting in the view not getting updated. This is mainly because both old and new states refer to the same questionnaireResponseItem. In most implementations of the recycler view, I have seen that the viewItems have data that directly get displayed on the UI. So any change from UI reflects on the viewItem. In our case we use the viewItem only as a wrapper. Maybe a solution would be to keep a String property in the viewItem that is the value of the `questionnaireResponse.answer[index].value[x]`

PR #1487, opened 2022-07-05
Purge API implemented

### LINKED ISSUE #1472: Purge API implementation
**Describe the bug**
Problem with the current engine SDK design is if records are marked for deletion in the local database, these changes would be pushed to the server resulting in data loss.

Common scenarios	
CUJ 1: Purge all local data	
CUJ 2: Client initiated purge	
CUJ 3:  Server initiated purge

**Component**
Core library

**Expected behavior**
Engine SDK must provide API to support all 3 scenarios

**Related issues**
[856](https://github.com/google/android-fhir/issues/856)
[1178](https://github.com/google/android-fhir/issues/1178)

PallaviGanorkar:

Fixes #1472 

**Description**
Purge API - purge(), clearDatabase() and getLocalChange() API implemented and test cases added for all API

**Alternative(s) considered**
Yes, details are mentioned in design doc

**Type**
Choose one: | Feature

**Checklist**


engine/src/main/java/com/google/android/fhir/FhirEngine.kt
```diff
@@ -72,6 +73,37 @@ interface FhirEngine {
...
+
+  /**
+   * Purge resource from database based on resource type and id without any deletion of data from
+   * the server.
+   * @param type
+   * - The [ResourceType]
+   * @param id
+   * - The resource id [Resource.id]
+   * @param isLocalPurge
+   * - default value is false here resource will not be deleted from LocalChangeEntity table but it
```
jingtang10, line 101: formatting

engine/src/main/java/com/google/android/fhir/FhirEngine.kt
```diff
@@ -72,6 +73,37 @@ interface FhirEngine {
...
+   * resource from database. Each resource will have at most one
+   * [LocalChangeEntity](multiple changes are squashed).
+   * @param type The [ResourceType]
+   * @param id The resource id [Resource.id]
+   * @return [LocalChangeEntity] A squashed local changes for given resource type and id.
+   */
+  suspend fun getLocalChange(type: ResourceType, id: String): LocalChangeEntity
+
+  /**
+   * Purge resource from database based on resource type and id without any deletion of data from
```
jingtang10, line 94:
```suggestion
   * Purges a resource from the database based on resource type and id without any deletion of data from
```

jingtang10 (changes requested):

engine/src/main/java/com/google/android/fhir/db/impl/DatabaseImpl.kt
```diff
@@ -222,6 +223,53 @@ internal class DatabaseImpl(
...
+   * @param id resource id [Resource.id]
+   * @returns [LocalChangeEntity] A squashed local changes for given resource type and id.
+   */
+  override suspend fun getLocalChange(type: ResourceType, id: String): LocalChangeEntity {
+    return db.withTransaction {
+      val localChangeEntityList =
+        localChangeDao.getLocalChanges(resourceType = type, resourceId = id)
+      if (localChangeEntityList.isEmpty()) {
+        throw ResourceNotFoundException(type.name, id)
+      }
```
jingtang10, line 242: this is a legitimate case where a resource does not have any local change. in this case just return an empty thing instead of throwing.

but you probably should throw if the resource itself actually doesn't exist.

jingtang10 (changes requested):

engine/src/main/java/com/google/android/fhir/db/impl/DatabaseImpl.kt
PallaviGanorkar: Done

engine/src/main/java/com/google/android/fhir/FhirEngine.kt
```diff
@@ -72,6 +73,37 @@ interface FhirEngine {
 
   /** Returns the timestamp when data was last synchronized. */
   suspend fun getLastSyncTimeStamp(): OffsetDateTime?
+
+  /**
+   * Clears all database tables without resetting the auto-increment value generated by
+   * PrimaryKey.autoGenerate.
```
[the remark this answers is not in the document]
PallaviGanorkar: Done

engine/src/main/java/com/google/android/fhir/FhirEngine.kt
PallaviGanorkar: Done

engine/src/main/java/com/google/android/fhir/impl/FhirEngineImpl.kt
```diff
@@ -67,6 +68,18 @@ internal class FhirEngineImpl(private val database: Database, private val contex
     return DatastoreUtil(context).readLastSyncTimestamp()
   }
 
+  override suspend fun clearDatabase() {
+    database.clearDatabase()
+  }
+
+  override suspend fun getLocalChange(type: ResourceType, id: String): LocalChangeEntity? {
```
jingtang10, line 75: i wonder if we should define a public local change data class, that encapsulates the local change entity class.

the local change entity class should be internal and for db layer only.... so if something happens in the future and we change our db implmementation, the public api which deveopers depend on will not be affected.

what do you think?

engine/src/test/java/com/google/android/fhir/impl/FhirEngineImplTest.kt
```diff
@@ -194,6 +196,122 @@ class FhirEngineImplTest {
     testingUtils.assertResourceEquals(TEST_PATIENT_2, fhirEngine.get<Patient>(TEST_PATIENT_2_ID))
   }
 
+  @Test
+  fun getLocalChanges_withSingleLoaleChange_shouldReturnSingleLocalChanges() = runBlocking {
```
jingtang10, line 200: wonder if you can use the new kotlin test case names in reverse single quotes: https://kotlinlang.org/docs/coding-conventions.html#names-for-test-methods

jingtang10 (changes requested):

engine/src/test/java/com/google/android/fhir/impl/FhirEngineImplTest.kt
PallaviGanorkar: done

engine/src/main/java/com/google/android/fhir/impl/FhirEngineImpl.kt
PallaviGanorkar: Yes , I agreed, there must be separation for hiding internal implementation as well as app developer should not be affected due internal changes

jingtang10 (approved):
move local change data class outside of db package.

PallaviGanorkar:
> move local change data class outside of db package.

Done

PR #1490, opened 2022-07-08
build optimizations

yigit:
This CL is a mixture of changes to optimize the build.

A sample build can be seen here: https://github.com/google/android-fhir/actions/runs/2633666016
Where we compile & test the project ~3 minutes (when no code changed). Emulator tests are still slow :(

List of changes:

*  move the CI setup into 2 actions, each use a common action to setup the machine
   * build and test on host machine (on linux)
   * run emulator tests (mac)
* enabled gradle caching (this is where the main win happens)
* set gradle.properties similar to andoridx, removed ci properties
* removed buildToolsVersion, which is deprecated
* run device tests in 1 gradle execution, to avoid killing and re-creating the emulator between each run
* added --scan to all gradle invocations. Scan results are added to the build summary
* moved code generation into buildSrc and made it a cacheable task. This is the other major win wrt caching because changing sources while building will mess up caching.  (we should never write to input directories)
* another nice win related to codegen is that, we no longer commit generated files
* code generator was hitting "method too big"error, so changed it to divide into sub-methods
* limited memory for test runs. they get forked from current process so easily starve the device.


jingtang10 (approved):
@yigit this is lit 🔥

jingtang10:
Honestly this is so amazing we love you Yigit!

PR #1522, opened 2022-08-02
rename GitHub Action Files

omarismail94:


.github/workflows/deviceTests.yml
```diff
@@ -12,7 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-name: DeviceTests
+name: "GitHub Actions: DeviceTests"
```
jingtang10, line 15: should this be called instrumentation tests too? 

or should we change kokoro to device tests?

jingtang10 (changes requested):

.github/workflows/deviceTests.yml
omarismail94: changed the name of Kokoro to be the same as GitHub Actions

yigit (approved):

.github/workflows/deviceTests.yml
omarismail94: done

jingtang10 (approved):

PR #1552, opened 2022-08-10
Cursor moves to start position while entering or deleting decimal values in text field.

### LINKED ISSUE #1551: sdc-beta04 regression in decimal data entry
**Describe the bug**
When attempting to enter or delete numbers into a `type='decimal'` field, the cursor jumps around mangling input and making deletion difficult.

**Component**
SDC Library

**To Reproduce**

1. Load a questionnaire including a `decimal` field
2. Type more than 3 numbers consecutively
3. Tap to move the cursor to the end of the number then tap delete a few times

**Expected behavior**
Text to enter and delete normally; cursor should not move unexpectedly.

**Screenshots**

https://user-images.githubusercontent.com/7772901/183772633-f658c1c1-48db-4690-ad21-d2e8be04b373.mp4

 - Device: Pixel 6 Pro physical and Pixel 5 emulator. Details below for Pixel 6
 - Android version: 12
 - Build number: SQ3A.220705.003

**Additional context**
Problem went away after down-grading to beta03.

My specific questionnaire:

```json
{
  "status": "draft",
  "title": "Height & weight",
  "meta": {
    "profile": [
      "http://hl7.org/fhir/uv/sdc/StructureDefinition/sdc-questionnaire|2.7"
    ],
    "tag": [
      {
        "code": "lformsVersion: 29.2.1"
      }
    ]
  },
  "item": [
    {
      "type": "decimal",
      "code": [
        {
          "code": "3137-7",
          "display": "Body height Measured",
          "system": "http://loinc.org"
        }
      ],
      "extension": [
        {
          "url": "http://hl7.org/fhir/StructureDefinition/questionnaire-unit",
          "valueCoding": {
            "display": "cm"
          }
        }
      ],
      "required": false,
      "linkId": "/3137-7",
      "text": "Body height Measured"
    },
    {
      "type": "quantity",
      "code": [
        {
          "code": "3141-9",
          "display": "Weight Measured",
          "system": "http://loinc.org"
        }
      ],
      "extension": [
        {
          "url": "http://hl7.org/fhir/StructureDefinition/questionnaire-unitOption",
          "valueCoding": {
            "display": "[lb_av]"
          }
        },
        {
          "url": "http://hl7.org/fhir/StructureDefinition/questionnaire-unitOption",
          "valueCoding": {
            "display": "kg"
          }
        }
      ],
      "required": false,
      "linkId": "/3141-9",
      "text": "Weight Measured"
    }
  ]
}
```

santosh-pingle:

Fixes #1551 

**Description**
EditTextviewFactory view was just comparing string representation of decimal values while updating text, as result `x.0` and `x` was considered different values.

**Alternative(s) considered**

**Type**
Choose one: Bug fix 

Just demo purpose.

https://user-images.githubusercontent.com/86107848/183932176-e8b33de4-7f94-4e76-99df-d878f07c1ba7.mov

https://user-images.githubusercontent.com/86107848/183932229-31a45d43-c0e1-4969-850d-6cba74b683d3.mov

**Checklist**


datacapture/src/main/java/com/google/android/fhir/datacapture/views/QuestionnaireItemEditTextViewHolderFactory.kt
```diff
@@ -82,8 +82,15 @@ abstract class QuestionnaireItemEditTextViewHolderDelegate(private val rawInputT
 
     textInputEditText.removeTextChangedListener(textWatcher)
     val text = getText(questionnaireItemViewItem.answers.singleOrNull())
-    if (text != textInputEditText.text.toString()) {
-      textInputEditText.setText(getText(questionnaireItemViewItem.answers.singleOrNull()))
+    when (rawInputType) {
```
jingtang10, line 85: i think this is the parent class doing the work for the child classes. an alternative way of doing this is to create an open function that has a default implementation to update answer, and then in your decimal cases you override that.

datacapture/src/main/java/com/google/android/fhir/datacapture/views/QuestionnaireItemEditTextViewHolderFactory.kt
```diff
@@ -98,6 +105,18 @@ abstract class QuestionnaireItemEditTextViewHolderDelegate(private val rawInputT
     }
   }
 
+  private fun updateDecimalAnswer(answer: String) {
+    if (answer.isNotEmpty() && textInputEditText.text.toString().isNotEmpty()) {
+      if (answer.toDouble() != textInputEditText.text.toString().toDouble()) {
```
jingtang10, line 110: this is great. an alternative way of handing this, is that we just don't send the new answer to the view model at all. (see line 102 above). so do this check when you're updating the answer.

pro: we avoid an entire code path that recalculates the entire screen.
con: you'll have to introduce a variable in the UI to compare the new input with to decide if you send the answer to the view model

please try that - and discuss which way is better. thanks!

jingtang10 (changes requested):

datacapture/src/main/java/com/google/android/fhir/datacapture/views/QuestionnaireItemEditTextViewHolderFactory.kt
santosh-pingle: This check is required while text get updated in textfield.

Consider these 2 use cases : 

Case 1 : textfield is empty, answer is empty
User typed the values, now textfield has value, value get updated in answer. Now answers triggers the callback to update the value in textfield again, therefore check is required while text get updated in textfield.

Case 2 : textfield is empty, answer has value
Now answer value is set to the textfield in view factory, but it does not trigger the flow to update the value in answer. therefore check is not required while text get updated in answer.

jingtang10: what i'm saying is instead of doing the check here do the check in the on focus change listener.

jingtang10: please let me know when this is addressed.

santosh-pingle: looking into it.

santosh-pingle: It will not fix the issue if we  add this check in on focus change listener, because issue only occurs when user start typing the input. While user is typing the input an answer get updated for every single input which triggers the callback to update the text in bind api, therefore check is required in bind api before to set text in textfield.

datacapture/src/main/java/com/google/android/fhir/datacapture/views/QuestionnaireItemEditTextQuantityViewHolderFactory.kt
```diff
@@ -47,5 +44,15 @@ internal object QuestionnaireItemEditTextQuantityViewHolderFactory :
       ): String {
         return answer?.valueQuantity?.value?.toString() ?: ""
       }
+
+      override fun isTextUpdatesRequired(answerText: String, inputText: String): Boolean {
+        return if (answerText.isNotEmpty() && inputText.isNotEmpty()) {
+          answerText.toDouble() != inputText.toDouble()
+        } else {
+          true
+        }
```
jingtang10, line 53:
```suggestion
        if (answerText.isEmpty() && inputText.isEmpty()) {
          return false
        }
        if (answerText.isEmpty() || inputText.isEmpty()) {
          return true
        }
        // Avoid shifting focus by updating text field if the values are the same
        return answerText.toDouble() != inputText.toDouble()
```

jingtang10 (approved):

PR #1569, opened 2022-08-24
Optional support of Full-text search for Patient.name 

### LINKED ISSUE #608: Optional support for fuzzy search on Patient name
Patient names may be misspelled and when searching for an existing patient this may be missed leading to duplicate records fro an existing patient

- Provide option to enable fuzzy search on the name field
- Enabling it will create a new FTS index for the Patient.name[] field (TBD) and access to a search API
- By default this will be disabled

**Additional context**
FTS comes at a storage cost. 
This solution will allow implementers to determine whether fuzzy search is required for their use case.

shoaibmushtaq25:

Fixes #608 

**Description**
This PR is around optional support of Full-text search for a Patient.name. The implemented approach is that 
- Made a virtual/FTS table for `StringIndexEntity` named as [`FullTextStringIndexEntity`](https://github.com/google/android-fhir/pull/1569/files#diff-be01b8cc3854558e87b01eeff087ecacd27c080d188f9db082e1b1ddb496ea2eR26), the `StringIndexEntity` table will be available as [content table](https://developer.android.com/reference/androidx/room/Fts4#contentEntity()) for `FullTextStringIndexEntity` table
- Introduced a new `StringFilterModifier` named [`MATCHES_FTS`](https://github.com/google/android-fhir/pull/1569/files#diff-83c3131ca2790a2a016d1bb82e30328d2a664c288c89ebfb8a6e681e3c1846b3R155) to use it for MATCH/Fts searches as FTS uses `MATCH` operator in the query, [Syntax is like this](https://github.com/google/android-fhir/pull/1569/files#diff-df7794ca376f28815a4b43f1026f2db52fbc9237720d5b8abd9ab9d3946f5f54R45) `"MATCH '*' || ? || '*'"`
- This feature is made Optional like this, If any implementer wants to use Full-text search they should use the new StringFilterModifier `MATCHES_FTS` for this purpose else use the different StringFilterModifiers available `STARTS_WITH`, `MATCHES_EXACTLY` and `CONTAINS`
- When a user tries to search using FTS (which uses the `MATCH` operator in the query), It returns the search results which start with the search query. For Example, If the search query is `e`, it returns search results starting with e(either is a first name or last name) i.e, `Efeturi Money`, `Elly Kitoto` and `Njenga Elijah` etc. and for the search query `efe`, it returns only `Efeturi Money`, [Video is also attached here for more details](https://user-images.githubusercontent.com/12672919/192805927-961dc2d7-63f2-46e0-89c8-727e7bbf5ded.mp4)
- It also supports `OR/AND` operators in the form of `UNION` and `INTERSECT` with FTS queries. Let's say If our search query is `efe sho` with `OR` operator, it would return `Efeturi Money` and `Muhammad Shoaib Mushtaq` in search results. But with `AND` operator, It would return empty search results. 

Videos for FTS with 

1. [FTS with OR/UNION Operator](https://user-images.githubusercontent.com/12672919/198204676-619ba20a-3722-4287-ae79-7b0688fd595b.mp4) 
2. [FTS with AND/INTERSECT Operator](https://user-images.githubusercontent.com/12672919/198205524-85208105-dd52-4717-ae80-20fc5653344b.mp4) 

**Alternative(s) considered**

**Type**
Choose one: Feature

FTS Search on Patient.name is working as below,
https://user-images.githubusercontent.com/12672919/192805927-961dc2d7-63f2-46e0-89c8-727e7bbf5ded.mp4

**Checklist**


shoaibmushtaq25:
I am not sure why instrumentation tests are failing, looks like there is an issue at CI/CD level as it is happening for other PRs as well. cc: @jingtang10 
Screenshot from 2022-09-29 11-03-14

engine/src/androidTest/java/com/google/android/fhir/db/impl/DatabaseImplTest.kt
```diff
@@ -734,6 +734,32 @@ class DatabaseImplTest {
     assertThat(result.single().id).isEqualTo("Patient/${patient.id}")
   }
 
+  @Test
```
maimoonak, line 737: Add negative test as well. i.e. fts should not match xyz records

engine/src/test/java/com/google/android/fhir/search/SearchTest.kt
```diff
@@ -926,6 +926,42 @@ class SearchTest {
       )
   }
 
+  @Test
+  fun search_filter_string_matches_fts() {
```
maimoonak, line 930: test name with backtick and add negative tests as well

engine/src/androidTest/java/com/google/android/fhir/db/impl/DatabaseImplTest.kt
```diff
@@ -734,6 +734,32 @@ class DatabaseImplTest {
...
+  fun search_string_match_fts() = runBlocking {
```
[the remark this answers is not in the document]
shoaibmushtaq25: Considering all other tests in this class has names without backtics so to keep it consistent for this class, I write them without backtics. I think, Instead of mixing backtics and other format, It would be better to refactor all test names of this file into backtics format. What's your thoughts on that

engine/src/androidTest/java/com/google/android/fhir/db/impl/DatabaseImplTest.kt
shoaibmushtaq25: Done.

engine/src/test/java/com/google/android/fhir/search/SearchTest.kt
shoaibmushtaq25: @maimoonak, I Couldn't think of any negative tests here as we are only checking the query and params after passing the filters into Search APIs. I would appreciate If you can mention any negative cases here to cover in the test.

maimoonak: for example fts query with search string 'joh' should not return any patient if list have 'simon' or 'jane'

shoaibmushtaq25: This is done in `DatabaseImplTest` class.

shoaibmushtaq25: I changed it to backticks `` format for newly added unit test in this class.

engine/src/androidTest/java/com/google/android/fhir/db/impl/DatabaseImplTest.kt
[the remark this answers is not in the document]
shoaibmushtaq25: @maimoonak, Since DatabaseImpTest lies under instrumentation tests ( not tests) so using of backticks `` format is not supported yet in instrumentation tests as mentioned in the [kotlin documentation here](https://kotlinlang.org/docs/coding-conventions.html#names-for-test-methods)
>In tests (and only in tests), you can use method names with spaces enclosed in backticks. Note that such method names are currently not supported by the Android runtime. Underscores in method names are also allowed in the test code.

maimoonak (approved):

shoaibmushtaq25:
Had a call yesterday with @jingtang10 to review this PR https://github.com/google/android-fhir/pull/1569 and after having some initial review, Jing felt like we need to have more discussion with @fredhersch to understand what we want to achieve with this issue. It looks like, we are not achieving much from this issue, so we may not even need to implement this in FHIR SDK.

Jing said we will also be discussing the upcoming Android FHIR SDK Developers' call around this to consolidate what we want to achieve with this issue.

@jingtang10 / @maimoonak Can you please add If I have missed something here? cc: @f-odhiambo

shoaibmushtaq25:
@maimoonak , Please review this PR. I am done with the changes to support OR/AND with FTS queries.

jingtang10:
closing in favour of https://github.com/google/android-fhir/pull/1853

PR #1610, opened 2022-09-19
Skip hidden field from validation

### LINKED ISSUE #1579: [Form Validation] Hidden Required fields should not be validated
**Describe the bug**
- Currently, when a field on the Questionnaire is hidden the Required validation check is still applied to it. 
- This means the user will get an error prompting them to fill all required fields. The user is however not able to fill in the field since it's hidden. 

**To Reproduce**
Steps to reproduce the behaviour:
1. Add a Required field in a questionnaire. 
2. Fill up the questionnaire and try to submit it. 
3. An error asking the user to confirm that all the fields were filled up will be thrown. 

**Expected behaviour**
- [ ] All visible required fields should be validated
- [ ] All hidden required fields should not be validated on submission. 

**General Implementation Idea**
- [ ] We should not validate any field when the value of the hidden extension is set to `true`
    - This means we first check for the hidden extension and then run validation on the field.  

**Assumption**
- Skip Logic also uses the hidden extension to also hide/show fields. 

**Screenshots**
https://user-images.githubusercontent.com/4817531/188339502-7f9d17a0-7292-4412-ba86-bc97687e985d.mp4

Hidden field JSON
```json
{
  "linkId": "83943853-d9c9-46ad-da39-2cc55e93b974",
  "type": "integer",
  "text": "Birth weight",
  "required": true,
  "extension": [
    {
      "url": "http://hl7.org/fhir/StructureDefinition/questionnaire-hidden",
      "valueBoolean": true
    }
  ]
}
```

 - Device: Emulator & Xiaomi 11T Pro
 - Android version: 10 & 12 

Related SDK Issue
- https://github.com/google/android-fhir/issues/1579

maimoonak:
Fixes #1579 

**Description**
Hidden field does not qualify for any validation. Response validation marks hidden field as valid and passes validation irrespective of it value

**Alternative(s) considered**
The validation skips the hidden field in when doing questionnaire-reponse validation. The other option could have doing it inside each and every validator but that makes it repetitive, non central, and does not keep room to revert it easily for specific validator types in future. 

**Type**
Choose one: Bug fix 

**Checklist**


datacapture/src/main/java/com/google/android/fhir/datacapture/validation/QuestionnaireResponseItemValidator.kt
```diff
@@ -39,7 +40,12 @@ internal object QuestionnaireResponseItemValidator {
     answers: List<QuestionnaireResponse.QuestionnaireResponseItemAnswerComponent>,
     context: Context
   ): ValidationResult {
-    val validationResults = validators.map { it.validate(questionnaireItem, answers, context) }
+    val validationResults =
+      validators.map {
+        if (questionnaireItem.isHidden)
+          ConstraintValidator.ConstraintValidationResult(true, "Field is hidden")
```
ellykits, line 46: Should the text `"Field is hidden"` be translatable?

shoaibmushtaq25 (approved):
Thanks @maimoonak. LGTM. :+1:

datacapture/src/main/java/com/google/android/fhir/datacapture/validation/QuestionnaireResponseItemValidator.kt
maimoonak: This is not used anywhere on UI, so translation is not required.

ellykits (approved):
LGTM :+1:

jingtang10:
@maimoonak please use the assignee field to assign PRs to reviewers for a faster turn around

datacapture/src/main/java/com/google/android/fhir/datacapture/validation/QuestionnaireResponseItemValidator.kt
jingtang10: please move this logic to the beginning of this function.

if a questionnaire item is hidden, validation should be skipped altogether.

at the moment, you're still going through all the possible validators and generating a valid result for each of them. this is unnecessary.

datacapture/src/main/java/com/google/android/fhir/datacapture/validation/QuestionnaireResponseValidator.kt
```diff
@@ -23,9 +23,6 @@ import org.hl7.fhir.r4.model.Type
 
 object QuestionnaireResponseValidator {
 
-  /** Maps linkId to [ValidationResult]. */
-  private val linkIdToValidationResultMap = mutableMapOf<String, MutableList<ValidationResult>>()
-
```
jingtang10, line 28: thank you for catching this problem and fixing it.

datacapture/src/test/java/com/google/android/fhir/datacapture/validation/QuestionnaireResponseValidatorTest.kt
```diff
@@ -375,6 +376,140 @@ class QuestionnaireResponseValidatorTest {
...
+        questionnaireResponse,
+        context
+      )
+
+    assertThat(result.entries.first().key).isEqualTo("valid-hidden-item")
+    assertThat(result.entries.first().value.first()).isEqualTo(Valid)
+  }
+
+  @Test
+  fun `validation passes for required questionnaire item with hidden extension when value specified`() {
```
jingtang10, line 421: personally, i don't think this test case is needed. it's not exercising any new code paths compared with the previous test case.

datacapture/src/main/java/com/google/android/fhir/datacapture/validation/QuestionnaireResponseItemValidator.kt
jingtang10: i also think in this case you probably should return `NotValidated`

jingtang10 (changes requested):
thanks @maimoonak !

jingtang10 (approved):
Thanks

PR #1669, opened 2022-10-17
Fix sync state exception handling

### LINKED ISSUE #1654: Download fails to complete when exception is encountered
**Describe the bug**
When an exception is occurs while downloading a resource, the `com.google.android.fhir.sync.download.DownloaderImpl` attempts to emit a `com.google.android.fhir.sync.DownloadState.Failure` failure in the `catch` block which results to a FATAL_EXCEPTION.  

**Component**
Core library

**To Reproduce**
1. Attempt downloading a resource that has an exception. (Example in our use case was an `Encounter` resource missing a `coding.code`)

**Expected behavior**
In case of a Resource sync failure `com.google.android.fhir.sync.DownloadState.Failure` should be emitted with the parameter `ResourceSyncException` wrapping the sync exception without violating the flow exception transparency.

**Error logs**
```java
FATAL EXCEPTION: main
                                                                                                    Process: org.smartregister.fhircore.ecbis, PID: 10487
                                                                                                    java.lang.IllegalStateException: Flow exception transparency is violated:
                                                                                                        Previous 'emit' call has thrown exception java.lang.NullPointerException: coding.code must not be null, but then emission attempt of value 'Failure(syncError=ResourceSyncException(resourceType=Encounter, exception=java.lang.NullPointerException: coding.code must not be null))' has been detected.
                                                                                                        Emissions from 'catch' blocks are prohibited in order to avoid unspecified behaviour, 'Flow.catch' operator can be used instead.
                                                                                                        For a more detailed explanation, please refer to Flow documentation.
                                                                                                    	at kotlinx.coroutines.flow.internal.SafeCollector.exceptionTransparencyViolated(SafeCollector.kt:140)
                                                                                                    	at kotlinx.coroutines.flow.internal.SafeCollector.checkContext(SafeCollector.kt:104)
                                                                                                    	at kotlinx.coroutines.flow.internal.SafeCollector.emit(SafeCollector.kt:83)
                                                                                                    	at kotlinx.coroutines.flow.internal.SafeCollector.emit(SafeCollector.kt:66)
                                                                                                    	at com.google.android.fhir.sync.download.DownloaderImpl$download$2.invokeSuspend(DownloaderImpl.kt:56)
                                                                                                    	at kotlin.coroutines.jvm.internal.BaseContinuationImpl.resumeWith(ContinuationImpl.kt:33)
                                                                                                    	at kotlinx.coroutines.flow.internal.SafeCollector.invokeSuspend(SafeCollector.kt:48)
                                                                                                    	at kotlin.coroutines.jvm.internal.BaseContinuationImpl.resumeWith(ContinuationImpl.kt:33)
                                                                                                    	at kotlinx.coroutines.flow.internal.SafeCollector.invokeSuspend(SafeCollector.kt:48)
                                                                                                    	at kotlin.coroutines.jvm.internal.BaseContinuationImpl.resumeWith(ContinuationImpl.kt:33)
                                                                                                    	at kotlinx.coroutines.DispatchedTask.run(DispatchedTask.kt:104)
                                                                                                    	at android.os.Handler.handleCallback(Handler.java:938)
                                                                                                    	at android.os.Handler.dispatchMessage(Handler.java:99)
                                                                                                    	at android.os.Looper.loop(Looper.java:223)
                                                                                                    	at android.app.ActivityThread.main(ActivityThread.java:7656)
                                                                                                    	at java.lang.reflect.Method.invoke(Native Method)
                                                                                                    	at com.android.internal.os.RuntimeInit$MethodAndArgsCaller.run(RuntimeInit.java:592)
                                                                                                    	at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:947)
                                                                                                    	Suppressed: kotlinx.coroutines.DiagnosticCoroutineContextException: [StandaloneCoroutine{Cancelling}@1662c1a, Dispatchers.Main.immediate]

```
Libraries used:
`org.jetbrains.kotlinx:kotlinx-coroutines-core:1.6.2`
`org.jetbrains.kotlinx:kotlinx-coroutines-android:1.6.2`

ellykits:

Fixes #1654

**Description**
Surround the call site of `com.google.android.fhir.impl.FhirEngineImpl#saveRemoteResourcesToDatabase` and `com.google.android.fhir.impl.FhirEngineImpl#saveResolvedResourcesToDatabase` in `try-catch` block.

**Alternative(s) considered**
The exception we were getting was a `NullPointerException`
```
java.lang.NullPointerException: coding.code must not be null
2022-10-18 10:38:10.162  7941-7985  FhirEngine...ed$collect org.smartregister.fhircore.map       E  	at com.google.android.fhir.index.ResourceIndexer.tokenIndex(ResourceIndexer.kt:319)
2022-10-18 10:38:10.162  7941-7985  FhirEngine...ed$collect org.smartregister.fhircore.map       E  	at com.google.android.fhir.index.ResourceIndexer.extractIndexValues(ResourceIndexer.kt:92)
2022-10-18 10:38:10.162  7941-7985  FhirEngine...ed$collect org.smartregister.fhircore.map       E  	at com.google.android.fhir.index.ResourceIndexer.index(ResourceIndexer.kt:71)
2022-10-18 10:38:10.162  7941-7985  FhirEngine...ed$collect org.smartregister.fhircore.map       E  	at com.google.android.fhir.db.impl.dao.ResourceDao.insertResource(ResourceDao.kt:194)
```

Solution would be to add the null check before using the property on this line `com/google/android/fhir/index/ResourceIndexer.kt:317`. This will however only address this type of exception and ignore any other exceptions that may happen.

**Type**
Choose one: Bug fix

**Checklist**


engine/src/main/java/com/google/android/fhir/sync/download/DownloaderImpl.kt
```diff
@@ -53,7 +53,7 @@ internal class DownloaderImpl(
           )
         )
       } catch (exception: Exception) {
-        emit(DownloadState.Failure(ResourceSyncException(resourceTypeToDownload, exception)))
```
maimoonak, line 56: we may not want to remove this line. sync should emit failure when resource syncing fails.

engine/src/main/java/com/google/android/fhir/impl/FhirEngineImpl.kt
```diff
@@ -88,20 +88,24 @@ internal class FhirEngineImpl(private val database: Database, private val contex
...
+              resolveConflictingResources(
+                resources,
+                getConflictingResourceIds(resources),
+                conflictResolver
+              )
+            saveRemoteResourcesToDatabase(resources)
+            saveResolvedResourcesToDatabase(resolved)
+          }
+        } catch (exception: Exception) {
+          Timber.e(exception, "Save remote and resolved conflicting resources to database failed")
```
maimoonak, line 108: we may not want to just catch exception for resources which have failed to save. it should either show  this error to user, or it should do something to log error somewhere to resolve the issue with DB. this can lead to suppress app issues with data collection without getting noticed.

engine/src/main/java/com/google/android/fhir/sync/download/DownloaderImpl.kt
ellykits: I see. I have reverted this line so this state can be emitted.

ellykits: But this needs to be handled not only by the implementers but also in `FHirEngineImpl` where the download function is called.

engine/src/main/java/com/google/android/fhir/impl/FhirEngineImpl.kt
ellykits: I agree we may need to do more with this error log but do we really want to show the user this error? Is there a strategy in SDK for handling/logging errors?

f-odhiambo:
@omarismail94 Do you mind providing some feedback on this PR too . Thanks!

omarismail94:
@f-odhiambo yep! Im working on trying to replicate the issue in a unit test

ellykits:
Reopened this as the issue was not resolved via https://github.com/google/android-fhir/pull/1917

ellykits:
Reopened this as the issue was not resolved via https://github.com/google/android-fhir/pull/1917

ellykits:
@omarismail94 I had closed this PR in favor of https://github.com/google/android-fhir/pull/1917 however after testing, the issue is not yet resolved on the PR you had created. Maybe this needs to be investigated further.

jingtang10:
hey @ellykits - sorry for not having closed this earlier. looking at this PR again i'm not confident this is the route we want to take to fix the problem. what is done here is wrapping the problem in a try catch instead of addressing the exception. for example, if it's network issue, we might need to try again, or if this is a failure we can't recover from we need a way for the application to know about that.

additionally, @santosh-pingle is working on https://github.com/google/android-fhir/pull/2142 which is very relevant to this problem.

so i'll close this PR - but we should still keep the issue open to continue exploring the correct solution.

PR #1717, opened 2022-11-14
Skip hidden page

### LINKED ISSUE #1622: Apply hidden extension to a page
Be able to hide a page

Using the hidden extension should hide the page as shown below. Currently, it just renders a blank page. This implementation would be similar to how a page is hidden when using an enable when in https://github.com/google/android-fhir/pull/1532. 

```
"extension": [
                {
                    "url": "http://hl7.org/fhir/StructureDefinition/questionnaire-itemControl",
                    "valueCodeableConcept": {
                        "coding": [
                            {
                                "system": "http://hl7.org/fhir/questionnaire-item-control",
                                "code": "page",
                                "display": "Page"
                            }
                        ],
                        "text": "Page"
                    }
                },
                {
                    "url": "http://hl7.org/fhir/StructureDefinition/questionnaire-hidden",
                    "valueBoolean": true
                }
            ],
            "linkId": "page-chw-assigned",
            "type": "group",
            "text": "<h1> CHW Assignment</h1>",
            "item": [
            ]
}
```

**Use case**
There are some questionnaires where we use items to perform fhirpath calculations that don't need to be seen by the user. For example, if a patient is younger than 6 years then we don't want the user asking if the patient is wearing a mask. To do this, we first create an item that checks if the patient is younger than 6 years:
```
{
      "extension": [
        {
          "url": "http://hl7.org/fhir/uv/sdc/StructureDefinition/sdc-questionnaire-initialExpression",
          "valueExpression": {
            "language": "text/fhirpath",
            "expression": "Patient.birthDate >= (today() - ((6 * 365).toString() + ' days').toQuantity())"
          }
        },
        {
          "url": "http://hl7.org/fhir/StructureDefinition/questionnaire-hidden",
          "valueBoolean": true
        }
      ],
      "linkId": "is-younger-than-6-years",
      "text": "Is younger than 6 years",
      "type": "boolean",
      "required": true,
      "readOnly": true
    }
```
Once this is done we then check this item as shown below:
```
{
      "linkId": "does-patient-have-mask",
      "text": "* <b>ASK:</b> Does the patient have a mask?",
      "type": "boolean",
      "enableWhen": [
        {
          "question": "is-younger-than-6-years",
          "operator": "=",
          "answerBoolean": false
        }
      ],
      "required": true
    }
```

While this works well for a questionnaire with a single page, it's quite challenging for a paginated questionnaire. This is because it is difficult to decide where to store these calculations such that they will be easily accessible. Ideally, we would want to create a page that includes all the questionnaires calculations. This page would be at the end of the questionnaire and we would make it hidden so that it doesn't interfere with the user's workflow. In addition, the developers know exactly where to find it when reviewing questionnaires or accessing any items during structure map based extraction.

jingtang10:

Fixes #1622 

**Description**
Respect hidden extension for pages.
Also renamed the repeated group questionnaire json file name to be consistent with the other files.

**Alternative(s) considered**

**Type**
Bug fix

**Checklist**
- [x] I have signed the Google [Individual CLA](https://cla.developers.google.com/about/google-individual), or I am cxvered by my company's [Corporate CLA](https://cla.developers.google.com/about/google-corporate ).


datacapture/src/main/java/com/google/android/fhir/datacapture/QuestionnaireViewModel.kt
```diff
@@ -360,33 +366,19 @@ internal class QuestionnaireViewModel(application: Application, state: SavedStat
 
   /** [QuestionnaireState] to be displayed in the UI. */
   internal val questionnaireStateFlow: StateFlow<QuestionnaireState> =
-    combine(modificationCount, currentPageIndexFlow, reviewFlow) { _, pagination, reviewFlow ->
+    combine(modificationCount, currentPageIndexFlow, reviewFlow) { _, currentPageIndex, reviewFlow
```
omarismail94, line 369: since `getQuestionannireState` does not take in `currentPageIndex` any more, do  we need to  `currentPageIndex`? If not, let's substitute it with `_`

datacapture/src/main/java/com/google/android/fhir/datacapture/QuestionnaireViewModel.kt
```diff
@@ -360,33 +366,19 @@ internal class QuestionnaireViewModel(application: Application, state: SavedStat
...
-          )
+          getQuestionnaireState(reviewMode = reviewFlow)
         } else {
-          getQuestionnaireState(
-            questionnaireItemList = questionnaire.item,
-            questionnaireResponseItemList = questionnaireResponse.item,
-            currentPageIndex = pagination,
-            reviewMode = reviewFlow
-          )
+          getQuestionnaireState(reviewMode = reviewFlow)
```
omarismail94, line 374: should we change this `if/else` to a `when` statement:

```
        when (reviewFlow) {
          true -> getQuestionnaireState(reviewMode = true )
          false -> getQuestionnaireState(reviewMode = false)
        }
      }
```

datacapture/src/main/java/com/google/android/fhir/datacapture/QuestionnaireViewModel.kt
```diff
@@ -757,10 +742,21 @@ internal class QuestionnaireViewModel(application: Application, state: SavedStat
       null
     }
 
-  private fun Questionnaire.getInitialPages() =
+  private fun getQuestionnairePages() =
```
omarismail94, line 745: Do we see this function being called on non-paginated questionnaires?  If not, do we need to have the if/else to check if the questionnaire is paginated?

datacapture/src/main/java/com/google/android/fhir/datacapture/QuestionnaireViewModel.kt
```diff
@@ -840,6 +836,7 @@ internal data class QuestionnairePagination(
 internal data class QuestionnairePage(
   val index: Int,
   val enabled: Boolean,
+  val hidden: Boolean,
```
omarismail94, line 839: how do you feel about naming this to `display`, where true means display, and false means hidden? This would make the logic of whether to display a page more readable. Right now, we do  `!it.hidden`, if we switch to the suggested way, we can do `it.display`

datacapture/src/test/java/com/google/android/fhir/datacapture/QuestionnaireViewModelTest.kt
```diff
@@ -1971,9 +1983,150 @@ class QuestionnaireViewModelTest(
...
+                QuestionnairePage(0, enabled = true, hidden = false),
+                QuestionnairePage(1, enabled = false, hidden = false),
+                QuestionnairePage(2, enabled = true, hidden = false),
+              ),
+            currentPageIndex = 2
+          )
+        )
+    }
+  }
+
```
omarismail94, line 1995: can we add a test to show submit or review button on the penultimate page if the last page is a hidden page?

datacapture/src/main/java/com/google/android/fhir/datacapture/QuestionnaireViewModel.kt
jingtang10: thanks for catching! done.

datacapture/src/main/java/com/google/android/fhir/datacapture/QuestionnaireViewModel.kt
jingtang10: hey so this can actually be further simplified to `getQuestionnaireState(reviewFlow)`

jingtang10 (commented):
thanks @omarismail94

datacapture/src/main/java/com/google/android/fhir/datacapture/QuestionnaireViewModel.kt
omarismail94: I initially thought we can remove the if/else condition in this function, but I think it makes sense to keep it just in case this function gets called  for a non-paginated questionnaire

omarismail94 (approved):
Excellent! 🚀

PR #1727, opened 2022-11-23
Add integer field limiter

### LINKED ISSUE #1718: Catalog: Integer field deletes itself after 11th digit
**Describe the bug**
While typing into "ID number" which is an integer field, once you type the 11th number the field auto-deletes all the contents.

**Component**
SDC gallery app

**To Reproduce**
1. Select "Text field" in catalog app
2. Select "ID number" field
3. Enter an 11 digit number

**Expected behavior**
Field does not delete itself.

**Screenshots**

https://user-images.githubusercontent.com/7772901/201771093-4cbd5ca1-d353-482b-a687-06c4d6e8a194.mp4

 - Device: Pixel 4

**Additional context**
Different than #1594 - the auto-delete behavior is new, in 1594 the text just scrolls so it's not visible.

FikriMilano:

Fixes #1718 

**Description**
Validate integer upper limit:
- Show error if value exceed limit
- Prevents text field to clear itself, exclusive for integer field type

~- Limit integer field value to it's maximum allowed amount~
~- Change `editText.setText(text)` to `editText.append(text)`, so the cursor can stay at the end~
~- `editText.setText(emptyText)` is only used to clear the previous text before appending the new one~

**Alternative(s) considered**
N/A

**Type**
Choose one: Bug fix

**Checklist**


FikriMilano:
Facing this issue
[Bug.webm](https://user-images.githubusercontent.com/62053304/203463156-3a4659d5-5161-4025-be9c-be4ef4f4cab5.webm)

What I'm trying to achieve is:
1. The cursor stay at the end
2. The user won't be able to input more than the integer limit

datacapture/src/main/java/com/google/android/fhir/datacapture/views/QuestionnaireItemEditTextIntegerViewHolderFactory.kt
```diff
@@ -33,8 +34,14 @@ internal object QuestionnaireItemEditTextIntegerViewHolderFactory :
       override fun getValue(
         text: String
       ): QuestionnaireResponse.QuestionnaireResponseItemAnswerComponent? {
-        return text.toIntOrNull()?.let {
-          QuestionnaireResponse.QuestionnaireResponseItemAnswerComponent().setValue(IntegerType(it))
+        return text.toBigIntegerOrNull()?.let {
+          if (it > Int.MAX_VALUE.toBigInteger()) {
+            QuestionnaireResponse.QuestionnaireResponseItemAnswerComponent()
```
maimoonak, line 39: move this out of if else and value should should be evaluated inside if else to remove duplication

datacapture/src/main/java/com/google/android/fhir/datacapture/views/QuestionnaireItemEditTextIntegerViewHolderFactory.kt
```diff
@@ -33,8 +34,14 @@ internal object QuestionnaireItemEditTextIntegerViewHolderFactory :
...
+              .setValue(IntegerType(Int.MAX_VALUE))
```
maimoonak, line 40: We are explicitly setting value to something else here. Kindly add a comment here how app is handling the error propagation to user to let them know that their input value is not entertained.. For e.g. text can be 999999999999999... but int is changing it to int.max-value. How we make sure that this does not go unnoticed. by user

datacapture/src/main/java/com/google/android/fhir/datacapture/views/QuestionnaireItemEditTextViewHolderFactory.kt
```diff
@@ -103,6 +116,10 @@ abstract class QuestionnaireItemEditTextViewHolderDelegate(private val rawInputT
 
   open fun isTextUpdatesRequired(answerText: String, inputText: String) = (answerText != inputText)
 
+  open fun isTextExceedLimit(editable: Editable?) = false
+
+  open fun getTextDefaultLimit() = ""
```
maimoonak, line 121: this seems confusing. text default limit should be null or max possible value. empty string makes it confusing.

datacapture/src/main/java/com/google/android/fhir/datacapture/views/QuestionnaireItemEditTextViewHolderFactory.kt
```diff
@@ -85,11 +85,24 @@ abstract class QuestionnaireItemEditTextViewHolderDelegate(private val rawInputT
...
+        updateAnswer(editable)
+
+        // Prevents to do further operations everytime the empty setText() is triggered
+        if (editable.toString().isEmpty()) {
+          return@doAfterTextChanged
+        }
+
+        if (isTextExceedLimit(editable)) {
+          textInputEditText.setText("")
+          textInputEditText.append(getTextDefaultLimit())
```
maimoonak, line 103: update answer actually sets the response-answer-value. the code updates the answer value to something else the edit text shows a different value by overriding the value to default

datacapture/src/main/java/com/google/android/fhir/datacapture/views/QuestionnaireItemEditTextViewHolderFactory.kt
```diff
@@ -85,11 +85,24 @@ abstract class QuestionnaireItemEditTextViewHolderDelegate(private val rawInputT
     textInputEditText.removeTextChangedListener(textWatcher)
     val text = getText(questionnaireItemViewItem.answers.singleOrNull())
     if (isTextUpdatesRequired(text, textInputEditText.text.toString())) {
-      textInputEditText.setText(getText(questionnaireItemViewItem.answers.singleOrNull()))
+      textInputEditText.setText("")
```
maimoonak, line 88: why we are doing this?

FikriMilano: - Changed `editText.setText(text)` to `editText.append(text)`, so the cursor can stay at the end
- `editText.setText(emptyText)` is only used to clear the previous text before appending the new one

FikriMilano: if we use `editText.setText(text)` - when it's triggered while the edit text has focus, the cursor will go to the start.

I tried to use .setSelection() to move the cursor, but it doesn't seem to work.

datacapture/src/main/java/com/google/android/fhir/datacapture/views/QuestionnaireItemEditTextIntegerViewHolderFactory.kt
FikriMilano: sure

datacapture/src/main/java/com/google/android/fhir/datacapture/views/QuestionnaireItemEditTextIntegerViewHolderFactory.kt
FikriMilano: sure I will add some comments.
and we will also add an error message to let the user know, when their input is not entertained

datacapture/src/main/java/com/google/android/fhir/datacapture/views/QuestionnaireItemEditTextViewHolderFactory.kt
FikriMilano: ok, I think `null` makes more sense here. 
because input doesn't necessarily always a number, which has max value

datacapture/src/main/java/com/google/android/fhir/datacapture/views/QuestionnaireItemEditTextViewHolderFactory.kt
FikriMilano: yes that's right, let me know if you have more suggestion about this.

by appending the new text, it will trigger a recursive, in which it will call the update answer again, then break off from the recursive since it no longer exceeds the limit

FikriMilano:
@maimoonak 
There was a change in the approved solution.

The new solution is:
- Show error if value exceed limit
- Prevents text field to clear itself, exclusive for integer field type

datacapture/src/main/java/com/google/android/fhir/datacapture/views/QuestionnaireItemEditTextViewHolderFactory.kt
```diff
@@ -89,20 +89,32 @@ abstract class QuestionnaireItemEditTextViewHolderDelegate(private val rawInputT
...
+          validateTextUpperLimit(editable)
+        }
+      }
   }
 
   private fun updateAnswer(editable: Editable?) {
     val input = getValue(editable.toString())
     if (input != null) {
-      questionnaireItemViewItem.setAnswer(input)
+      if (input.hasValue()) {
```
ndegwamartin, line 104: Perhaps we should combine these two checks?

datacapture/src/androidTest/java/com/google/android/fhir/datacapture/views/QuestionnaireItemEditTextIntegerViewHolderFactoryEspressoTest.kt
```diff
@@ -0,0 +1,83 @@
...
+        answersChangedCallback = { _, _, _ -> },
+      )
+    runOnUI { viewHolder.bind(questionnaireItemViewItem) }
+
+    onView(withId(R.id.text_input_edit_text)).perform(click())
+    onView(withId(R.id.text_input_edit_text)).perform(typeText("${Int.MAX_VALUE}9"))
+    assertThat(
+        viewHolder.itemView.findViewById<TextInputLayout>(R.id.text_input_layout).error.toString()
+      )
+      .isNotEmpty()
```
ndegwamartin, line 70: Is it possible to also assert the actual value?

If the concern is localisation I think regardless of the local CI runs with at the time the correct value will be used. This will require the expected value to be retrieved from the context via string_id

FikriMilano: yeah, let's do that

datacapture/src/main/java/com/google/android/fhir/datacapture/views/QuestionnaireItemEditTextViewHolderFactory.kt
FikriMilano: actually, we need to keep this, with a bit of change and comments.

I made a change in which getValue returns an empty value instead a null, to prevent it clear itself.
null happens because the value surpasses the integer limit.
https://github.com/google/android-fhir/pull/1727/files#diff-6cb9f6285e5c442c229a831e85323b5c22edab8427019f3e735fe67a48108ecbR43

line 104 makes sure it only accepts non empty value, perhaps I need to change this to non empty checks

ndegwamartin: Got it, sounds good!

ndegwamartin (approved):

FikriMilano:
@jingtang10 this is ready for review

jingtang10 (changes requested):
A couple of high level comments

1. Please check out the `validation` package in the SDC library code, it is a collection of validators that the view model uses to validate answers the user inputs. I think you can add a new validator (maybe name it `IntegerRangeValidator` or something similar) and by doing this, we are using an existing validation mechanism. Probably a bit more modularised and easier to maintain.
2. As I commented in the issue itself, I think you need to format the number correctly depending on the locale. Try this API and see if it works https://developer.android.com/reference/android/icu/text/NumberFormat. And maybe change your device's locale to see if the number format has changed.

FikriMilano:
thanks, sure Jing

FikriMilano:
@jingtang10 
was having some issues with this.
the IntegerType answer value never surpasses it's limit, hence whenever we do validation, it shows no error.

the validation mechanism is not feasible, unless we use DecimalType for integer field widget.

FikriMilano:
closed by https://github.com/google/android-fhir/pull/1873
cc @jingtang10

Issue #1729, opened 2022-11-23
Implement KnowledgeManager module 


ktarasenko:
**Is your feature request related to a problem? Please describe.**
Load implementation guide and execute Knowledge artifacts from them. 
Solve the problem of storing bigger FHIR resources

PR #1730, opened 2022-11-23
Add compound Fhir Evaluator Api to let FhirOperator support multiple sources

ktarasenko:

Issue #1729

**Description**
Prework for integrating implementation guide module to the workflow module. Introduce FHIR Engine API implementation that allow to add multiple data sources to the FHIR Operator. In the future, the ImplementationGuide  module will become a second source for libraries and value sets. 

**Alternative(s) considered**

**Type**
Choose one: Feature

**Checklist**
- [ ] I have read and acknowledged the [Code of conduct](https://github.com/google/android-fhir/blob/master/docs/CODE_OF_CONDUCT.md).

### DIFF
--- a/workflow/src/androidTest/java/com/google/android/fhir/workflow/FhirOperatorLibraryEvaluateTest.kt
+++ b/workflow/src/androidTest/java/com/google/android/fhir/workflow/FhirOperatorLibraryEvaluateTest.kt
@@ -1,5 +1,5 @@
 /*
- * Copyright 2021 Google LLC
+ * Copyright 2022 Google LLC
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -40,6 +40,7 @@ class FhirOperatorLibraryEvaluateTest {
 
   private lateinit var fhirEngine: FhirEngine
   private lateinit var fhirOperator: FhirOperator
+  private lateinit var fhirEngineApiProvider: FhirEngineApiProvider
 
   private val fhirContext = FhirContext.forCached(FhirVersionEnum.R4)
   private val jsonParser = fhirContext.newJsonParser()
@@ -55,7 +56,8 @@ class FhirOperatorLibraryEvaluateTest {
   @Before
   fun setUp() = runBlocking {
     fhirEngine = FhirEngineProvider.getInstance(ApplicationProvider.getApplicationContext())
-    fhirOperator = FhirOperator(fhirContext, fhirEngine)
+    fhirEngineApiProvider = FhirEngineApiProvider(fhirContext, fhirEngine)
+    fhirOperator = FhirOperator(fhirContext, fhirEngineApiProvider)
   }
 
   /**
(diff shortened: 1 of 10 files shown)


vitorpamplona:
So, the `CompoundLibraryContentProvider` makes sense to me, but the other Compounds not so much. 

# CompoundTerminologyProvider

Shouldn't terminologies be imported into the Engine when downloading an IG? I assume non-workflow parts of the app will also need to reach into Terminologies from an IG and it might not make sense to use the workflow library to get them. 

# CompoundRetrieveProvider and CompoundFhirDal

I don't think IGs provide datasets. Do they? If they do, it would trigger a similar situation to the CompoundTerminologyProvider. 

# Easiness of use

With this change, the FhirOperator's constructor will require a list of providers (engine's database provider + file system provider). I believe this structure exposes the internals of the library to the user and thus not a good idea. Users don't need to know where IGs are being saved and/or where data/libraries are coming from. 

# Rethinking

I am starting to favor a plugin to the Engine's module as opposed to the Workflow. IG downloads shouldn't require the workflow lib and the Engine's Sync is in a better position to choose which objects need to be imported into the DB and which objects can be forwarded to the filesystem.

ktarasenko:
> So, the `CompoundLibraryContentProvider` makes sense to me, but the other Compounds not so much.
> 
> ### CompoundTerminologyProvider
> Shouldn't terminologies be imported into the Engine when downloading an IG? I assume non-workflow parts of the app will also need to reach into Terminologies from an IG and it might not make sense to use the workflow library to get them.

My initial assumption was that ValueSets are only referred from Libraries/StructureDefinitions and are subject to the same versioning as the Libraries and other parts of IGs. I also wanted to make sure there's a possibility to distinguish value sets coming from different IGs and/or different version of IGs.  Is that right? 

You mentioned that ValueSets are stored in memory on the server side, we might want to have a similar approach here - cache used ValueSets and/or preload them if we know we'll need them (take the advantage of relatedArtifact/depends-on section of the library metadata) 

I can imagine ways of sharing IG data with other parts of the SDK, but let's discuss this in the "approach" section. 

> 
> ### CompoundRetrieveProvider and CompoundFhirDal
> I don't think IGs provide datasets. Do they? If they do, it would trigger a similar situation to the CompoundTerminologyProvider.

Technically, some IGs have example data. I think the priority of supporting it is low, but I'd like to keep this in mind. 
> 
> ### Easiness of use
> With this change, the FhirOperator's constructor will require a list of providers (engine's database provider + file system provider). I believe this structure exposes the internals of the library to the user and thus not a good idea. Users don't need to know where IGs are being saved and/or where data/libraries are coming from.
> 

Right. This probably could be done in a less obvious way. I thought it might give us some flexibility in the future, if we want to add more data sources to FhirOperator at some point (for example now there's an implicit LibraryContentProvider for compiled Libraries) 

> ### Rethinking the approach
> I am starting to favor a plugin to the Engine's module as opposed to the Workflow. IG downloads shouldn't require the workflow lib and the Engine's Sync is in a better position to choose which objects need to be imported into the DB and which objects can be forwarded to the filesystem.

One of the ways to do it would be to provide APIs similar to FhirEvaluator and let FhirEngine lookup certain types of resources elsewhere. I'd still avoid mixing up the IG data with user-generated one though, unless there's a strict advantage of it. 
@jingtang10 you had some concerns about this approach, do you remember any?

PR #1750, opened 2022-12-05
Add FhirAdapter, UploadWorkManager Interface, and SearchParamType enum to SDK

### LINKED ISSUE #1706: Support different FHIR versions in the FHIR Engine lib

Support different FHIR versions in the FHIR Engine lib

Ability to deploy an app that uses the FHIR Engine using R5 resources

omarismail94:
Fixes #1706 

**Description**
* Add data class, `FhirAdapter`, that holds lambdas to apply operations on FHIR-version specific resources
* Add interface, UploadWorkManager, that will be implemented by `TransactionBundleGenerator`  in a follow-up PR
* Add a new data class, `ResourceMetadataToUpdate`, to save response results retrieved from remote server to DB. This class will be used later to remove the need for the `FhirEngine` to know of the returned response is a Bundle or something different 
* Implement an R4 adapter, making public any classes in the engine library that the R4 adapter needs to use
* Add our own `SearchParamType` based on the R4 and R5 one (same thing in both libs) so that in the future, we are no longer dependent on either lib

**Alternative(s) considered**
Yes, see  PRs:
 #1734
  #1538
 #1707 

**Type**
Choose one: (Bug fix | **Feature** | Documentation | Testing | Code health | Builds | Releases | Other)

N/A

**Checklist**


engine/src/main/java/com/google/android/fhir/index/SearchParamType.kt
```diff
@@ -0,0 +1,141 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.google.android.fhir.index
+
+enum class SearchParamType {
```
yigit, line 19: why not declare this as

```
enum class SearchParamType(
   val code: String,
   val system: String,
   val definition: String,
    ... etc)
   ```

instead of when statements?

File: engine/src/main/java/com/google/android/fhir/index/SearchParamType.kt [hunk 2]
```diff
@@ -0,0 +1,141 @@
...
+        QUANTITY -> "Quantity"
+        URI -> "URI"
+        SPECIAL -> "Special"
+        NULL -> null
+      }
+    }
+
+  companion object {
+    @Throws(Exception::class)
+    fun fromCode(codeString: String?): SearchParamType? {
```
yigit, line 127: i think it is better if you made this iterate over values and hold a map.
e.g.

```
companion object {
    private val codeMapping = SearchParamType.values().associateBy { it.code }
   fun fromCode(codeString: String?)  = codeMapping[codeString] ?: error("...")
}
```
so you won't need to update any of this class when you add a new enum (with the other constructor change)

yigit (approved):
just 1 small nit in the enum class.

engine/src/main/java/com/google/android/fhir/index/SearchParamType.kt
omarismail94: great idea! done

engine/src/main/java/com/google/android/fhir/index/SearchParamType.kt
omarismail94: adopted!

jingtang10 (changes requested):
This pr will break release as of today.

Issue being that the gradle publish task in engine will not automatically include the source code from the new r4 module.

Can you test this? To fix this you might need to change the build rule (add source code path) in the engine module.

jingtang10:
> This pr will break release as of today.
> 
> Issue being that the gradle publish task in engine will not automatically include the source code from the new r4 module.
> 
> Can you test this? To fix this you might need to change the build rule (add source code path) in the engine module.

Actually can you please hold on. Does it really make sense for us to create a separate module? Or does it make more sense simply to add different implementations inside the engine module. The bulk of binary size probably comes from HAPI structures, and our implementation to the adapter interface is probably a very small addition. Do we really want to manage multiple releases for each fhir version, or can we just put them in the same lib? For now, at least.

omarismail94:
> This pr will break release as of today.
> 
> Issue being that the gradle publish task in engine will not automatically include the source code from the new r4 module.
> 
> Can you test this? To fix this you might need to change the build rule (add source code path) in the engine module.

@jingtang10 From our in-person discussion, I've moved everything back into the engine library so we can maintain one release as opposed to (X+1) releases each time we update the FHIR Engine (X being the number of FHIR versions we support and the one for the core engine lib)

PR #1753, opened 2022-12-05
Workflow: Cache ModelResolver and use cached FhirContext

vitorpamplona:
ModelResolver is one of the slowest objects to load: 5.12 seconds in debug. 

This PR
1. Makes R4FhirModelResolver a singleton. 
2. Makes use of a cached context for increased performance. 
 
The default constructor R4FhirModelResolver() constructor recreates the FhirContext.

Fixes #[issue number]

**Description**

**Alternative(s) considered**

**Type**

**Checklist**


workflow/src/main/java/com/google/android/fhir/workflow/FhirOperator.kt
```diff
@@ -55,6 +55,11 @@ import org.opencds.cqf.cql.evaluator.measure.r4.R4MeasureProcessor
 import org.opencds.cqf.cql.evaluator.plandefinition.r4.OperationParametersParser
 import org.opencds.cqf.cql.evaluator.plandefinition.r4.PlanDefinitionProcessor
 
+// Uses the already cached FhirContext instead of creating a new one
+// on the default protected R4FhirModelResolver() constructor.
+// This is a heavy object to load. Avoid having to create a new one.
+object cachedR4FhirModelResolver : R4FhirModelResolver(FhirContext.forR4Cached())
```
jingtang10, line 61: if we are gonna make this a top level object i feel we should make this initialised lazily:
```
val r4FhirModelResolver : R4FhirModelResolver by lazy { ... }
```

but i don't even know if that's necessary - do we need R4FhirModelResolver also cached or just a matter of using the cached version of the FhirContext? if so, can we just simply modify `R4FhirModelResolver()` to `R4FhirModelResolver(FhirContext.forR4Cached())` in line 71?

jingtang10 (changes requested):

workflow/src/main/java/com/google/android/fhir/workflow/FhirOperator.kt
vitorpamplona: The constructor R4FhirModelResolver(cxt) is protected. So, the only way to access it is by extending the class, which is what the line does behind the scenes. 

Aren't all singletons (object defs) lazy by default?

vitorpamplona: I just run a few tests and I can confirm `object` will be initialized whenever the (singleton) object is called the first time, as static objects are lazily initialized.

jingtang10 (approved):

PR #1787, opened 2023-01-03
Pass original response items instead of copy to getEnabledResponses

### LINKED ISSUE #1776: Repeated Group: Answers within group are not present in the QuestionnaireResponse, if child question depends on enableWhen condition of another child question
**Describe the bug**
Suppose a child question depends on enableWhen condition of another child question and both the questions are part of repeated groups, answers are not present in the QuestionnaireResponse for the question which depends on enableWhen condition.

**Component**
SDC library

**To Reproduce**
Use [test-questionnaire.txt](https://github.com/google/android-fhir/files/10284902/test-questionnaire.txt) as questionnaire json.

https://user-images.githubusercontent.com/82660329/209086244-0d64468c-d58a-46aa-9fc8-674cd584a930.mp4

@jingtang10

joiskash:

Fixes #1776 

**Description**
In `getQuestionnaireResponse` we are passing the questionnaire response items of the copy of the quesitonnaireResponse and in the EnablementEvaluator we create the `questionnaireResponseItemPreOrderList` and `questionnaireResponseItemParentMap` from the questionnaireResponse. This would result in itemIndex = -1 for all indexOf search, since we are trying to search for objects in the original list whereas the list item that is being searched for is part of the copy list. 
The code change is to pass `this@QuestionnaireViewModel.questionnaireResponse.item` . This will search for items in the original list and the enablementEvaluator will work as expected

**Alternative(s) considered**
No

**Type**
Choose one: Bug fix 

**Checklist**


jingtang10:
ok so i think the root cause of this bug is actually here:

https://github.com/google/android-fhir/blob/854c23cfbd63f01f183b60aa06e572500b7271fe/datacapture/src/main/java/com/google/android/fhir/datacapture/QuestionnaireViewModel.kt#L634

where we are creating the enablement evaluator from the questionnaire response in the view model.

this kinda means that the current interface (if we see it as an internal/private interface) of `getEnabledResponseItems` is kinda broken. it only works if questionnaire response items passed to it belong to the questionnaire response in the view model.

maybe create another wrapper function in which you create the evaluator and then make the current function take an enablement evaluator with the condition that it is created from the same questionnaire response?

jingtang10 (changes requested):
see comment.

also would be excellent if we have a test case that can catch this.

PR #1840, opened 2023-02-02
Fix for removing old indexes when resource is updated

### LINKED ISSUE #1825: Old and outdated resource indexes still present in the database
**Describe the bug**
FHIR Engine search results sometimes returns outdated results because old indexes are not deleted

**Component**
Core library

**To Reproduce**
1. Select a resource to test with by it's ID and resourceUuid from the database
2. Write code to fetch and call `fhirEngine.update` on the same resource
3. Check the  `DateTimeIndexEntity` table using Android Studio app inspection for the `_lastUpdated` indexes for the specific resource
4. You'll realize that there are multiple indexes for the same property

**Expected behavior**

Inspecting the database should show only a single record for the resource's `_lastUpdated` index. This should be the latest index

**Screenshots**
N/A

 - Device: Nexus 5X API 28 emulator
 - Android version: 9
 - Build number: PSR1.180720.117
 - Version 28

**Additional context**

aditya-07:

Fixes #1825

**Description**
Issue: Updating resource via `FhirEngine.update ` api doesn't delete previous indexes.

Solution: In the database, for the `ResourceEntity` do `Insert` instead of `Update` as it results in the Deletion of old `ResourceEntity` (based on ConflictStrategy) that causes of its `IndexEntities` to get deleted as well based on their `ForeignKey` constrain.

**Alternative(s) considered**

The other approach is to have a `Unique Index` for each IndexEntity that will Delete it based on the `insert*Index`'s `OnConflictStrategy`. 

Proposed solution is inline with how we handle updates for remote updates.

**Type**
Choose one: Bug fix

**Checklist**

### DIFF
--- a/engine/src/main/java/com/google/android/fhir/db/impl/dao/ResourceDao.kt
+++ b/engine/src/main/java/com/google/android/fhir/db/impl/dao/ResourceDao.kt
@@ -54,8 +54,12 @@ internal abstract class ResourceDao {
   open suspend fun update(resource: Resource) {
     getResourceEntity(resource.logicalId, resource.resourceType)?.let {
       val entity = it.copy(serializedResource = iParser.encodeResourceToString(resource))
+      // The foreign key in Index entity tables is set with cascade delete constraint and
+      // insertResource has REPLACE conflict resolution. So, when we do an insert to update the
+      // resource, it deletes old resource and corresponding index entities (based on foreign key
+      // constrain) before inserting the new resource.
       insertResource(entity)
-      val index = ResourceIndexer.index(resource)
+      val index = resourceIndexer.index(resource)
       updateIndicesForResource(index, entity, it.resourceUuid)
     }
       ?: throw ResourceNotFoundException(resource.resourceType.name, resource.id)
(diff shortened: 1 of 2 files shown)


engine/src/main/java/com/google/android/fhir/db/impl/dao/ResourceDao.kt
```diff
@@ -51,22 +51,9 @@ internal abstract class ResourceDao {
...
-          id = 0,
-          resourceType = resource.resourceType,
-          resourceUuid = it.resourceUuid,
-          resourceId = resource.logicalId,
-          serializedResource = iParser.encodeResourceToString(resource),
-          versionId = it.versionId,
-          lastUpdatedRemote = it.lastUpdatedRemote
-        )
+      val entity = it.copy(serializedResource = iParser.encodeResourceToString(resource))
+      insertResource(entity)
```
jingtang10, line 56: can you comment on the fact that we can do this because we have the conflict resolution strategy of replace in the resoure table?

jingtang10 (approved):
Thanks for fixing this!

engine/src/main/java/com/google/android/fhir/db/impl/dao/ResourceDao.kt
aditya-07: Done

PR #1938, opened 2023-03-27
Add _tag and _profile search parameters for x-fhir-query

### LINKED ISSUE #1935: Support for _tag and _profile as search parameters in x-fhir-query expression
Currently, it is not possible to use _tag or _profile with x-fhir-query for example you can not have Location?_tag=salima-catchments with within x-fhir-query expression. You will get an exception that the _tag not found in resource Location.

Be able to tag resources and use the tags to filter them using x-fhir-query expressions

**Additional context**

I have a potential fix for this issue

gosso22:

Fixes #1935

**Description**
In the toSearchParamDefinitionValueMap method, I added _tag and _profile in the paramDefinition.

**Alternative(s) considered**

**Type**

**Checklist**


engine/src/main/java/com/google/android/fhir/search/query/XFhirQueryTranslator.kt
```diff
@@ -182,7 +182,20 @@ object XFhirQueryTranslator {
     type: ResourceType
   ): List<Pair<SearchParamDefinition, String>> {
     return this.map { (paramKey, paramValue) ->
-      val paramDefinition = paramKey.toSearchParamDefinition(type)
+      val paramDefinition =
+        when (paramKey) {
+          "_tag" ->
```
maimoonak, line 187: We can add _tag and _ profile as Constants?

engine/src/test/java/com/google/android/fhir/search/query/XFhirQueryTranslatorTest.kt
```diff
@@ -337,4 +337,28 @@ class XFhirQueryTranslatorTest {
       }
     assertThat(exception.message).isEqualTo("SPECIAL type not supported in x-fhir-query")
   }
+
+  @Test
```
maimoonak, line 341: Add a test in com.google.android.fhir.impl.FhirEngineImplTest#search() by x-fhir-query should return female patients for gender param as well for _tag and _profile for integration testing.

maimoonak (changes requested):

engine/src/test/java/com/google/android/fhir/impl/FhirEngineImplTest.kt
```diff
@@ -253,6 +255,79 @@ class FhirEngineImplTest {
     assertThat(exception.message).isEqualTo("customParam not found in Patient")
   }
 
+  @Test
+  fun `search() by x-fhir-query should return patients for gender param as well for _tag`() =
```
maimoonak, line 259: Keep it simple. Just run it for tag. Remove gender based search and properties

engine/src/test/java/com/google/android/fhir/impl/FhirEngineImplTest.kt
```diff
@@ -253,6 +255,79 @@ class FhirEngineImplTest {
...
+
+      val result =
+        fhirEngine.search("Patient?gender=female&_tag=cardiology-department").map { it as Patient }
+
+      assertThat(result.size).isEqualTo(1)
+      assertThat(result.all { it.gender == Enumerations.AdministrativeGender.FEMALE }).isTrue()
+    }
+
+  @Test
+  fun `search() by x-fhir-query should return patients for gender param as well for _profile`() =
```
maimoonak, line 295: same as above comment. Keep it simple and just test profile. Remove gender

maimoonak (changes requested):

engine/src/main/java/com/google/android/fhir/search/query/XFhirQueryTranslator.kt
```diff
@@ -182,7 +184,20 @@ object XFhirQueryTranslator {
     type: ResourceType
   ): List<Pair<SearchParamDefinition, String>> {
     return this.map { (paramKey, paramValue) ->
-      val paramDefinition = paramKey.toSearchParamDefinition(type)
+      val paramDefinition =
+        when (paramKey) {
+          XFHIR_QUERY_TAG_PARAM ->
+            SearchParamDefinition("_tag", Enumerations.SearchParamType.TOKEN, "$type.meta.tag")
```
maimoonak, line 190: use XFHIR_QUERY_TAG_PARAM constant here... do the same for profile

engine/src/test/java/com/google/android/fhir/impl/FhirEngineImplTest.kt
```diff
@@ -253,6 +255,80 @@ class FhirEngineImplTest {
...
+            Meta()
+              .setProfile(
+                mutableListOf(
+                  CanonicalType(
+                    "http://fhir.org/STU3/StructureDefinition/Example-Patient-Profile-1"
+                  )
+                )
+              )
+        },
+        buildPatient("4", "C", Enumerations.AdministrativeGender.MALE).apply {
```
maimoonak, line 306: change it FEMALE so that only difference is profile

engine/src/main/java/com/google/android/fhir/search/query/XFhirQueryTranslator.kt
```diff
@@ -182,7 +184,24 @@ object XFhirQueryTranslator {
...
-      val paramDefinition = paramKey.toSearchParamDefinition(type)
+      val paramDefinition =
+        when (paramKey) {
+          XFHIR_QUERY_TAG_PARAM ->
+            SearchParamDefinition(
+              XFHIR_QUERY_TAG_PARAM,
+              Enumerations.SearchParamType.TOKEN,
+              "$type.meta.tag"
+            )
+          XFHIR_QUERT_PROFILE_PARAM ->
```
maimoonak, line 195: Rename spelling to quer**Y**.. its **T** here...

maimoonak (approved):
Just one issue with spelling.. rest is fine.

gosso22:
@jingtang10 - can you approve and merge these changes?

engine/src/test/java/com/google/android/fhir/impl/FhirEngineImplTest.kt
```diff
@@ -253,6 +255,80 @@ class FhirEngineImplTest {
...
+                mutableListOf(
+                  Coding("https://d-tree.org/", "cardiology-department", "Cardiology Dept")
+                )
+              )
+        },
+        buildPatient("4", "C", Enumerations.AdministrativeGender.FEMALE).apply {
+          meta =
+            Meta()
+              .setTag(
+                mutableListOf(Coding("http://d-tree.org/", "salima-catchment", "Salima Patients"))
```
MJ1998, line 275: I think we should have generalised names like Tag1, Patient1

engine/src/main/java/com/google/android/fhir/search/query/XFhirQueryTranslator.kt
```diff
@@ -47,6 +47,8 @@ import org.hl7.fhir.r4.model.ResourceType
 object XFhirQueryTranslator {
   private const val XFHIR_QUERY_SORT_PARAM = "_sort"
   private const val XFHIR_QUERY_COUNT_PARAM = "_count"
+  private const val XFHIR_QUERY_TAG_PARAM = "_tag"
+  private const val XFHIR_QUERY_PROFILE_PARAM = "_profile"
```
MJ1998, line 51: Can we use com.google.android.fhir.index.getSearchParamList instead ?

MJ1998 (changes requested):
So `_tag` and `_profile` fall under standard search parameters for all resources.
link: https://hl7.org/fhir/search.html#all
For these standard search params, we already have `SearchParamDefinition` defined in `getSearchParamList` under "Resource" option. So instead of rewriting this code we can make advantage of that.
source code: com.google.android.fhir.index.getSearchParamList

gosso22:
> So `_tag` and `_profile` fall under standard search parameters for all resources. link: https://hl7.org/fhir/search.html#all For these standard search params, we already have `SearchParamDefinition` defined in `getSearchParamList` under "Resource" option. So instead of rewriting this code we can make advantage of that. source code: com.google.android.fhir.index.getSearchParamList

@MJ1998 - Thank you for looking at this and for the feedback. XFhirQueryTranslator is already using the method com.google.android.fhir.index.getSearchParamList [here](https://github.com/google/android-fhir/blob/96af7f12cc8f95a2922dd184ef969745a544641a/engine/src/main/java/com/google/android/fhir/search/query/XFhirQueryTranslator.kt#L177), I expected it to pick _tag and _profile however this is not happening and it throws an error that _tag is not found in ResourceName. I am not sure why that is happening, that's why I decided to handle it explicitly.

MJ1998:
> > So `_tag` and `_profile` fall under standard search parameters for all resources. link: https://hl7.org/fhir/search.html#all For these standard search params, we already have `SearchParamDefinition` defined in `getSearchParamList` under "Resource" option. So instead of rewriting this code we can make advantage of that. source code: com.google.android.fhir.index.getSearchParamList
> 
> @MJ1998 - Thank you for looking at this and for the feedback. XFhirQueryTranslator is already using the method com.google.android.fhir.index.getSearchParamList [here](https://github.com/google/android-fhir/blob/96af7f12cc8f95a2922dd184ef969745a544641a/engine/src/main/java/com/google/android/fhir/search/query/XFhirQueryTranslator.kt#L177), I expected it to pick _tag and _profile however this is not happening and it throws an error that _tag is not found in ResourceName. I am not sure why that is happening, that's why I decided to handle it explicitly.

@gosso22 Either we need to modify static method `resourceSearchParameters` so that it searches for standard parameters under "Resource" first and then for the specific resourceType (eg, "Patient") OR we need to modify `SearchParameterRepositoryGenerator.kt` to generate function `getSearchParamList` that also searches for standard parameters under "Resource". I prefer the latter solution.

gosso22:
> > > So `_tag` and `_profile` fall under standard search parameters for all resources. link: https://hl7.org/fhir/search.html#all For these standard search params, we already have `SearchParamDefinition` defined in `getSearchParamList` under "Resource" option. So instead of rewriting this code we can make advantage of that. source code: com.google.android.fhir.index.getSearchParamList
> > 
> > 
> > @MJ1998 - Thank you for looking at this and for the feedback. XFhirQueryTranslator is already using the method com.google.android.fhir.index.getSearchParamList [here](https://github.com/google/android-fhir/blob/96af7f12cc8f95a2922dd184ef969745a544641a/engine/src/main/java/com/google/android/fhir/search/query/XFhirQueryTranslator.kt#L177), I expected it to pick _tag and _profile however this is not happening and it throws an error that _tag is not found in ResourceName. I am not sure why that is happening, that's why I decided to handle it explicitly.
> 
> @gosso22 Either we need to modify static method `resourceSearchParameters` so that it searches for standard parameters under "Resource" first and then for the specific resourceType (eg, "Patient") OR we need to modify `SearchParameterRepositoryGenerator.kt` to generate function `getSearchParamList` that also searches for standard parameters under "Resource". I prefer the latter solution.

@MJ1998 - I am also in favor of the latter solution.

gosso22:
@MJ1998 - I spent sometime to look on how to do the suggested:
_we need to modify SearchParameterRepositoryGenerator.kt to generate function getSearchParamList that also searches for standard parameters under "Resource"._
My thinking is to include the search parameters included in Resource for example _tag, _profile to also be available for other resources in such a way that when I do getSearchParamList for Location, _tag and _profile are also included with their path defined as Location.meta.tag and Location.meta.profile respectively. What do you think? Would this work?

MJ1998:
@gosso22 
Apologies for late response.
Yes your thinking is correct. We are on the same page. But we would like to modify the generation of `getSearchParamList` in a way to search for only a subset of [standard search parameters](https://hl7.org/fhir/search.html#all) for every resource.
`SearchParameterRepository_Generated:getResource` is already a subset of the list given in the above link. 
@jingtang10 Do you think subset defined in `SearchParameterRepository_Generated:getResource` is enough ? More technically, do you think modifying generator to add statement `addAll(getResource())` is the right thing to do ?

engine/src/test/java/com/google/android/fhir/index/SearchParameterRepositoryGeneratedTest.kt
```diff
@@ -32,22 +32,68 @@ class SearchParameterRepositoryGeneratedTest(private val resource: Resource) {
...
-      .filter { it.path.isNotEmpty() }
-      .map {
-        SearchParamDefinition(
-          it.name,
-          Enumerations.SearchParamType.valueOf(it.type.toUpperCase()),
-          it.path
-        )
-      }
-      .toMutableList()
+    val searchParams = mutableListOf<SearchParamDefinition>()
```
MJ1998, line 35: Thanks for modifying this test! 
The current set of supported search parameters for all resources are given in codegen/src/main/res/search-parameters.json. We would somehow like to use that instead of hardcoding the subset here.

buildSrc/src/main/kotlin/codegen/SearchParameterRepositoryGenerator.kt
```diff
@@ -100,6 +110,19 @@ internal object SearchParameterRepositoryGenerator {
...
+      // add search parameters defined in base -> Resource to this resource
+      if (resource != "Resource") {
+        definitions.addAll(
+          baseResourceSearchParameters.map { baseSearchParam ->
+            SearchParamDefinition(
+              className = baseSearchParam.className,
+              name = baseSearchParam.name,
+              paramTypeCode = baseSearchParam.paramTypeCode,
+              path = baseSearchParam.path.replace("Resource", resource)
+            )
```
MJ1998, line 122: This is nice and definitely one way of solving! 
But what if instead of adding duplicate codes to every function we just add a new statement like ```"addAll(getResource())"```  ? (Maybe on line 120 of the original file)
We would also **not** need to store baseResourceSearchParams then. Unless storing it makes sense for using it elsewhere.

MJ1998 (dismissed):

buildSrc/src/main/kotlin/codegen/SearchParameterRepositoryGenerator.kt
gosso22: @MJ1998 - I thought about adding `addAll(getResource())`, but I think the challenge there is the path. For example if we consider tag for  Location resource, with the above statement, the path for tag will be Resource.meta.tag instead of Location.meta.tag. I am no so sure if this works.

engine/src/test/java/com/google/android/fhir/index/SearchParameterRepositoryGeneratedTest.kt
gosso22: For this one, what I found is the resources defined in model classes  inherits from Resource which does not have these standard search parameters annotated `SearchParamDefinition`. I am not sure how to proceed with this one.

buildSrc/src/main/kotlin/codegen/SearchParameterRepositoryGenerator.kt
MJ1998: That's correct. Resolving.

engine/src/test/java/com/google/android/fhir/index/SearchParameterRepositoryGeneratedTest.kt
MJ1998: @jingtang10

jingtang10: i don't think modifying that file would work for this test case. the "expected" search parameters here are generated from using reflection on HAPI structures java classes... so we'd have to hardcode somehow these search parameters that don't exist in those java classes.

i do think @MJ1998 is right in the way that these can be extracted to a separate function just to make it a bit more explicit, something like `fun getBaseSearchParameters(...)`... and add some comments to say that we do this because these don't exist in the HAPI java classes...

buildSrc/src/main/kotlin/codegen/SearchParameterRepositoryGenerator.kt
jingtang10, line 1: optional:

explore the alternative approach of creating a new function that returns the base search parametesr for all resources, and calling that function in `getSearchParamList()` so that you're not adding the same code to each function for each resoure type.

jingtang10 (changes requested):
i think you might need to change/remove this function `addIndexesFromResourceClass` in `ResourceIndexer.kt` because now the base search parameters are already provided for each resource type when the indexer calls `getSearchParamList(...)` function.

hopefullly you won't need to change the indexer's tests much.

with this change, instead of adding `_tag` and `_profile`, or even `_lastUpdated` etc in the indexer manually, we're  including these base search paramters for each resource type in the search parameter repository. this is a *very good* change - makes our code more unfied.

thanks @gosso22!

gosso22:
> i think you might need to change/remove this function `addIndexesFromResourceClass` in `ResourceIndexer.kt` because now the base search parameters are already provided for each resource type when the indexer calls `getSearchParamList(...)` function.
> 
> hopefullly you won't need to change the indexer's tests much.
> 
> with this change, instead of adding `_tag` and `_profile`, or even `_lastUpdated` etc in the indexer manually, we're including these base search paramters for each resource type in the search parameter repository. this is a **very good** change - makes our code more unfied.
> 
> thanks @gosso22!

@jingtang10 - I removed the above function `addIndexesFromResourceClass` in the `ResourceIndexer.kt`, and change a few things. First I changed the type for _profile in `search-parameters.json` from uri to reference as defined in the fhir specification and also as it was implemented in `addIndexesFromResourceClass`. Also I had to handle the case for _id. But other than that all tests passed without the `addIndexesFromResourceClass`.

jingtang10 (changes requested):

gosso22:
@jingtang10, @aditya-07 - I have updated this, I think it is ready.

aditya-07 (changes requested):

buildSrc/src/main/resources/search-parameters.json
```diff
@@ -189,7 +189,7 @@
       "description" : "Profiles this resource claims to conform to",
       "code" : "_profile",
       "base" : ["Resource"],
-      "type" : "uri",
+      "type" : "reference",
```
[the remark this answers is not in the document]
aditya-07: I am not sure about the rationale behind **_profile** being added as a reference in [here](https://github.com/google/android-fhir/blob/f05cd76e84b91643313093bc394128f1303efcdc/engine/src/main/java/com/google/android/fhir/index/ResourceIndexer.kt#L144). But, I think we should keep the search-parameters.json as-is since its dowloaded directly from the [registry](https://www.hl7.org/fhir/searchparameter-registry.html).

jingtang10 (commented):
Thanks for this change! I love this PR!

Once @aditya-07 's comment is addressed in the json file i'm happy to approve this!

buildSrc/src/main/resources/search-parameters.json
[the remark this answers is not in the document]
gosso22: @jingtang10 and @aditya-07 - I have resolved this and modified the tests that were using reference.

buildSrc/src/main/kotlin/codegen/SearchParameterRepositoryGenerator.kt
```diff
@@ -41,8 +42,10 @@ import org.hl7.fhir.r4.model.SearchParameter
  * should be regenerated to reflect any change.
  *
  * To do this, replace the content of the file `codegen/src/main/res/search-parameters.json` with
- * the content at `http://www.hl7.org/fhir/search-parameters.json` and run the `main` function in
- * the `codegen` module.
+ * the content at `http://www.hl7.org/fhir/search-parameters.json` and execute the gradle task
+ * `generateSearchParamsTask`. If you are using Android Studio, you can usually find this task in
```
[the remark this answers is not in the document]
gosso22: Thanks @jingtang10

jingtang10 (approved):

aditya-07 (approved):

PR #1959, opened 2023-04-07
Cqf expression implementation for dynamic question title

### LINKED ISSUE #1945: CQF expression support
Display `questionnaire.item.text` from FHIRPath expression. In a nutshell, it's like calculated expression which's for answer, but this time for `questionnaire.item.text`.

Add support for CQF expression to do this.

N/A

**Additional context**
- The name "CQF" expression is a bit misleading. It was named that way because it's initially from CQL IG. But FHIRPath can be used as well.
- To do that exact scenario, CQF expression was suggested by Lloyd via [Zulip chat](https://chat.fhir.org/#narrow/stream/179255-questionnaire/topic/Questionnaire.2Eitem.2Etext.20from.20previously.20answered.20questions)
- CQF expression [FHIR doc](http://hl7.org/fhir/R4B/extension-cqf-expression.html)
- CQF expression is also part of the [Advanced Rendering Questionnaire](https://build.fhir.org/ig/HL7/sdc/StructureDefinition-sdc-questionnaire-render.html)

Yeah

maimoonak:
Fixes #1945 

**Description**
Display questionnaire.item.text from FHIRPath expression. In a nutshell, it's like calculated expression which's for answer, but this time for questionnaire.item.text. [Advance rendering questionnaire](https://build.fhir.org/ig/HL7/sdc/StructureDefinition-sdc-questionnaire-render.html )allows the cf-expression on title to display it based on other question`s answers [example](https://build.fhir.org/ig/HL7/sdc/examples.html#using-text-and-cqf-expression) 

**Type**
Choose one: Feature

**Checklist**


FikriMilano:
@maimoonak awesome work here!
I will spare some time tomorrow to review this.

FikriMilano (changes requested):
nicely done! 🚀

can you write a test that makes sure when an answer from a QuestionnaireItem was changed, the next QuestionnaireItem.title is updated?

FikriMilano (commented):
@maimoonak can you also check if the dynamic title works with variable expression?
Try to save the answer to variable expression, then call that variable in cqf expression.

I tried this, it doesn't seem to work, though it works with calculated expression

maimoonak:
Good catch! Thanks for this. Feedback resolved and PR is ready for review

jingtang10:
@shelaghm please take a look at the video @maimoonak attached.

datacapture/src/main/java/com/google/android/fhir/datacapture/views/QuestionnaireViewItem.kt
```diff
@@ -80,6 +86,7 @@ data class QuestionnaireViewItem(
     {
       emptyList()
     },
+  private val resolveCqfExpression: (Element) -> List<Base> = { emptyList() },
```
jingtang10, line 89: i don't think this should be a callback in the questionnaire view item. i know that the other callbacks (`resolveAnswerValueSet`, `resolveAnswerExpression`) may be pointing to this pattern... but i think cql expression should be resolved in questionnaire view model already, and this should probably be totally opaque to the UI.

so either create a new extension property in questionnaire item, or create another data field in questionnaire view item. i would probably prefer the former rather than the latter.

datacapture/src/main/java/com/google/android/fhir/datacapture/QuestionnaireViewModel.kt
```diff
@@ -533,6 +537,24 @@ internal class QuestionnaireViewModel(application: Application, state: SavedStat
     return options
   }
 
+  internal fun resolveCqfExpression(
+    questionnaireItem: QuestionnaireItemComponent,
+    questionnaireResponseItem: QuestionnaireResponseItemComponent,
+    element: Element,
```
jingtang10, line 543: cqf expression is on text only right? i think it doesn't need to be an element. in fact this can probably be an extension function of questionnaire item...

see my comment below.

jingtang10 (changes requested):
thanks for this pr @maimoonak! great work!

some high-level comments first to address. thanks!

catalog/src/main/res/drawable/ic_dynamic_title_behavior.xml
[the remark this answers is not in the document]
shelaghm: Is it this? I think we created this for "range picker" but it also works here! Screenshot 2023-04-24 at 17 01 41

shelaghm:
@maimoonak Great work! Thanks for adding the video. Looks good to me. +1 to Jing's suggestion to use generic copy.

FikriMilano (approved):
@maimoonak approved 👍
nice work :D

datacapture/src/main/java/com/google/android/fhir/datacapture/views/QuestionnaireViewItem.kt
maimoonak: The cqf-expression is an extension on textElement (nested on another element level down the hierarchy). This approach is changed to add text on response-item if it is supposed to be dynamic. We have to output it into response as well at the end of submission.

datacapture/src/main/java/com/google/android/fhir/datacapture/QuestionnaireViewModel.kt
maimoonak: cqf-expression https://build.fhir.org/ig/HL7/sdc/behavior.html#cqf-expression is very dynamic and flexible and can be present on various elements including questionnaireItem or its nested elements which can include minOccurs, maxOccurs, required, repeats, readOnly , answerValueSet, text. This means that this expression would run on any "Element" and not only on "QuestionnaireItemComponent"

maimoonak:
@jingtang10 Kindly check the new approach where we are using questionnaire-reponse text as primary source for dynamic title.

santosh-pingle:
I am reviewing this, thanks!

datacapture/src/main/java/com/google/android/fhir/datacapture/QuestionnaireViewModel.kt
```diff
@@ -674,6 +696,14 @@ internal class QuestionnaireViewModel(application: Application, state: SavedStat
       } else {
         NotValidated
       }
+
+    // set title dynamically from cqf expression on textElement
+    questionnaireResponseItem.apply {
+      resolveCqfExpression(questionnaireItem, this, questionnaireItem.textElement)
+        .firstOrNull()
+        ?.let { text = it.primitiveValue() }
+    }
```
FikriMilano, line 705: I like the addition of this, now we can see the selected answers in the QuestionnaireResponse, which doesn't happen previously 👍

FikriMilano (approved):
@maimoonak Thanks, I find the new change very useful :D
Our use case doesn't seem to be affected; it's getting better instead, and the new code looks great to me 👍

cc @f-odhiambo

catalog/src/main/res/values/strings.xml
```diff
@@ -43,6 +43,9 @@
     <string
         name="behavior_name_skip_logic_info"
     >If Yes is selected, a follow-up question is displayed. If No is selected, no follow-up questions are displayed.</string>
+    <string
+        name="behavior_name_dynamic_question_title"
+    >Dynamic Question Title</string>
```
jingtang10, line 48:
```suggestion
    >Dynamic Question Text</string>
```

datacapture/src/main/java/com/google/android/fhir/datacapture/views/QuestionnaireViewItem.kt
```diff
@@ -199,6 +202,15 @@ data class QuestionnaireViewItem(
         }
       }
 
+  /**
+   * Fetches the question title that should be displayed to user. The title is fetched from
+   * [Questionnaire.QuestionnaireResponseItemComponent] if exists, otherwise it is derived from
+   * translatable textElement property of [QuestionnaireResponse.QuestionnaireItemComponent]
+   */
+  internal val questionTitle: Spanned? by lazy {
```
jingtang10, line 210:
```suggestion
  internal val questionText: Spanned? by lazy {
```

datacapture/src/main/java/com/google/android/fhir/datacapture/views/QuestionnaireViewItem.kt
```diff
@@ -210,7 +222,8 @@ data class QuestionnaireViewItem(
    */
   internal fun hasTheSameItem(other: QuestionnaireViewItem) =
     questionnaireItem === other.questionnaireItem &&
-      questionnaireResponseItem === other.questionnaireResponseItem
+      questionnaireResponseItem === other.questionnaireResponseItem &&
+      questionTitle == other.questionTitle
```
jingtang10, line 226: i think this change needs to be in `hasTheSameAnswer`.

and you can probably change the name of that function to `hasTheSameResponse`

jingtang10 (changes requested):

datacapture/src/main/java/com/google/android/fhir/datacapture/views/HeaderView.kt
```diff
@@ -49,7 +48,7 @@ internal class HeaderView(context: Context, attrs: AttributeSet?) : LinearLayout
       questionnaireItem = questionnaireViewItem.questionnaireItem
     )
     prefix.updateTextAndVisibility(questionnaireViewItem.questionnaireItem.localizedPrefixSpanned)
-    question.updateTextAndVisibility(questionnaireViewItem.questionnaireItem.localizedTextSpanned)
+    question.updateTextAndVisibility(questionnaireViewItem.questionTitle)
```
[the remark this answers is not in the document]
santosh-pingle: please add also add unit test to just check cqfExpression text value is shown over questionnaireItem.localizedTextSpanned

datacapture/src/main/java/com/google/android/fhir/datacapture/views/QuestionnaireViewItem.kt
```diff
@@ -199,6 +202,15 @@ data class QuestionnaireViewItem(
         }
       }
 
+  /**
+   * Fetches the question title that should be displayed to user. The title is fetched from
+   * [Questionnaire.QuestionnaireResponseItemComponent] if exists, otherwise it is derived from
+   * translatable textElement property of [QuestionnaireResponse.QuestionnaireItemComponent]
+   */
```
santosh-pingle, line 209: Could you please update the documentation to include that if a CQF expression result is present, it will be used as the question text, otherwise, localizedTextSpanned will be used?

datacapture/src/main/java/com/google/android/fhir/datacapture/views/HeaderView.kt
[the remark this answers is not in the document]
santosh-pingle: Maybe you need to update the same change in `GroupHeaderView` class as well.

datacapture/src/main/java/com/google/android/fhir/datacapture/fhirpath/ExpressionEvaluator.kt
```diff
@@ -81,6 +83,40 @@ object ExpressionEvaluator {
...
+          this
+        )
+      }
+    return fhirPathEngine.evaluate(
+      appContext,
+      questionnaireResponse,
+      null,
+      questionnaireResponseItem,
+      expression.expression
+    )
```
santosh-pingle, line 117: optional : If you think existing code is more readable, then in that case you can please ignore this suggestion.

```suggestion
    val appContext = mutableMapOf<String, Base?>()
    return with(appContext) {
      extractDependentVariables(
        expression,
        questionnaire,
        questionnaireResponse,
        questionnaireItemParentMap,
        questionnaireItem,
        appContext
      )
      fhirPathEngine.evaluate(
        appContext,
        questionnaireResponse,
        null,
        questionnaireResponseItem,
        expression.expression
      )
    }
```

santosh-pingle (approved):
@maimoonak thanks!
Is there a way to add unit test to `HeaderViewTest` and 'GroupHeaderViewTest' where dynamic text from cqf expression can be shown, or android test?
otherwise it looks good to me.

datacapture/src/main/java/com/google/android/fhir/datacapture/extensions/MoreExpressions.kt
```diff
@@ -16,7 +16,18 @@
...
+import org.hl7.fhir.r4.utils.ToolingExtensions
+
+internal const val EXTENSION_CQF_EXPRESSION_URL: String =
+  "http://hl7.org/fhir/StructureDefinition/cqf-expression"
+
+internal val Element.cqfExpression: Expression?
+  get() =
+    ToolingExtensions.getExtension(this, EXTENSION_CQF_EXPRESSION_URL)?.value?.let {
+      it.castToExpression(it)
+    }
```
jingtang10, line 30: Can you put this in a new file called `MoreElements.kt`?

datacapture/src/main/java/com/google/android/fhir/datacapture/fhirpath/ExpressionEvaluator.kt
```diff
@@ -81,6 +83,40 @@ object ExpressionEvaluator {
...
+        appContext
+      )
+      fhirPathEngine.evaluate(
+        appContext,
+        questionnaireResponse,
+        null,
+        questionnaireResponseItem,
+        expression.expression
+      )
+    }
```
jingtang10, line 117: i don't see why `with` here is more readable? we're referring to `appContext` within the `with` block as `appContext` anyway? at the very least it could be `this`?

but i think we should just not use `with` here. i don't see the value.

datacapture/src/main/java/com/google/android/fhir/datacapture/fhirpath/ExpressionEvaluator.kt
```diff
@@ -81,6 +83,40 @@ object ExpressionEvaluator {
       }
   }
 
+  /**
+   * Returns a list of [Base] as the evaluated value for cqf expression extension. Expression runs
+   * over questionnaireResponse with fhirpath supplements values
+   * https://build.fhir.org/ig/HL7/sdc/expressions.html#fhirpath-supplements. %resource =
+   * [QuestionnaireResponse] %context = [QuestionnaireResponseItemComponent]
```
jingtang10, line 90:
```suggestion
   * Returns the evaluation result of the CQF expression.
   * 
   * FHIRPath supplements are handled according to
   * https://build.fhir.org/ig/HL7/sdc/expressions.html#fhirpath-supplements.
   *
   * %resource = [QuestionnaireResponse]
   * %context = [QuestionnaireResponseItemComponent]
```

datacapture/src/main/java/com/google/android/fhir/datacapture/fhirpath/ExpressionEvaluator.kt
```diff
@@ -81,6 +83,40 @@ object ExpressionEvaluator {
...
+   */
+  fun evaluateCqfExpression(
```
jingtang10, line 92: is this actually a function for cql expression or any expression would work?

datacapture/src/main/java/com/google/android/fhir/datacapture/QuestionnaireViewModel.kt
```diff
@@ -534,6 +538,24 @@ internal class QuestionnaireViewModel(application: Application, state: SavedStat
     return options
   }
 
+  private fun resolveCqfExpression(
+    questionnaireItem: QuestionnaireItemComponent,
+    questionnaireResponseItem: QuestionnaireResponseItemComponent,
+    element: Element,
+  ): List<Base> {
+    val cqfExpression = element.cqfExpression ?: return emptyList()
+    if (cqfExpression.isFhirPath) {
```
jingtang10, line 547: can you flip this `if` so that we throw the exception if it's not fhir path.

and you don't need `else` any more.

datacapture/src/main/java/com/google/android/fhir/datacapture/QuestionnaireViewModel.kt
```diff
@@ -675,6 +697,14 @@ internal class QuestionnaireViewModel(application: Application, state: SavedStat
       } else {
         NotValidated
       }
+
+    // set title dynamically from cqf expression on textElement
```
jingtang10, line 701:
```suggestion
    // Set question text dynamically from CQL expression
```

datacapture/src/main/java/com/google/android/fhir/datacapture/views/GroupHeaderView.kt
```diff
@@ -48,7 +47,8 @@ internal class GroupHeaderView(context: Context, attrs: AttributeSet?) :
       questionnaireItem = questionnaireViewItem.questionnaireItem
     )
     prefix.updateTextAndVisibility(questionnaireViewItem.questionnaireItem.localizedPrefixSpanned)
-    question.updateTextAndVisibility(questionnaireViewItem.questionnaireItem.localizedTextSpanned)
+    // use questionText to ensure cqf-expression derived text is prioritized over localized text
```
jingtang10, line 50:
```suggestion
    // CQF expression takes precedence over static question text
```

datacapture/src/main/java/com/google/android/fhir/datacapture/views/HeaderView.kt
```diff
@@ -49,7 +48,8 @@ internal class HeaderView(context: Context, attrs: AttributeSet?) : LinearLayout
       questionnaireItem = questionnaireViewItem.questionnaireItem
     )
     prefix.updateTextAndVisibility(questionnaireViewItem.questionnaireItem.localizedPrefixSpanned)
-    question.updateTextAndVisibility(questionnaireViewItem.questionnaireItem.localizedTextSpanned)
+    // use questionText to ensure cqf-expression derived text is prioritized over localized text
```
jingtang10, line 51:
```suggestion
    // CQF expression takes precedence over static question text
```

jingtang10 (changes requested):
Great change! Thanks very much! Just a few comments below:

jingtang10 (approved):
thanks @maimoonak! great job!

Issue #1972, opened 2023-04-13
Items nested in group do not reflect in questionnaire-response with slip logic


maimoonak:
**Describe the bug**
Item which are in group / repeat group are missed when getting final questionnaire-response if any item prior to them is skipped due to any enablement condition. This leads to a critical issue of data missed from final submitted response irrespective of that the use actually filled the data.
On close inpecation and debugging of code it was not that the MoreQuestionnaireResponsesKt#unpackRepeatedGroups is the real issue where it is using zip-indexing i.e. index to index of items for unpacking. This leads to invalid mapping when one of the item was missed due to enablement and hence the next question does not map to the group and does not loop over the nested items.
The items which are not group although work fine for missed enablement condition items displayed prior to it. There is no issue in rendering and form views work fine.

**Component**
SDC library

**To Reproduce**
Replace skip-logic-behavior questionnaire with below and choose 'Yes' for first item. Then provide values for all questions. See the date is missing from final output because there are 2 other items which were not enabled on 'Yes'.
```
{
  "resourceType": "Questionnaire",
  "item": [
    {
      "extension": [ {
          "url": "http://hl7.org/fhir/StructureDefinition/questionnaire-itemControl",
          "valueCodeableConcept": {
            "coding": [
              {
                "system": "http://hl7.org/fhir/questionnaire-item-control",
                "code": "radio-button",
                "display": "Radio Button"
              }
            ]
          }
        } ], 
      "linkId": "1.0",
      "type": "choice",
      "text": "Has the client received this year's seasonal flu vaccine?",
      "answerOption": [ {
          "valueCoding": {
            "code": "Y",
            "display": "Yes",
            "system": "custom"
          }
        }, {
          "valueCoding": {
            "code": "N",
            "display": "No",
            "system": "custom"
          }
        }, {
          "valueCoding": {
            "code": "unknown",
            "display": "Unknown",
            "system": "custom"
          }
        } ], 
      "item": [ {
          "extension": [
            {
              "url": "http://hl7.org/fhir/StructureDefinition/questionnaire-displayCategory",
              "valueCodeableConcept": {
                "coding": [
                  {
                    "system": "http://hl7.org/fhir/questionnaire-display-category",
                    "code": "instructions"
                  }
                ]
              }
            }
          ],
          "linkId": "1-select-one",
          "text": "Select one",
          "type": "display"
        }
      ]
    }, {
      "text": "Confirm that vaccine date is known",
      "type": "boolean",
[...]

PR #1973, opened 2023-04-13
Fix zip index based mapping issue in uppack repeat groups extension

maimoonak:
Fixes https://github.com/google/android-fhir/issues/1972

**Description**
The issue is that the zip index mapping for questionnaire and questionnaire-response item can not  be used when unpacking the items while generating questionnaire-response. If items before the group are missed due to enablement it just skip the data entered into item embedded into groups.

**Alternative(s) considered**
zip index mapping for questionnaire and questionnaire-response item can not  be used because there could be missing items from questionnaire-response due to enablement condition

**Type**
Choose one: Bug fix

**Checklist**

### DIFF
--- a/datacapture/src/main/java/com/google/android/fhir/datacapture/extensions/MoreQuestionnaireResponses.kt
+++ b/datacapture/src/main/java/com/google/android/fhir/datacapture/extensions/MoreQuestionnaireResponses.kt
@@ -89,10 +89,12 @@ private fun unpackRepeatedGroups(
   questionnaireItems: List<Questionnaire.QuestionnaireItemComponent>,
   questionnaireResponseItems: List<QuestionnaireResponse.QuestionnaireResponseItemComponent>
 ): List<QuestionnaireResponse.QuestionnaireResponseItemComponent> {
-  return questionnaireItems.zip(questionnaireResponseItems).flatMap {
-    (questionnaireItem, questionnaireResponseItem) ->
-    unpackRepeatedGroups(questionnaireItem, questionnaireResponseItem)
-  }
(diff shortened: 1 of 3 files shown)


datacapture/src/main/java/com/google/android/fhir/datacapture/extensions/MoreQuestionnaireResponses.kt
```diff
@@ -89,10 +89,12 @@ private fun unpackRepeatedGroups(
   questionnaireItems: List<Questionnaire.QuestionnaireItemComponent>,
   questionnaireResponseItems: List<QuestionnaireResponse.QuestionnaireResponseItemComponent>
 ): List<QuestionnaireResponse.QuestionnaireResponseItemComponent> {
-  return questionnaireItems.zip(questionnaireResponseItems).flatMap {
-    (questionnaireItem, questionnaireResponseItem) ->
-    unpackRepeatedGroups(questionnaireItem, questionnaireResponseItem)
-  }
+  return questionnaireItems
+    .map { qItem -> qItem to questionnaireResponseItems.find { it.linkId == qItem.linkId } }
```
jingtang10, line 93: can you use the `zipByLinkId` method in the `QuestionnaireViewModel`?

you might need to increase its visibility.

jingtang10 (changes requested):
thanks for this @maimoonak. 

about `should not modify other items while unpacking repeated groups`: i think the test is wrong, the link ids in the questionnaire response should be non-repeated... can you fix please? thanks!

jingtang10 (approved):
thanks @maimoonak

PR #2004, opened 2023-05-10
Support sending gzipped request body when uploading

### LINKED ISSUE #1988: Support sending gzipped request body
To optimise for networks bandwidth, it would be good to support gzipping the HTTP request body in FHIR Engine

Have an option to automatically gzip the HTTP request body. This should be optional as the server might or might not be able to understand the compressed request body.

No other alternatives for this

If bandwidth allows

cc: @jingtang10

omarismail94:

Fixes #1988 

**Description**

* Support sending gzipped request body when uploading to the server
* Make uploading with gzip an option that developers can set
* No need to support download as OkHTTP does this for us automatically
* Tested on local HAPI server and it works

**Type**
Choose one: (Bug fix | **Feature** | Documentation | Testing | Code health | Builds | Releases | Other)

**Checklist**

### DIFF
--- a/engine/src/main/java/com/google/android/fhir/FhirEngineProvider.kt
+++ b/engine/src/main/java/com/google/android/fhir/FhirEngineProvider.kt
@@ -161,5 +161,7 @@ data class NetworkConfiguration(
   /** Read timeout (in seconds) for network connection. The default is 10 seconds. */
   val readTimeOut: Long = 10,
   /** Write timeout (in seconds) for network connection. The default is 10 seconds. */
-  val writeTimeOut: Long = 10
+  val writeTimeOut: Long = 10,
+  /** Compresses requests when uploading to a server that supports gzip. */
+  val uploadWithGzip: Boolean = false
 )
--- a/engine/src/main/java/com/google/android/fhir/sync/remote/GzipUploadInterceptor.kt
+++ b/engine/src/main/java/com/google/android/fhir/sync/remote/GzipUploadInterceptor.kt
@@ -1,0 +1,59 @@
+/*
+ * Copyright 2022 Google LLC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *       http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.google.android.fhir.sync.remote
+
+import okhttp3.Interceptor
+import okhttp3.MediaType
+import okhttp3.RequestBody
+import okhttp3.Response
+import okio.BufferedSink
+import okio.GzipSink
+import okio.buffer
+
+const val CONTENT_ENCODING_HEADER_NAME = "Content-Encoding"
+
+/** Compresses uploads with gzip */
+class GzipUploadInterceptor : Interceptor {
+  override fun intercept(chain: Interceptor.Chain): Response {
+    val uncompressedRequest = chain.request()
+    if (uncompressedRequest.body == null ||
+        uncompressedRequest.header(CONTENT_ENCODING_HEADER_NAME) != null
+    ) {
+      return chain.proceed(uncompressedRequest)
+    }
+
+    val compressedRequest =
+      uncompressedRequest
+        .newBuilder()
+        .header(CONTENT_ENCODING_HEADER_NAME, "gzip")
+        .method(uncompressedRequest.method, gzip(uncompressedRequest.body!!))
+        .build()
+
+    return chain.proceed(compressedRequest)
+  }
+
+  private fun gzip(body: RequestBody): RequestBody =
+    object : RequestBody() {
+      override fun contentType(): MediaType? = body.contentType()
+
+      override fun writeTo(sink: BufferedSink) {
+        val gzipBufferedSink: BufferedSink = GzipSink(sink).buffer()
+        body.writeTo(gzipBufferedSink)
+        gzipBufferedSink.close()
+      }
+    }
+}
--- a/engine/src/main/java/com/google/android/fhir/sync/remote/RetrofitHttpService.kt
+++ b/engine/src/main/java/com/google/android/fhir/sync/remote/RetrofitHttpService.kt
@@ -74,6 +74,9 @@ internal interface RetrofitHttpService : FhirHttpService {
                 }
               )
             }
+            if (networkConfiguration.uploadWithGzip) {
+              addInterceptor(GzipUploadInterceptor())
+            }
           }
           .build()
       return Retrofit.Builder()
--- a/engine/src/test/java/com/google/android/fhir/sync/remote/GzipUploadInterceptorTest.kt
+++ b/engine/src/test/java/com/google/android/fhir/sync/remote/GzipUploadInterceptorTest.kt
@@ -1,0 +1,72 @@
+/*
+ * Copyright 2022 Google LLC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *       http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.google.android.fhir.sync.remote
+
+import com.google.android.fhir.MediaTypes
+import com.google.common.truth.Truth.assertThat
+import java.util.zip.GZIPInputStream
+import okhttp3.OkHttpClient
+import okhttp3.Request
+import okhttp3.RequestBody.Companion.toRequestBody
+import okhttp3.mockwebserver.MockResponse
+import okhttp3.mockwebserver.MockWebServer
+import org.junit.Rule
+import org.junit.Test
+import org.junit.runner.RunWith
+import org.robolectric.RobolectricTestRunner
+
+@RunWith(RobolectricTestRunner::class)
+class GzipUploadInterceptorTest {
+
+  @JvmField @Rule val server = MockWebServer()
+  private val client = OkHttpClient.Builder().addInterceptor(GzipUploadInterceptor()).build()
+
+  @Test
+  fun `uncompressed request body received by server should match request body sent`() {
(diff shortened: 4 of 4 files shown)


engine/src/main/java/com/google/android/fhir/sync/remote/GzipUploadInterceptor.kt
```diff
@@ -0,0 +1,59 @@
...
+  private fun gzip(body: RequestBody): RequestBody =
+    object : RequestBody() {
+      override fun contentType(): MediaType? = body.contentType()
+
+      override fun writeTo(sink: BufferedSink) {
+        val gzipBufferedSink: BufferedSink = GzipSink(sink).buffer()
+        body.writeTo(gzipBufferedSink)
+        gzipBufferedSink.close()
+      }
+    }
```
aditya-07, line 58: Should we use this [RequestBody.gzip](https://square.github.io/okhttp/4.x/okhttp/okhttp3/-request-body/-companion/gzip/)() ?

aditya-07 (commented):
Nice PR @omarismail94 . Left a few comments.

engine/src/main/java/com/google/android/fhir/sync/remote/RetrofitHttpService.kt
```diff
@@ -74,6 +74,9 @@ internal interface RetrofitHttpService : FhirHttpService {
                 }
               )
             }
+            if (networkConfiguration.uploadWithGzip) {
+              addInterceptor(GzipUploadInterceptor())
+            }
```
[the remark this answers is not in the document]
omarismail94: great comment! I moved the addition of the GZIP interceptor before the Http logger. After I did so, the HTTP logger showed this:
```
--> POST http://localhost:8095/fhir/
Content-Type: application/fhir+json; charset=utf-8
Content-Encoding: gzip
--> END POST (binary -1-byte body omitted)
```

to get the size to be added, I needed to add the `addContentLength` function to the Gzip interceptor. After doing so, I get this:
```
--> POST http://localhost:8095/fhir/
Content-Type: application/fhir+json; charset=utf-8
Content-Length: 270
Content-Encoding: gzip
--> END POST (binary 270-byte body omitted)
```

if we decide to revert, and have the interceptor after the logger, the output of the logger would be:
```
--> POST http://localhost:8095/fhir/
Content-Type: application/fhir+json; charset=utf-8
Content-Length: 454
{"resourceType":"Bundle","type":"transaction","entry":[{"fullUrl":"Patient/e52aa76c-bbf0-4fd1-9ded-1fa9c7e9573e","resource":{"resourceType":"Patient","id":"e52aa76c-bbf0-4fd1-9ded-1fa9c7e9573e","active":true,"name":[{"family":"K","given":["J"]}],"telecom":[{"system":"phone","value":"1"}],"gender":"female","birthDate":"2023-05-10","address":[{"city":"2","country":"3"}]},"request":{"method":"PUT","url":"Patient/e52aa76c-bbf0-4fd1-9ded-1fa9c7e9573e"}}]}
--> END POST (454-byte body)
```

However, this is wrong as that is not the request being sent!

engine/src/main/java/com/google/android/fhir/FhirEngineProvider.kt
```diff
@@ -161,5 +161,7 @@ data class NetworkConfiguration(
   /** Read timeout (in seconds) for network connection. The default is 10 seconds. */
   val readTimeOut: Long = 10,
   /** Write timeout (in seconds) for network connection. The default is 10 seconds. */
-  val writeTimeOut: Long = 10
+  val writeTimeOut: Long = 10,
+  /** Compresses requests when uploading to a server that supports gzip. */
+  val uploadWithGzip: Boolean = false,
```
jingtang10, line 166: do you think it's better if this is true?

engine/src/main/java/com/google/android/fhir/sync/remote/RetrofitHttpService.kt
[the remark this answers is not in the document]
jingtang10: @joiskash @nsabale7 can you comment on how you use the http logger intercepter's output? do you need to inspect the uncompressed data, or do you care about the packet size more?

engine/src/main/java/com/google/android/fhir/sync/remote/GzipUploadInterceptor.kt
```diff
@@ -0,0 +1,74 @@
...
+import okhttp3.RequestBody
+import okhttp3.Response
+import okio.Buffer
+import okio.BufferedSink
+import okio.GzipSink
+import okio.buffer
+
+const val CONTENT_ENCODING_HEADER_NAME = "Content-Encoding"
+
+/** Compresses upload requests with gzip */
```
jingtang10, line 30:
```suggestion
/** Compresses upload requests with gzip. */
```

engine/src/main/java/com/google/android/fhir/sync/remote/GzipUploadInterceptor.kt
```diff
@@ -0,0 +1,74 @@
...
+object GzipUploadInterceptor : Interceptor {
+  override fun intercept(chain: Interceptor.Chain): Response {
+    val uncompressedRequest = chain.request()
+    if (uncompressedRequest.body == null ||
+        uncompressedRequest.header(CONTENT_ENCODING_HEADER_NAME) != null
```
jingtang10, line 35: should you be appending `, gzip` to the header in this case?

https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Encoding#syntax

engine/src/main/java/com/google/android/fhir/sync/remote/GzipUploadInterceptor.kt
```diff
@@ -0,0 +1,74 @@
...
+      override fun contentType(): MediaType? = body.contentType()
+
+      override fun writeTo(sink: BufferedSink) {
+        val gzipBufferedSink: BufferedSink = GzipSink(sink).buffer()
+        body.writeTo(gzipBufferedSink)
+        gzipBufferedSink.close()
+      }
+    }
+
+  private fun addContentLength(requestBody: RequestBody): RequestBody {
```
jingtang10, line 61: can you test this with and without the "re-sinking"... and profile it?

jingtang10 (changes requested):
Great PR thanks omar!

if you can, please also test on gcp, and if possible, azure, ibm, firely. and knowing the default behavious of all these servers will help us make an informed decision on the default value of the new network config.

engine/src/main/java/com/google/android/fhir/sync/remote/GzipUploadInterceptor.kt
joiskash: We only care about the packet size and HEADER.

omarismail94: @joiskash how do you use the packet size? Depending on if we put the Gzip interceptor before or after the HTTP logging interceptor, it might cause some changes. If we put the GZIP interceptor before, the packet size value you will see in the HTTP logger will be the compressed size of the body. If the GZIP interceptor is put after the HTTP logger, the packet size you will see in the HTTP logger will be of the uncompressed body, before it gets compressed and sent to your server. Does the order of the interceptors matter?

engine/src/main/java/com/google/android/fhir/sync/remote/RetrofitHttpService.kt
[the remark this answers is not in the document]
joiskash: We just use it mainly to get an understanding of how much internet data is being used.

engine/src/main/java/com/google/android/fhir/FhirEngineProvider.kt
omarismail94: I tested uploading a compressed request to a GCP, Azure, and Firely, and only the GCP-managed FHIR store worked. Requests to Azure and Firely throw an error that the request received is invalid JSON.

engine/src/main/java/com/google/android/fhir/sync/remote/GzipUploadInterceptor.kt
omarismail94: I added a benchmark class to this code, and ran it twice. Once with the addContentLength function, and once without. There doesn't seem to be a correlation between adding the extra step of another buffer, and the size of the bundle uploaded:

```
withAddContentLength
203,503,581   ns    trace    I_GzipUploadInterceptorBenchmark.upload_10patientsWithGzip
36,540,975   ns    trace    I_GzipUploadInterceptorBenchmark.upload_100patientsWithGzip
217,292,400   ns    trace    I_GzipUploadInterceptorBenchmark.upload_1000patientsWithGzip

withoutAddContentLength
191,387,330   ns    trace    I_GzipUploadInterceptorBenchmark.upload_10patientsWithGzip
45,564,209   ns    trace    I_GzipUploadInterceptorBenchmark.upload_100patientsWithGzip
243,149,251   ns    trace    I_GzipUploadInterceptorBenchmark.upload_1000patientsWithGzip
```

omarismail94: @jingtang10 I re-ran the test to compare the time differences. The results from last time were only from one run, but the ones below are the average from 50 times. The benchmark config was setup to only run once, and not 50. I've asked Anchita to remove the setting limiting the runs to one in her PR.

Anyway, the results now make sense, adding content length does increase the time taken. There is ~3% increase.  

```
withAddContentLength
51,566,691   ns       11749 allocs    trace    I_GzipUploadInterceptorBenchmark.upload_10patientsWithGzip
45,305,826   ns      107252 allocs    trace    I_GzipUploadInterceptorBenchmark.upload_100patientsWithGzip
130,954,468   ns     1062344 allocs    trace    I_GzipUploadInterceptorBenchmark.upload_1000patientsWithGzip
```

```
withoutAddContentLength
51,104,940   ns       11743 allocs    trace    I_GzipUploadInterceptorBenchmark.upload_10patientsWithGzip
44,362,508   ns      107251 allocs    trace    I_GzipUploadInterceptorBenchmark.upload_100patientsWithGzip
126,258,545   ns     1062365 allocs    trace    I_GzipUploadInterceptorBenchmark.upload_1000patientsWithGzip
```

I also looked at the native implementation in the Okhttp package which Aditya suggested, where they add the logger for gzip themselves, but  looking at the [blame](https://github.com/square/okhttp/blame/master/okhttp-logging-interceptor/src/main/kotlin/okhttp3/logging/HttpLoggingInterceptor.kt), I don't think the change to log the gzip request body size is part of 4.x stable release

omarismail94:
@jingtang10 resolved all comments

jingtang10 (approved):

omarismail94:
I re-ran the benchmark to see how adding content-length influences the time taken to upload and plotted the results below. Yes, it does increase at a faster rate as more data is compressed, but I think for now, this is OK. Even if we do upgrade to the OkHttp library that adds the content length for us, it does so in the same way we do, so we would face the same issue. The alternative is to have the gzip interceptor after the HttpLoggingInterceptor, however, that would mean the LoggingInterceptor would be printing the content-length of the uncompressed request

with addContentLength and without addContentLength

PR #2022, opened 2023-06-01
Upload work manager

### LINKED ISSUE #2021: Interface to control over what should be uploaded and how?
As of now, developers have a lot of control over what to download and how to persist it, but no control over what to upload and how to upload.

Similar to the DownloadWorkManager for downloads, have an interface for UploadWorkManager for uploads.

NA

**Additional context**
NA

Yes

anchita-g:

Fixes #2021 

**Description**
- Create an interface for UploadWorkManager which will be able to orchestrate over what to upload from the list of local changes and how to upload (create Bundle or individual requests from the list of changes to be uploaded)
- Refactored the existing BundleUploader to be used as a default way to create Bundles from LocalChanges 

**Alternative(s) considered**
NA

**Type**
 Feature

upload-worker

**Checklist**


anchita-g:
To contain the size of the changes, I have just added major refactoring and interface changes in this PR. I will follow this one with another PR to support other upload requests (per resource based uploads).

engine/src/main/java/com/google/android/fhir/sync/UploadWorkManager.kt
```diff
@@ -0,0 +1,42 @@
...
+ * Manager that pre-processes the local FHIR changes and handles how to upload them to the server.
+ */
+abstract class UploadWorkManager(private val uploadRequestGenerator: UploadRequestGenerator) {
+
+  /**
+   * Transform the [localChanges] to the final set of changes that needs to be uploaded to the
+   * server. The default implementation is to squash all the changes by resource type so that there
+   * is at most one local change to be uploaded per resource
+   */
+  open fun preprocessLocalChanges(localChanges: List<LocalChange>): List<LocalChange> {
```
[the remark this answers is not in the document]
omarismail94: actually, on looking at the code again, can we create an interface, and combine both these functions into one? Each implementation of `createUploadRequestsFromChanges` can decide what it wants to do with the changes (squash or not)

engine/src/main/java/com/google/android/fhir/sync/UploadWorkManager.kt
```diff
@@ -0,0 +1,42 @@
...
+    return localChanges
+      .groupBy { it.resourceId to it.resourceType }
+      .values.map { localResourceChanges -> LocalChangeUtils.squash(localResourceChanges) }
+  }
+
+  /** Generates a list of [UploadRequest] from the [LocalChange] to be sent to the server */
+  fun createUploadRequestsFromChanges(localChanges: List<LocalChange>): List<UploadRequest> {
```
omarismail94, line 39: nit:
```suggestion
  fun createUploadRequestsFromLocalChanges(localChanges: List<LocalChange>): List<UploadRequest> {
```

engine/src/main/java/com/google/android/fhir/sync/upload/UploaderImpl.kt
```diff
@@ -32,26 +32,36 @@ import org.hl7.fhir.r4.model.Resource
...
+ * Implementation of the [Uploader]. It orchestrates the pre processing of [LocalChange] and
+ * constructing appropriate upload requests via [UploadWorkManager] and uploading of requests via
+ * [DataSource]. [Uploader] clients should call upload and listen to the various states emitted by
+ * [UploadWorkManager] as [UploadResult].
+ */
+internal class UploaderImpl(
   private val dataSource: DataSource,
-  private val bundleGenerator: TransactionBundleGenerator,
-  private val localChangesPaginator: LocalChangesPaginator
+  private val uploadWorkManager: UploadWorkManager
```
omarismail94, line 43: nit: add trailing comma at the end so when we go to add more parameters (hopefully never), we don't have to edit this line
```suggestion
  private val uploadWorkManager: UploadWorkManager,
```

engine/src/main/java/com/google/android/fhir/sync/UploadWorkManager.kt
```diff
@@ -0,0 +1,42 @@
...
+
+package com.google.android.fhir.sync
+
+import com.google.android.fhir.LocalChange
+import com.google.android.fhir.db.impl.dao.LocalChangeUtils
+
+/**
+ * Manager that pre-processes the local FHIR changes and handles how to upload them to the server.
+ */
+abstract class UploadWorkManager(private val uploadRequestGenerator: UploadRequestGenerator) {
```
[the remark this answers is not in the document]
anchita-g: The abstract class was created to provide an out-of-the-box implementation for `preprocessing` changes - to squash changes. This was a default implementation till now, hence, was of the opinion to provide this as is.

engine/src/main/java/com/google/android/fhir/sync/Uploader.kt
```diff
@@ -20,23 +20,38 @@ import com.google.android.fhir.LocalChange
...
    * Uploads the local changes to the [DataSource]. Particular implementations should take care of
-   * transforming the [SquashedLocalChange]s to particular network operations. If [ProgressCallback]
-   * is provided it also reports the intermediate progress
+   * transforming the [LocalChange]s to particular network operations. If [ProgressCallback] is
+   * provided it also reports the intermediate progress
    */
   suspend fun upload(localChanges: List<LocalChange>): Flow<UploadResult>
 }
 
+sealed class UploadRequest(
```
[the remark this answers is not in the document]
anchita-g: IMO we should keep Download requests different from Upload requests as the Upload requests could be of different types like `PATCH`, `PUT`, `DELETE` (as my follow up PR would do). I agree with keeping the Request out of DWM. We should also consider renaming `Request` here with `DownloadRequest`

engine/src/main/java/com/google/android/fhir/sync/remote/FhirHttpDataSource.kt
```diff
@@ -33,6 +36,6 @@ internal class FhirHttpDataSource(private val fhirHttpService: FhirHttpService)
       is BundleRequest -> fhirHttpService.post(request.bundle, request.headers)
```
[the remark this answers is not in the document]
anchita-g: Makes sense. Will do this!

engine/src/main/java/com/google/android/fhir/sync/UploadWorkManager.kt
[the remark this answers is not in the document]
anchita-g: We could do this. I was thinking about doing something similar. However, felt like it would be really helpful to provide squashing of changes as a default implementation of pushing changes. If that is to happen, it would need the changes transformation to be different from how to upload them.

engine/src/main/java/com/google/android/fhir/sync/remote/FhirHttpDataSource.kt
```diff
@@ -33,6 +36,6 @@ internal class FhirHttpDataSource(private val fhirHttpService: FhirHttpService)
...
     }
 
-  override suspend fun upload(request: BundleRequest) =
-    fhirHttpService.post(request.bundle, request.headers)
+  override suspend fun upload(request: UploadRequest): Resource =
```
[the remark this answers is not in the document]
anchita-g: If we keep the Download Requests separately from the Upload Requests, we might not need this.

engine/src/main/java/com/google/android/fhir/sync/UploadWorkManager.kt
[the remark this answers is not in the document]
omarismail94: I think we can still do that if we switch this to an interface. See `ResourceParamsBasedDownloadWorkManager` as an example of an implementation of `DownloadWorkManager` that gives an out-of-the box implementation of the whole class. In general we should favour composition over inheritance

engine/src/main/java/com/google/android/fhir/sync/UploadWorkManager.kt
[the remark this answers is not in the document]
omarismail94: We can still do this with a default implementation of the interface that is public, if we go the route of converting this class to an interface

engine/src/main/java/com/google/android/fhir/sync/Request.kt
```diff
@@ -0,0 +1,106 @@
...
+ */
+data class BundleDownloadRequest
+internal constructor(val bundle: Bundle, override val headers: Map<String, String> = emptyMap()) :
+  DownloadRequest(headers)
+
+/**
+ * Structure represents a request that can be made to upload resources/resource modifications to the
+ * FHIR server.
+ */
+sealed class UploadRequest(
```
omarismail94, line 92: could you make the creation of this class similar to that of `DownloadRequest`, i.e. having a companion object with `of` functions and having UploadRquest just take in headers in its constructor

engine/src/main/java/com/google/android/fhir/sync/UploadWorkManager.kt
```diff
@@ -17,26 +17,18 @@
...
-      .groupBy { it.resourceId to it.resourceType }
-      .values.map { localResourceChanges -> LocalChangeUtils.squash(localResourceChanges) }
-  }
+  fun preprocessLocalChanges(localChanges: List<LocalChange>): List<LocalChange>
 
-  /** Generates a list of [UploadRequest] from the [LocalChange] to be sent to the server */
-  fun createUploadRequestsFromChanges(localChanges: List<LocalChange>): List<UploadRequest> {
-    return uploadRequestGenerator.generateUploadRequests(localChanges)
-  }
+  /** Generates a list of [UploadRequest] from the [LocalChange]s to be uploaded to the server */
```
omarismail94, line 32: nit
```suggestion
  /** Generates a list of [UploadRequest]s from the [LocalChange]s to be uploaded to the server */
```

engine/src/main/java/com/google/android/fhir/sync/UploadWorkManager.kt
```diff
@@ -17,26 +17,18 @@
...
-import com.google.android.fhir.db.impl.dao.LocalChangeUtils
 
 /**
  * Manager that pre-processes the local FHIR changes and handles how to upload them to the server.
  */
-abstract class UploadWorkManager(private val uploadRequestGenerator: UploadRequestGenerator) {
+interface UploadWorkManager {
 
   /**
    * Transform the [localChanges] to the final set of changes that needs to be uploaded to the
```
omarismail94, line 27: nit"
```suggestion
   * Transform the [LocalChange]s to the final set of changes that needs to be uploaded to the
```

engine/src/main/java/com/google/android/fhir/sync/remote/FhirHttpDataSource.kt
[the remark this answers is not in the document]
omarismail94: what are the advantage of keeping Download Request separate from Upload requests that make it worth it to have two functions doing almost the same thing (sending post/get request)

engine/src/main/java/com/google/android/fhir/sync/UploadWorkManager.kt
```diff
@@ -17,26 +17,18 @@
...
-   * server. The default implementation is to squash all the changes by resource type so that there
-   * is at most one local change to be uploaded per resource
+   * server.
```
omarismail94, line 28: if you want to maintain having a generator as a property, you can have it in the interface:

https://kotlinlang.org/docs/interfaces.html#properties-in-interfaces

 i don't know how useful it is. What are your thoughts?

engine/src/main/java/com/google/android/fhir/testing/Utilities.kt
```diff
@@ -115,7 +113,8 @@ open class TestDownloadManagerImpl(
 ) : DownloadWorkManager {
   private val urls = LinkedList(queries)
 
-  override suspend fun getNextRequest(): Request? = urls.poll()?.let { Request.of(it) }
+  override suspend fun getNextRequest(): DownloadRequest? =
+    urls.poll()?.let { DownloadRequest.of(it) }
```
omarismail94, line 117: nit: add space
```suggestion
    urls.poll()?.let { DownloadRequest.of(it) }
    
```

engine/src/main/java/com/google/android/fhir/sync/UploadWorkManager.kt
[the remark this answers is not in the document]
omarismail94: unresolving this as the function name is still the same

engine/src/main/java/com/google/android/fhir/sync/remote/FhirHttpDataSource.kt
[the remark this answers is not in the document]
anchita-g: The nuances of what needs to go in a Upload/Download request is different. For upload, we need to manage local change tokens, resource type information of the resource to be uploaded etc. The semantics of constructing a request for Download is different from Upload. As for post/get and other types of requests, we do have `FhirHttpService` where in we do not differentiate on the type of request (download/upload). These are simple HTTP method calls.
`DataSource` will help us map the correct Upload/Download request to the corresponding HTTP method call in `FhirHttpService`.  I like the way where we have clear way to `download` using the `DownloadRequest` as it clearly specifies the type of requests that are supported for downloading. Similarly for uploading

engine/src/main/java/com/google/android/fhir/sync/Request.kt
anchita-g: I am going to follow PR this with another PR, where in, I would add more types of `UploadRequest`. The `of` way of doing things might not be that intuitive for the developer. Can we keep this discussion for the follow-up PR?

engine/src/main/java/com/google/android/fhir/sync/UploadWorkManager.kt
anchita-g: TIL!!! Very helpful. Didn't know something like this exists. (Part of the reason why I wanted to go the abstract class way)

anchita-g: I am not sure if this will be useful to provide it as a property as we can not use it in the interface itself. But thanks for this, TIL!!

engine/src/main/java/com/google/android/fhir/testing/Utilities.kt
anchita-g: Oh you meant a new line!! Done

jingtang10 (changes requested):

engine/src/main/java/com/google/android/fhir/sync/Uploader.kt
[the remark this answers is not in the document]
jingtang10: would upload and download requests also need to handle other differences such as page size? i'm not in favour of unifying prematurely.

engine/src/main/java/com/google/android/fhir/sync/upload/LocalChangesPaginator.kt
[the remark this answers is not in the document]
anchita-g: I agree, removed the interface.

engine/src/main/java/com/google/android/fhir/sync/upload/HttpVerbBasedBundleEntryComponentGenerator.kt
```diff
@@ -25,13 +25,13 @@ import org.hl7.fhir.r4.model.Resource
...
- * [Resource] for the [LocalChangeEntity]. See [https://www.hl7.org/fhir/http.html#transaction] for
- * more info regarding the supported [Bundle.HTTPVerb].
+ * Abstract class for generating [Bundle.BundleEntryComponent] for a [LocalChange] to be added to
+ * the [Bundle] based on [Bundle.HTTPVerb] supported by the Fhir server. Concrete implementations of
+ * the class should provide implementation of [getEntryResource] to provide [Resource] for the
+ * [LocalChangeEntity]. See [https://www.hl7.org/fhir/http.html#transaction] for more info regarding
+ * the supported [Bundle.HTTPVerb].
  */
-internal abstract class HttpVerbBasedBundleEntryComponentGenerator(
+abstract class HttpVerbBasedBundleEntryComponentGenerator(
```
jingtang10, line 34:
```suggestion
abstract class BundleEntryComponentGenerator(
```

engine/src/main/java/com/google/android/fhir/sync/UploadWorkManager.kt
```diff
@@ -0,0 +1,33 @@
...
+package com.google.android.fhir.sync
+
+import com.google.android.fhir.LocalChange
+
+/**
+ * Manager that pre-processes the local FHIR changes and handles how to upload them to the server.
+ */
+interface UploadWorkManager {
+  /**
+   * Transform the [LocalChange]s to the final set of changes that needs to be uploaded to the
```
jingtang10, line 26:
```suggestion
   * Transforms the [LocalChange]s to the final set of changes that needs to be uploaded to the
```

jingtang10 (approved):

engine/src/main/java/com/google/android/fhir/sync/UploadWorkManager.kt
```diff
@@ -0,0 +1,42 @@
...
+  /**
+   * Transforms the [LocalChange]s to the final set of changes that needs to be uploaded to the
+   * server. The transformations can be of various types like squashing the [LocalChange]s by
+   * [Resource] e.g. [SquashedChangesUploadWorkManager] or filtering out certain [LocalChange]s or
+   * grouping the changes.
+   */
+  fun prepareChangesForUpload(localChanges: List<LocalChange>): List<LocalChange>
+
+  /** Generates a list of [UploadRequest]s from the [LocalChange]s to be uploaded to the server */
+  fun createUploadRequestsFromLocalChanges(localChanges: List<LocalChange>): List<UploadRequest>
```
MJ1998, line 34: As I see we are using these 2 APIs one after the other in UploaderImpl. Could we not combine these 2 APIs into 1 ?
If it's needed for probable future work then do you mind kdocing the use-case ?

engine/src/main/java/com/google/android/fhir/sync/UploadWorkManager.kt
```diff
@@ -0,0 +1,42 @@
...
+
+  /**
+   * Gets the [Int] to indicate the progress in terms of the pending uploads. The indicator could be
+   * determined at the resource level (by extracting resource information from the upload requests)
+   * etc.
+   */
+  fun getPendingUploadsIndicator(uploadRequests: List<UploadRequest>): Int
```
jingtang10, line 41: Discussed this with @anchita-g. One alternative way of handling this is to make the `UploadWorkManager` a little more stateful, similar to the `DownloadWorkManager`. In that case, there would be an API to feed local changes to the `UploadWorkManager`, there will be an API to pull next upload request to be sent to the FHIR server, and there'll be an API to get the sync status - almost like an iterator with a sync progress.

That would make the job of `Uploader` a little simpler - the while loop currently in `UplaoderImpl` would be simplified (because the work is done by the `UploadWorkManager`. This affords the developers a little more control - but admittedly we're not entirely sure if the additional contronl is urgently needed by the developers.

But that would make the design more alinged with the download workflow which @omarismail94 worked on.

FYI @aditya-07 as well.

jingtang10 (approved):
Thanks @anchita-g - happy for this to be merged but we might have some follow-ups to do.

engine/src/main/java/com/google/android/fhir/sync/UploadWorkManager.kt
anchita-g: Created an issue https://github.com/google/android-fhir/issues/2103 for this

engine/src/main/java/com/google/android/fhir/sync/UploadWorkManager.kt
anchita-g: These APIs will go for refactoring, will write up a proper wiki thereafter

PR #2032, opened 2023-06-12
Implement answerOption initialSelected to set initial value for choice

### LINKED ISSUE #2035: Implement answerOption initialSelected to set initial value for choice
**Describe the bug**
Fhir spec allows to add a boolean property 'initialSelected' to each answerOption to set that as initial when rendering questionnaire item as defined [here](http://hl7.org/fhir/R5/questionnaire-definitions.html#Questionnaire.item.answerOption.initialSelected) . 
Also [Rule 11](https://hl7.org/fhir/R5/questionnaire-definitions.html#Questionnaire.item.initial) specifies that "If one or more answerOption is present, initial cannot be present. Use answerOption.initialSelected instead"

**Component**
SDC library

**To Reproduce**
This questionnaire should work and allow setting initial values
```
{
  "item": [{
      "linkId": "2.0",
      "type": "choice",
      "text": "Initially selected multiple choice",
      "repeats": true,
      "answerOption": [{
          "valueCoding": {
            "code": "1st",
            "display": "First",
            "system": "custom"
          }
        }, {
          "valueCoding": {
            "code": "2nd",
            "display": "Second",
            "system": "custom"
          },
          "initialSelected": true
        }, {
          "valueCoding": {
            "code": "3rd",
            "display": "Third",
            "system": "custom"
          },
          "initialSelected": true
       } ]
    }, {
      "linkId": "2.0",
      "type": "string",
      "text": "Initially provided text value",
      "initial": [{
          "valueString": "Here is a sample text"
      }]
    }
]}
```

**Expected behavior**
Above questionnaire should work and select the values marked as initialSelected in answerOption.

A PR is created and this issue would be resolved by that

maimoonak:
Fixes #2035

**Description**
Fhir spec allows to add boolean property to set initial selection of answerOptions as defined [here](http://hl7.org/fhir/R5/questionnaire-definitions.html#Questionnaire.item.answerOption.initialSelected) . Also [Rule 11](https://hl7.org/fhir/R5/questionnaire-definitions.html#Questionnaire.item.initial) specifies that "If one or more answerOption is present, initial cannot be present. Use answerOption.initialSelected instead"

**Alternative(s) considered**
Rule specifies to use only one of the property for initial answers. When we are creating responseItem we are setting initial value hence updating answer there.

**Type**
Choose one: (Bug fix | Feature)

**Screenshots**

**Checklist**


datacapture/src/main/java/com/google/android/fhir/datacapture/extensions/MoreQuestionnaireItemComponents.kt
```diff
@@ -528,11 +528,16 @@ val Questionnaire.QuestionnaireItemComponent.enableWhenExpression: Expression?
...
+  }
+
   // https://build.fhir.org/ig/HL7/sdc/behavior.html#initial
   // quantity given as initial without value is for unit reference purpose only. Answer conversion
   // not needed
-  if (initial.isEmpty() ||
-      (initialFirstRep.hasValueQuantity() && initialFirstRep.valueQuantity.value == null)
+  if (answerOption.none { it.initialSelected } &&
+      (initial.isEmpty() ||
+        (initialFirstRep.hasValueQuantity() && initialFirstRep.valueQuantity.value == null))
```
jingtang10, line 540: i'm not actually so sure if this block is needed any more. can you test what happens if we have a quantity without value and how the widgets will behave?

catalog/src/main/java/com/google/android/fhir/catalog/BehaviorListViewModel.kt
```diff
@@ -52,6 +52,11 @@ class BehaviorListViewModel(application: Application) : AndroidViewModel(applica
       R.drawable.ic_dynamic_text_behavior,
       R.string.behavior_name_dynamic_question_text,
       "behavior_dynamic_question_text.json"
+    ),
+    INITIAL_VALUE(
```
jingtang10, line 56: i actually think this should be in widgets, but added as part of the misc components. can you move this there?

datacapture/src/main/java/com/google/android/fhir/datacapture/extensions/MoreQuestionnaireItemComponents.kt
```diff
@@ -545,16 +550,16 @@ private fun Questionnaire.QuestionnaireItemComponent.createQuestionnaireResponse
...
       "Questionnaire item $linkId can only have multiple initial values for repeating items. See rule que-13 at https://www.hl7.org/fhir/questionnaire-definitions.html#Questionnaire.item.initial."
     )
   }
 
   return initial
-    .map {
-      QuestionnaireResponse.QuestionnaireResponseItemAnswerComponent().apply { value = it.value }
-    }
+    .map { it.value }
+    .ifEmpty { answerOption.filter { it.initialSelected }.map { it.value } }
```
jingtang10, line 561: can you move up the `ifEmpty` so that you can just call the `it.value` once?

jingtang10 (changes requested):
Thanks @maimoonak! Great job! 🙏🙏🙏

datacapture/src/main/java/com/google/android/fhir/datacapture/extensions/MoreQuestionnaireItemComponents.kt
```diff
@@ -528,11 +528,16 @@ val Questionnaire.QuestionnaireItemComponent.enableWhenExpression: Expression?
  */
 private fun Questionnaire.QuestionnaireItemComponent.createQuestionnaireResponseItemAnswers():
   MutableList<QuestionnaireResponse.QuestionnaireResponseItemAnswerComponent>? {
+  require(answerOption.filter { it.initialSelected }.isEmpty() || initial.isEmpty()) {
+    "Questionnaire item $linkId has both initial value(s) and has answerOption. See rule que-11 at https://www.hl7.org/fhir/questionnaire-definitions.html#Questionnaire.item.initial."
```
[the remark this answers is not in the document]
maimoonak: That`s true but we may have people already using initial / initial-expression into existing questionnaires with answerOptions as SDK allowed i and use of initialSelected was not available as an alternate. If we can get a consensus from stakeholders on this then we can float this update. Otherwise it can cause breaking changes if users of SDK do not update their existing questionnaires. 
Let me know how to proceed with it

datacapture/src/main/java/com/google/android/fhir/datacapture/extensions/MoreQuestionnaireItemComponents.kt
maimoonak: I have removed the ifEmpty as we are already checking that its not possible to have both in same item. However, the ifEmpty can not be moved  above as initial is InitialAnswerComponent and answerOption is AnswerOptionComponent class.

datacapture/src/main/java/com/google/android/fhir/datacapture/extensions/MoreQuestionnaireItemComponents.kt
maimoonak: `createQuestionResponse should not set answer for quantity type with missing value` tests the behavior. Added the quantity to catalog as well. The output only has a correct quantity if user inputs data into quantity where initial was only to suggest unit

datacapture/src/main/java/com/google/android/fhir/datacapture/extensions/MoreQuestionnaireItemAnswerOptionComponents.kt
```diff
@@ -42,6 +44,10 @@ internal val Questionnaire.QuestionnaireItemAnswerOptionComponent.optionExclusiv
     return false
   }
 
+/** Get the answer options values with `initialSelected` set to true */
+internal val List<QuestionnaireItemAnswerOptionComponent>.initial: List<Type>
```
jingtang10, line 48: maybe add some unit tests for this one?

jingtang10 (approved):
Thanks @maimoonak - i do have some comments but this PR is good to be merged. Please just address the comments (in some cases just acknowledge them and create todo/issue). Thanks again for the great work! 🎉🙏👍

datacapture/src/main/java/com/google/android/fhir/datacapture/extensions/MoreQuestionnaireItemAnswerOptionComponents.kt
jingtang10, line 48: can you rename this to initialSelected?

```suggestion
internal val List<QuestionnaireItemAnswerOptionComponent>.initialSelected: List<Type>
```

datacapture/src/test/java/com/google/android/fhir/datacapture/extensions/MoreQuestionnaireItemAnswerOptionComponentsTest.kt
```diff
@@ -204,4 +204,44 @@ class MoreQuestionnaireItemAnswerOptionComponentsTest {
...
+  }
+
+  private fun answerOptionOf(
+    code: String,
+    url: String,
+    display: String,
+    initialSelected: Boolean?
+  ) =
+    Questionnaire.QuestionnaireItemAnswerOptionComponent().apply {
+      this.value = Coding().setCode(code).setDisplay(display).setSystem(url)
```
jingtang10, line 244:
```suggestion
      value = Coding().setCode(code).setDisplay(display).setSystem(url)
```

datacapture/src/test/java/com/google/android/fhir/datacapture/extensions/MoreQuestionnaireItemAnswerOptionComponentsTest.kt
```diff
@@ -204,4 +204,44 @@ class MoreQuestionnaireItemAnswerOptionComponentsTest {
...
+      listOf(
+        answerOptionOf("test-code 1", "http://code.com", "Test Code 1", null),
+        answerOptionOf("test-code 2", "http://code.com", "Test Code 2", null),
+        answerOptionOf("test-code 3", "http://code.com", "Test Code 3", false),
+        answerOptionOf("test-code 4", "http://code.com", "Test Code 4", false)
+      )
+    val selectedOptions = answerOptions.initialSelected
+
+    assertThat(selectedOptions).isEmpty()
+  }
```
jingtang10, line 235: A comment on writing tests: it's much better to write tests in modular ways, meaning:

```
should not select option with initialSelected as null
should not select option with initialSelected as false
should select option with intialSelected as true
should select multiple options with initialSelected as true 
```

in the first 3 test cases, you can just have one single answer option for each test case. and for the last test case, you can have a few answer options, maybe one with initialSelect as null, one as false, and two as true.

the reason is, if, for example, the logic to handle null case is broken, if we had the two test cases in this file, they would both fail, and it's hard to know what's the problem without debugging. but if we had more modular test cases, then you can see that one test case is failing and others are still passing - so you can very easily figure out what the bug is without even debugging.

datacapture/src/test/java/com/google/android/fhir/datacapture/extensions/MoreQuestionnaireItemComponentsTest.kt
```diff
@@ -1684,6 +1731,70 @@ class MoreQuestionnaireItemComponentsTest {
...
+        initial =
+          listOf(
+            Questionnaire.QuestionnaireItemInitialComponent().apply {
+              value = Coding("http://abc.com", "a", "A")
+            }
+          )
+        answerOption =
+          listOf(
+            Questionnaire.QuestionnaireItemAnswerOptionComponent().apply {
+              initialSelected = true
```
jingtang10, line 1778: remove this line. this is not necessary.

datacapture/src/test/java/com/google/android/fhir/datacapture/extensions/MoreQuestionnaireItemComponentsTest.kt
```diff
@@ -1684,6 +1731,70 @@ class MoreQuestionnaireItemComponentsTest {
...
+              value = Coding("http://abc.com", "a", "A")
+            },
+            Questionnaire.QuestionnaireItemAnswerOptionComponent().apply {
+              initialSelected = true
```
jingtang10, line 1782: remove this line

jingtang10 (approved):

datacapture/src/test/java/com/google/android/fhir/datacapture/extensions/MoreQuestionnaireItemAnswerOptionComponentsTest.kt
```diff
@@ -206,32 +206,44 @@ class MoreQuestionnaireItemAnswerOptionComponentsTest {
...
-      listOf(
-        answerOptionOf("test-code 1", "http://code.com", "Test Code 1", true),
-        answerOptionOf("test-code 2", "http://code.com", "Test Code 2", null),
-        answerOptionOf("test-code 3", "http://code.com", "Test Code 3", false),
-        answerOptionOf("test-code 4", "http://code.com", "Test Code 4", true)
-      )
+      listOf(answerOptionOf("test-code 1", "http://code.com", "Test Code 1", null))
+    val selectedOptions = answerOptions.initialSelected
+
+    assertThat(selectedOptions).isEmpty()
```
jingtang10, line 214:
```suggestion

    assertThat(answerOptions.initialSelected).isEmpty()
```

jingtang10 (approved):

PR #2034, opened 2023-06-13
Add support for Answer Options Toggle Extension

### LINKED ISSUE #1773: Support for Extension: Answer Options Toggle Expression
This extension is useful when A question has a list of possible answers but the list to be actually displayed depends on the previous question.

This extension solves this use case.

**Additional context**
[HL7 answerOptionsToggleExpression Extension Link](https://build.fhir.org/ig/HL7/sdc/StructureDefinition-sdc-questionnaire-answerOptionsToggleExpression.html)

cc: @jingtang10

LZRS:

Fixes #1773 

**Description**

**Alternative(s) considered**

**Type**

**Checklist**


FikriMilano:
@LZRS thanks for the PR :D
Is this ready for review?

LZRS:
> @LZRS thanks for the PR :D
> Is this ready for review?

Yes, it's ready for review

datacapture/src/test/java/com/google/android/fhir/datacapture/views/QuestionnaireViewItemTest.kt
```diff
@@ -797,4 +802,243 @@ class QuestionnaireViewItemTest {
...
+  }
+
+  @Test
+  fun `enabledAnswerOption should return only enabled answer options from evaluateAnswerOptionsToggleExpressions`() {
+    val questionnaireViewItem =
+      QuestionnaireViewItem(
+        Questionnaire.QuestionnaireItemComponent().apply {
+          repeats = true
+          linkId = "a-question"
+          extension =
```
jingtang10, line 849: what about using `addExtension` instead of `extension =`

datacapture/src/main/java/com/google/android/fhir/datacapture/views/QuestionnaireViewItem.kt
```diff
@@ -207,6 +216,31 @@ data class QuestionnaireViewItem(
         }
       }
 
+  /**
+   * Returns allowed answer options that are enabled based on the evaluation of
+   * [Questionnaire.QuestionnaireItemComponent.answerOptionsToggleExpressions] expressions
+   */
+  internal val enabledAnswerOption: List<Questionnaire.QuestionnaireItemAnswerOptionComponent>
+    get() =
+      runBlocking(Dispatchers.IO) {
```
jingtang10, line 225: as per my previous comment, this code should live in `QuestionnaireViewModel` - and you won't need to do this awkward `runBlocking` thing.

datacapture/src/test/java/com/google/android/fhir/datacapture/views/QuestionnaireViewItemTest.kt
jingtang10, line 1: you should add another test case to show that answer options that don't have an answer options toggle expression for them are included in the enabled answer options.

See this page: https://build.fhir.org/ig/HL7/sdc/StructureDefinition-sdc-questionnaire-answerOptionsToggleExpression.html and this quote: `Any options not listed by one of the answerOptionsToggleExpression occurrences are presumed to be ‘enabled’`

datacapture/src/test/java/com/google/android/fhir/datacapture/views/QuestionnaireViewItemTest.kt
```diff
@@ -797,4 +802,243 @@ class QuestionnaireViewItemTest {
...
+          )
+        }
+      )
+
+    val enabledOption = questionnaireViewItem.enabledAnswerOption
+    assertThat(enabledOption.map { it.valueCoding.code }).containsExactly("option1")
+  }
+
+  @Test
+  fun `enabledAnswerOption removes answers no longer in allowed answer options evaluated from evaluateAnswerOptionsToggleExpressions`() {
```
jingtang10, line 929: tihs test case should probably only be in `QuestionnaireViewModelTest`.

jingtang10 (changes requested):
Thanks for this change - really good tests and in general high quality code!

The most important comment is my first comment in `QuestionnaireViewItem.kt`. Please take a look and address. I think as a result this will also improve the general code quality of the SDC library. At the moment we're incorrectly handling some answer option logic in the UI code - this should be changed.

Thanks!

datacapture/src/main/java/com/google/android/fhir/datacapture/QuestionnaireViewModel.kt
```diff
@@ -552,6 +554,44 @@ internal class QuestionnaireViewModel(application: Application, state: SavedStat
...
+      allowed.flatMap {
+        val (_, options) = it
+        options
+      }
+
+    val disallowedOptions =
+      disallowed.flatMap {
+        val (_, options) = it
+        options.filterNot { option -> allowedOptions.any { type -> equals(type, option) } }
+      }
```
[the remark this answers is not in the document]
LZRS: Oh, for this there was this part of the spec 

> More than one repetition of this extension is possible. If an option appears in more than one repetition, then it will be enabled if it appears in ANY of the extensions whose expressions are evaluated as true. I.e. the expressions that pertain to a given option are treated as 'or'.

See: http://build.fhir.org/ig/HL7/sdc/expressions.html#answerOptionToggleExpression

datacapture/src/main/java/com/google/android/fhir/datacapture/QuestionnaireViewModel.kt
```diff
@@ -552,6 +554,44 @@ internal class QuestionnaireViewModel(application: Application, state: SavedStat
...
+  ): List<Questionnaire.QuestionnaireItemAnswerOptionComponent> {
+    val results =
+      item.answerOptionsToggleExpressions
+        .map {
+          val (expression, toggleOptions) = it
+          val evaluationResult =
+            if (expression.isFhirPath)
+              fhirPathEngine.convertToBoolean(
+                fhirPathEngine.evaluate(questionnaireResponse, expression.expression)
+              )
```
FikriMilano, line 569: @LZRS I think you need to also give access to other FHIRPath context e.g. variable extensions, %resource. 

It's important for FHIRPath expression to be able to access other context, so that questionnaire.item can 'communicate' with other items.

You can check the ExpressionEvaluator file, or here's a calculatedExpressions's code for reference:
https://github.com/google/android-fhir/blob/master/datacapture/src/main/java/com/google/android/fhir/datacapture/fhirpath/ExpressionEvaluator.kt#L126-L155

fyi @MJ1998, since you're currently working on something similar

FikriMilano (changes requested):
Thanks for the good work! :D
I left a comment.

datacapture/src/main/java/com/google/android/fhir/datacapture/QuestionnaireViewModel.kt
LZRS: Hmm I think it should be working, I have this test checking for that scenario 
https://github.com/google/android-fhir/blob/01728d0165b2d5aab8febb868f3b6bc3361e278b/datacapture/src/test/java/com/google/android/fhir/datacapture/QuestionnaireViewModelTest.kt#L4213

Could you maybe share a sample of the questionnaire you intend to use?

FikriMilano: @LZRS can you also add a test that uses variable extension? refer to this questionnaire
https://drive.google.com/file/d/1ZCcS60CzE-EdQawUYbCO8dbxFd2XsiUs/view?usp=sharing

FikriMilano: NVM about the %resource, it works as you mentioned.
But the variable extension doesn't works when you use it in the answer option toggle expression

LZRS: @FikriMilano It should now be working for variable extension expression

FikriMilano (approved):
Looks great, thanks @LZRS :D

datacapture/src/main/java/com/google/android/fhir/datacapture/QuestionnaireViewModel.kt
```diff
@@ -552,6 +556,47 @@ internal class QuestionnaireViewModel(application: Application, state: SavedStat
...
+          evaluationResult to toggleOptions
+        }
+        .partition { it.first }
+    val (allowed, disallowed) = results
+    val allowedOptions = allowed.flatMap { it.second }
+
+    val disallowedOptions =
+      disallowed.flatMap {
+        it.second.filterNot { option -> allowedOptions.any { type -> equals(type, option) } }
```
jingtang10, line 592: can you explain what's going on here?

datacapture/src/main/java/com/google/android/fhir/datacapture/QuestionnaireViewModel.kt
```diff
@@ -552,6 +556,47 @@ internal class QuestionnaireViewModel(application: Application, state: SavedStat
...
+      }
+
+    return answerOptions.filterNot { answerOption ->
```
jingtang10, line 595: according to: https://build.fhir.org/ig/HL7/sdc/StructureDefinition-sdc-questionnaire-answerOptionsToggleExpression.html

"Any options not listed by one of the answerOptionsToggleExpression occurrences are presumed to be ‘enabled’"

doesn't this mean that we can simply remove the disabled options from all the options and our job is done? in other words, we don't really need to take into account the allowed options. is this right?

datacapture/src/main/java/com/google/android/fhir/datacapture/QuestionnaireViewModel.kt
jingtang10: does the spec say anything about if allowed of disallowed options taking precedence?

if not, i'm happy for us to do the simplest thing here.

jingtang10 (changes requested):

datacapture/src/main/java/com/google/android/fhir/datacapture/QuestionnaireViewModel.kt
```diff
@@ -552,6 +556,47 @@ internal class QuestionnaireViewModel(application: Application, state: SavedStat
     return options
   }
 
+  private fun evaluateAnswerOptionsToggleExpressions(
```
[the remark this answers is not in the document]
LZRS: I've moved most of the functions related to evaluation of answer options to a separate class `EnabledAnswerOptionsEvaluator`

datacapture/src/main/java/com/google/android/fhir/datacapture/QuestionnaireViewModel.kt
LZRS: Yeah, on this [page](https://build.fhir.org/ig/HL7/sdc/expressions.html#answerOptionToggleExpression) it states that 

> More than one repetition of this extension is possible. If an option appears in more than one repetition, then it will be enabled if it appears in ANY of the extensions whose expressions are evaluated as true. I.e. the expressions that pertain to a given option are treated as 'or'.

datacapture/src/main/java/com/google/android/fhir/datacapture/QuestionnaireViewModel.kt
LZRS: @jingtang10 Yes, although we may have to take into account disabled options that are also included in enabled options. According to https://build.fhir.org/ig/HL7/sdc/expressions.html#answerOptionToggleExpression
>If an option appears in more than one repetition, then it will be enabled if it appears in ANY of the extensions whose expressions are evaluated as true. I.e. the expressions that pertain to a given option are treated as 'or'.

datacapture/src/main/java/com/google/android/fhir/datacapture/enablement/EnabledAnswerOptionsEvaluator.kt
jingtang10, line 1: enablement package is for enableWhen evaluation (maybe it should just be renamed to enablewhen but let's do that another time) - let's keep this class out of it.

i would advocate for creating a new package called expressions.

i really like what you've done with this class though - nicely packaged class with a single public function.

datacapture/src/main/java/com/google/android/fhir/datacapture/QuestionnaireViewModel.kt
```diff
@@ -490,7 +478,7 @@ internal class QuestionnaireViewModel(application: Application, state: SavedStat
         viewModelScope,
         SharingStarted.Lazily,
         initialValue =
-          runBlocking { getQuestionnaireState() }
+          QuestionnaireState(items = emptyList(), displayMode = DisplayMode.InitMode)
             .also { detectExpressionCyclicDependency(questionnaire.item) }
```
jingtang10, line 482: can you remove the two `.also` calls below if you're only creating an empty questionnaire state to begin with?

datacapture/src/main/java/com/google/android/fhir/datacapture/QuestionnaireViewModel.kt
```diff
@@ -1054,6 +865,8 @@ internal data class QuestionnaireState(
 internal sealed class DisplayMode {
   class EditMode(val pagination: QuestionnairePagination) : DisplayMode()
   data class ReviewMode(val showEditButton: Boolean, val showSubmitButton: Boolean) : DisplayMode()
+
+  object InitMode : DisplayMode()
```
jingtang10, line 869: i'm nervous about this, but also curious why this works. have you tested sufficiently this doesn't break questionnaire initialization?

can you add documentation to say why this is needed?

does this mean we always did more work than necessary when initializing the questionnaire?

datacapture/src/main/java/com/google/android/fhir/datacapture/QuestionnaireViewModel.kt
jingtang10: got it - thanks

jingtang10:
@omarismail94 @FikriMilano I want to bring this PR to your attention too. @LZRS managed to remove one call to initialize the questionnaire state - I wonder if this PR would have a positive impact on the performance. Can you two take a look at this?

Thanks
Jing

datacapture/src/main/java/com/google/android/fhir/datacapture/QuestionnaireViewModel.kt
```diff
@@ -486,7 +478,7 @@ internal class QuestionnaireViewModel(application: Application, state: SavedStat
         viewModelScope,
         SharingStarted.Lazily,
         initialValue =
```
jingtang10, line 480: @yigit is this right? this seems to work for us and is possibly a big improvement gain for us as at the moment it looks like we're doing initialization twice... is the initial state flow value ever useful?

datacapture/src/main/java/com/google/android/fhir/datacapture/QuestionnaireViewModel.kt
LZRS: I tested using the demo and catalog app, that questionnaires are initialized correctly and also ran local unit tests to ensure they were not broke.

Yeah, the questionnaire state was being loaded twice in the `initial` and also when `questionnaireStateFlow` is collected since all the flows in the combine() have default initial states. The `questionnaireStateFlow` is only accessed by collection in the `questionnaireFragment` which essentially made an unnecessary getQuestionnaireState call on initial=, since the flow would also emit on initial collection.

jingtang10: thanks for this

datacapture/src/main/java/com/google/android/fhir/datacapture/QuestionnaireViewModel.kt
```diff
@@ -768,8 +694,14 @@ internal class QuestionnaireViewModel(application: Application, state: SavedStat
...
-            resolveAnswerValueSet = { resolveAnswerValueSet(it) },
-            resolveAnswerExpression = { resolveAnswerExpression(it) },
+            enabledAnswerOptions =
+              answerOptionsEvaluator.evaluate(
+                questionnaireItem,
+                questionnaireResponseItem,
+                questionnaireResponse,
+                questionnaireItemParentMap,
+                answersDisabledCallback = this@QuestionnaireViewModel::removeDisabledAnswers
+              ),
```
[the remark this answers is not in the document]
LZRS: The `answersDisabledCallback ` is used when a previously selected answers in the questionnaireResponse becomes disabled and are no longer in the enabledAnswerOptions. It  removes the disabled answers that are no longer valid in the questionnaireResponse. Some answers may become invalid on re-evaluation of answerOptionsToggle expressions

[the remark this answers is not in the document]
jingtang10: wait hang on - great catch @MJ1998... @LZRS what if instead of passing the callback, we simply remove answers that are not enabled after the calculation. i think having this callback passed into the evaluator is a bit strange.

jingtang10 (changes requested):

jingtang10 (approved):

datacapture/src/main/java/com/google/android/fhir/datacapture/expressions/EnabledAnswerOptionsEvaluator.kt
```diff
@@ -0,0 +1,252 @@
...
+  /**
+   * The answer expression referencing an x-fhir-query has its evaluated data cached to avoid
+   * reloading resources unnecessarily. The value is updated each time an item with answer
+   * expression is evaluating the latest answer options.
+   */
+  private val answerExpressionMap =
+    mutableMapOf<String, List<Questionnaire.QuestionnaireItemAnswerOptionComponent>>()
+
+  /**
+   * Returns allowed answer options that are enabled based on the evaluation of
```
MJ1998, line 55: Can you update the kdoc what this function is returning ?

datacapture/src/main/java/com/google/android/fhir/datacapture/expressions/EnabledAnswerOptionsEvaluator.kt
```diff
@@ -0,0 +1,252 @@
...
+   * [Questionnaire.QuestionnaireItemComponent.answerOptionsToggleExpressions] expressions
+   */
+  internal suspend fun evaluate(
```
MJ1998, line 58: Also this function could be renamed? At first, I thought its evaluating enabled answer options only.

datacapture/src/main/java/com/google/android/fhir/datacapture/QuestionnaireViewModel.kt
```diff
@@ -758,6 +684,20 @@ internal class QuestionnaireViewModel(application: Application, state: SavedStat
...
+        questionnaireResponse,
+        questionnaireItemParentMap
+      )
+    if (disabledQuestionnaireResponseAnswers.isNotEmpty()) {
+      removeDisabledAnswers(
+        questionnaireItem,
+        questionnaireResponseItem,
+        disabledQuestionnaireResponseAnswers
+      )
+    }
```
MJ1998, line 700: I don't understand why we need to invoke `answersChangedCallback` in `removeDisabledAnswers` when constructing the QuestionnaireViewItems. Infact why do we even need fn:`removeDisabledAnswers`..

Correct me if I am wrong. What I think `enabledAnswerOptions` mean is that we calculate a set of enabled-answer-options based on `toggleAnswerOptionExpression`  (by filtering from all possible answer options). Then we can simply pass this list to the QuestionnaireViewItem. Upon change in answer, answersChangedCallback is invoked by the QuestionnaireViewItem which will update the enabledAnswerOptions.

jingtang10 (approved):
thanks for this - just a really small comment otherwise good to merge

PR #2038, opened 2023-06-13
Support basic HTTP authentication

### LINKED ISSUE #2037: Support Basic HTTP authentication
At the moment we only support authentication by token. Whilst this is a widely used method it does not necessarily meet our user's needs. We need to generalize the authenticator interface to support other authentication schemes, most importatly basic http authentication: https://en.wikipedia.org/wiki/Basic_access_authentication

Generalize the authenticator interface and provide a sealed class so that the user can provide information for different authentication methods.

NA

**Additional context**
https://developer.mozilla.org/en-US/docs/Web/HTTP/Authentication

yes

jingtang10:

Fixes #2037 

**Description**
Generalize the authenticator interface to accept multiple authentication methods.

**Alternative(s) considered**
Alternatively, we could open up the interface totally and let the user provide the authorization header, we think this is at this point unnecessary since most users would want to do either basic or bearer authentication. We can consider adding other authentication methods if there is need. Or open up the interface when there's greater need.

**Type**
Feature

**Checklist**


engine/src/main/java/com/google/android/fhir/sync/HttpAuthenticator.kt
```diff
@@ -0,0 +1,52 @@
...
+ * Provides an authorization method for the HTTP requests FHIR Engine sends to the FHIR server.
+ *
+ * FHIR Engine does not handle user authentication. The application should handle user
+ * authentication and provide the appropriate authentication method so the HTTP requests FHIR Engine
+ * sends to the FHIR server contain the correct user information for the request to be
+ * authenticated.
+ */
+interface HttpAuthenticator {
+  fun getAuthenticationMethod(): HttpAuthenticationMethod
+}
```
omarismail94, line 32: is there a reason we need this interface? Can we just rename the sealed interface below to `HttpAuthenticator` to remove the need to call `getAuthenticationMethod`? it would look like this:

```
sealed interface HttpAuthenticator {
  /** @return The authorization header for the engine to make requests on user's behalf. */
  @WorkerThread fun getAuthorizationHeader(): String

  /** See https://datatracker.ietf.org/doc/html/rfc7617. */
  data class Basic(val username: String, val password: String) : HttpAuthenticator {
    override fun getAuthorizationHeader() = Credentials.basic(username, password)
  }

  /** See https://datatracker.ietf.org/doc/html/rfc6750. */
  data class Bearer(val token: String) : HttpAuthenticator {
    override fun getAuthorizationHeader() = "Bearer: $token"
  }
}
```

jingtang10: i thought about this, i think the problem is that the application needs to be able to dynamically provide a different bearer token or username+password pair. but the httpauthenticator is provided to the fhir engine upon initialization.

so if the application decides it needs to provide a different token, we have to add another level of abstraction.

so essentially the http authenticator is like a "authorization header provider provider".... :)

omarismail94: ahhh makes sense! is it possible to add a test that shows this?

jingtang10: sorry for the delay omar...

i decided against adding a test case. i feel it's a bit like testing the return type of a function is of the correct type (e.g. a function returning a string, and we test that it's indeed a string...) - it's superfulous...

but i did improve the documentation of the interface class - please take a look. hope that's enough.

jingtang10:
@omarismail94 can you please take a look at my comment. thanks!

engine/src/main/java/com/google/android/fhir/sync/HttpAuthenticator.kt
omarismail94: would you not be testing that the string changes, which confirms that we can dynamically provide a different bearer token or username+password pair.

omarismail94 (approved):

PR #2047, opened 2023-06-26
Fix Date/DateTime indexes to speed up Left joins on Search 

### LINKED ISSUE #2040: Search method takes more time when load 7k household list
Uploaded 7k household to FHIR server with basic info like name and id. After initial syncing done it's take more time to load at pagination or search by name. It take 2 minute to load all household at SDK.
Like a query below
`SELECT a.serializedResource FROM ResourceEntity a
LEFT JOIN DateIndexEntity b
ON a.resourceType = b.resourceType AND a.resourceUuid = b.resourceUuid AND b.index_name = '_lastUpdated'
LEFT JOIN DateTimeIndexEntity c
ON a.resourceType = c.resourceType AND a.resourceUuid = c.resourceUuid AND c.index_name = '_lastUpdated'
WHERE a.resourceType = 'Group'
AND a.resourceUuid IN (
SELECT resourceUuid FROM TokenIndexEntity
WHERE resourceType = 'Group' AND index_name = 'active' AND index_value = 'true'
)
AND a.resourceUuid IN (
SELECT resourceUuid FROM TokenIndexEntity
WHERE resourceType = 'Group' AND index_name = 'type' AND (index_value = 'person' AND IFNULL(index_system,'') = 'http://hl7.org/fhir/group-type')
)
AND a.resourceUuid IN (
SELECT resourceUuid FROM TokenIndexEntity
WHERE resourceType = 'Group' AND index_name = 'code' AND (index_value = '35359004' AND IFNULL(index_system,'') = 'https://www.snomed.org')
)
ORDER BY b.index_from DESC, c.index_from DESC
LIMIT 10 OFFSET 30`
when using ORDER BY  it's take 50 sec per page or getting search result after search by name.

Device configuration: samsung galaxy tab A 8 enc(SM-T295), ram 2 GB, rom 16 GB.

Yes, we want to work in this issue .We'll need to add order by with household added datetimewise, household serial number,lastvisited timewise. so how can we implement this any tips for this

omarismail94:

Fixes #2040

**Description**

In the Search API, we perform a LEFT JOIN when sorting:

```
LEFT JOIN ${sortTableName.tableName} $tableAlias
      ON a.resourceType = $tableAlias.resourceType AND a.resourceUuid = $tableAlias.resourceUuid AND $tableAlias.index_name = ?
```

this join uses the resourceType, resourceUuid and index_name columns. To speed this up, we should create a composite index made up of these 3 columns

TESTED:
Downloaded 7000 households. Opened App Inspector and ran query generated. Returned results almost instantaneously as opposed to many seconds 

**Alternative(s) considered**

**Type**
Choose one: (**Bug fix** | Feature | Documentation | Testing | Code health | Builds | Releases | Other)

**Checklist**


engine/src/main/java/com/google/android/fhir/db/impl/ResourceDatabase.kt
```diff
@@ -50,7 +50,7 @@ import com.google.android.fhir.db.impl.entities.UriIndexEntity
       LocalChangeEntity::class,
       PositionIndexEntity::class
     ],
-  version = 3,
+  version = 4,
```
aditya-07, line 53: Since version 3 is not released yet, we can keep that version and add the migration as a step in MIGRATION_2_3 itself.

omarismail94: Reverted

engine/src/main/java/com/google/android/fhir/db/impl/ResourceDatabase.kt
```diff
@@ -70,7 +70,22 @@ val MIGRATION_2_3 =
   object : Migration(/* startVersion = */ 2, /* endVersion = */ 3) {
     override fun migrate(database: SupportSQLiteDatabase) {
       database.execSQL(
-        "CREATE INDEX IF NOT EXISTS `index_DateTimeIndexEntity_index_from` ON `DateTimeIndexEntity` (`index_from`)"
+        "CREATE INDEX IF NOT EXISTS `index_DateTimeIndexEntity_resourceType_index_name_resourceUuid_index_from_index_to` ON `DateTimeIndexEntity` (`resourceType`, `index_name`, `resourceUuid`, `index_from`, `index_to`)"
```
jingtang10, line 73: i think this needs to change.

the way migration works is that, if you're on db version n, the migration code would be run to migrate your db schema from n to n+1.

this means you should NOT modify existing migration code - people already on v3 would not get your new db index!

you should create a new version, and write another migration step to handle 3 to 4.

jingtang10 (changes requested):

engine/src/main/java/com/google/android/fhir/db/impl/ResourceDatabase.kt
jingtang10: @aditya-07 fyi since aditya added migration code to begin with.

aditya-07: @jingtang10 DB version in the latest released engine library is 2. DB version 3 was added for a new change and hasn't been released via the library.

jingtang10: sure - but as far as i know @khyativyasargus, @nsabale7, @dubdabasoduba fork our libs... I know theoretically we're not making promises in between releases, but i don't really see the downside of creating another migration step.

omarismail94: I reverted to how I have it but for some reason GitHub hated me doing that on this PR and the change is not getting picked up here. 

Created a new PR: https://github.com/google/android-fhir/pull/2063 that has the migration to a new version

PR #2085, opened 2023-07-20
Fix `engine` and `knowledge` library versions in `workflow` and `testing`

divyaramnath-13:

**Description**
Fix `engine` and `knowledge` library versions in `workflow` and `testing`

**Alternative(s) considered**

**Type**

**Checklist**


buildSrc/src/main/kotlin/Dependencies.kt
```diff
@@ -172,6 +172,12 @@ object Dependencies {
   const val androidFhirEngineModule = "engine"
   const val androidFhirEngine =
     "$androidFhirGroup:$androidFhirEngineModule:${Versions.androidFhirEngine}"
+  const val androidFhirKnowledgeModule = "knowledge"
```
omarismail94, line 175: instead of defining these and their versions (it sort of is getting ugly), can we do this:
```
androidFhirEngine = "${Releases.groupId}:${Releases.Engine.artifactId}:${Releases.Engine.version}"
```
this, way, we don't need to  update the version and all other field in this file, and in Releases

jingtang10: release version and dependency version should be kept separate. this is because we do need to release the library on maven before others can depend on it. so you always need 2 PRs

divyaramnath-13: Got it

testing/build.gradle.kts
```diff
@@ -16,7 +16,7 @@ android {
 }
 
 dependencies {
-  api(project(":engine"))
+  api(Dependencies.androidFhirEngine)
```
jingtang10, line 19: why this change? is this run normally in local testing environment or no kokoro, in which case, should this depend on the project module on HEAD actually?

jingtang10 (approved):
@divyaramnath-13 thanks for this - great job!

just one question actually - please address or leave reasoning. either way happy for this to be merged.

testing/build.gradle.kts
divyaramnath-13: This is because `testing` is a dependency in `workflow` and they would depend on different versions of `engine` without this change.

PR #2137, opened 2023-08-17
Add ResourceConsolidator

### LINKED ISSUE #2103: Upload Work Manager - API refactoring
Discussed this with @anchita-g. One alternative way of handling this is to make the `UploadWorkManager` a little more stateful, similar to the `DownloadWorkManager`. In that case, there would be an API to feed local changes to the `UploadWorkManager`, there will be an API to pull next upload request to be sent to the FHIR server, and there'll be an API to get the sync status - almost like an iterator with a sync progress.

That would make the job of `Uploader` a little simpler - the while loop currently in `UplaoderImpl` would be simplified (because the work is done by the `UploadWorkManager`. This affords the developers a little more control - but admittedly we're not entirely sure if the additional contronl is urgently needed by the developers.

But that would make the design more alinged with the download workflow which @omarismail94 worked on.

FYI @aditya-07 as well.

_Originally posted by @jingtang10 in https://github.com/google/android-fhir/pull/2022#discussion_r1277357351_

omarismail94:

Fixes #2103

**Description**

The resource consolidator would handle any post-processing that needs to be made to the database. Examples of this would be, updating the lastUpdated timestamp field, or deleting the local change from the database, or updating the resource IDs and payloads to correspond with the server’s feedback.

**Alternative(s) considered**

**Type**

**Checklist**


engine/src/main/java/com/google/android/fhir/impl/FhirEngineImpl.kt
```diff
@@ -123,89 +123,17 @@ internal class FhirEngineImpl(private val database: Database, private val contex
...
-    upload: suspend (List<LocalChange>) -> Flow<Pair<LocalChangeToken, Resource>>
+    uploadMode: UploadMode,
+    upload: suspend (List<LocalChange>, ResourceConsolidator) -> Unit,
   ) {
+    val resourceConsolidator =
+      when (uploadMode.resourceConsolidatorStrategy) {
+        ResourceConsolidatorStrategy.Default -> DefaultResourceConsolidator(database)
+        ResourceConsolidatorStrategy.IDUpdater -> error("Not yet implemented")
+      }
     val localChanges = database.getAllLocalChanges()
```
anchita-g, line 134: We would have to modify the fetcher as well based on the upload mode

engine/src/main/java/com/google/android/fhir/sync/upload/UploaderImpl.kt
```diff
@@ -43,7 +43,10 @@ internal class UploaderImpl(
   private val uploadWorkManager: UploadWorkManager,
 ) : Uploader {
 
-  override suspend fun upload(localChanges: List<LocalChange>): Flow<UploadState> = flow {
+  override suspend fun upload(
+    localChanges: List<LocalChange>,
```
anchita-g, line 47: Would we not need `UploadMode` here to decide how to create patches of request and request builder.

omarismail94: I didn't add the patch and request builder as they have yet to be created. In the future, the upload function would look like this:

```
override suspend fun upload(localChangeFetcher: LocalChangeFetcher, patchGenerator: PatchGenerator, requestBuilder: RequestBuilder, resourceConsolidator: ResourceConsolidator): Flow<UploadState>
```

determining what to pass into the `upload` function based on UploadMode can be done either in the FhirSynchronizer, or in FhirEngineImpl.syncUpload. For the latter, it is for interfaces whose initialisation relies on variables that are only accessible from FhirEngine, such as `Database`

omarismail94: Obsolete now as I removed UploadMode from this PR

engine/src/main/java/com/google/android/fhir/FhirEngine.kt
```diff
@@ -54,7 +55,8 @@ interface FhirEngine {
    * api caller should [Flow.collect] it.
    */
   suspend fun syncUpload(
-    upload: (suspend (List<LocalChange>) -> Flow<Pair<LocalChangeToken, Resource>>)
+    consolidatorMode: ConsolidatorMode,
```
jingtang10, line 58: this is unfortuantely glossed over in our previous discussions. do we really need a consolidator mode? or maybe one consolidator is enough? the same consolidator can clear the local changes and update resource id if needed?

jingtang10 (changes requested):

engine/src/main/java/com/google/android/fhir/sync/upload/ResourceConsolidator.kt
```diff
@@ -0,0 +1,123 @@
...
+      when (mode) {
+        is ConsolidatorMode.Default -> DefaultResourceConsolidator(database)
+        else -> error("$mode does not have an implementation yet.")
+      }
+  }
+}
+
+internal class DefaultResourceConsolidator(private val database: Database) : ResourceConsolidator {
+
+  override suspend fun consolidate(localChangeToken: LocalChangeToken, response: Resource) {
```
anchita-g, line 39: Will the information from the `localChangeToken` be enough to perform resource consolidation in more complex cases like ID updation, I was thinking there needs to be some kind of indicator for if the changes were at a per resource level, or were they POST changes, was the request a bundle?

engine/src/main/java/com/google/android/fhir/FhirEngine.kt
omarismail94: I thought about this, and I think one consolidator should be enough. Within the default consolidator, we could have when statements based on the HTTP verb

engine/src/main/java/com/google/android/fhir/sync/upload/ResourceConsolidator.kt
omarismail94: I think it should be? If it isn't, it should be able to figure out the consolidation by querying the database (that could come at a performance cost)

engine/src/main/java/com/google/android/fhir/sync/upload/ResourceConsolidator.kt
```diff
@@ -0,0 +1,113 @@
...
+package com.google.android.fhir.sync.upload
+
+import com.google.android.fhir.LocalChangeToken
+import com.google.android.fhir.db.Database
+import org.hl7.fhir.r4.model.Bundle
+import org.hl7.fhir.r4.model.Resource
+import org.hl7.fhir.r4.model.ResourceType
+import timber.log.Timber
+
+/** Represents a mechanism to consolidate resources after they are uploaded. */
```
jingtang10, line 26: this kdoc can probably be enriched a bit - why do we have this? what is this class actually gonna do? any edge cases?

PR #2178, opened 2023-09-16
Add API to support saving local only resources

### LINKED ISSUE #2123: New API to support saving local only resources
Sometimes we need to persist resources on the device's database without creating local changes to be synced to the server. Such resources may be downloaded from the server or perhaps generated by the client itself.

Currently the only way to persist a resource is to use the `FhirEngine.create` API method which flags the resources for sync.

This could be solved by exposing the `Database.insertRemote` API via the `FhirEngine` API by adding the following interface and implementation.

```kotlin
//FhirEngine.kt
suspend fun createRemote(vararg resource: Resource)
```

```kotlin
//FhirEngineImpl.kt
  override suspend fun createRemote(vararg resource: Resource){
    return database.insertRemote(*resource)
  }
```

An alternative approach could be to enhance the current `FhirEngine.create` API and pass a flag to determine whether it should be marked for sync or is a local only resource. 

```kotlin
//FhirEngine.kt
suspend fun create(val isLocalOnlyChange:Boolean = false, vararg resource: Resource): List<String>
```

Might be better to name the flag something like `isSyncable` or `syncUp` and set its default to `true` to avoid any confusion with the existing concept of local changes.

**Additional context**
N/A

Yes

ndegwamartin:

Fixes #2123 

**Description**
See #2123

**Alternative(s) considered**
See #2123 

**Type**
Feature

N/A

**Checklist**

### DIFF
--- a/engine/src/main/java/com/google/android/fhir/FhirEngine.kt
+++ b/engine/src/main/java/com/google/android/fhir/FhirEngine.kt
@@ -33,6 +33,10 @@ interface FhirEngine {
    */
   suspend fun create(vararg resource: Resource): List<String>
 
+  /**
+   * Creates one or more remote FHIR [resource]s in the local storage without flagging as local
+   * changes.
+   */
   suspend fun createRemote(vararg resource: Resource)
 
   /** Loads a FHIR resource given the class and the logical ID. */
@@ -55,7 +59,7 @@ interface FhirEngine {
    * api caller should [Flow.collect] it.
    */
   suspend fun syncUpload(
-    upload: (suspend (List<LocalChange>) -> Flow<Pair<LocalChangeToken, Resource>>)
+    upload: (suspend (List<LocalChange>) -> Flow<Pair<LocalChangeToken, Resource>>),
   )
 
   /**
@@ -64,7 +68,7 @@ interface FhirEngine {
    */
   suspend fun syncDownload(
     conflictResolver: ConflictResolver,
-    download: suspend () -> Flow<List<Resource>>
+    download: suspend () -> Flow<List<Resource>>,
   )
 
   /**
@@ -89,29 +93,32 @@ interface FhirEngine {
    * Retrieves a list of [LocalChange]s for [Resource] with given type and id, which can be used to
    * purge resource from database. If there is no local change for given [resourceType] and
    * [Resource.id], return an empty list.
+   *
    * @param type The [ResourceType]
    * @param id The resource id [Resource.id]
    * @return [List]<[LocalChange]> A list of local changes for given [resourceType] and
-   * [Resource.id] . If there is no local change for given [resourceType] and [Resource.id], return
-   * an empty list.
+   *   [Resource.id] . If there is no local change for given [resourceType] and [Resource.id],
+   *   return an empty list.
    */
   suspend fun getLocalChanges(type: ResourceType, id: String): List<LocalChange>
 
   /**
    * Purges a resource from the database based on resource type and id without any deletion of data
    * from the server.
+   *
    * @param type The [ResourceType]
    * @param id The resource id [Resource.id]
    * @param isLocalPurge default value is false here resource will not be deleted from
-   * LocalChangeEntity table but it will throw IllegalStateException("Resource has local changes
-   * either sync with server or FORCE_PURGE required") if local change exists. If true this API will
-   * delete resource entry from LocalChangeEntity table.
+   *   LocalChangeEntity table but it will throw IllegalStateException("Resource has local changes
+   *   either sync with server or FORCE_PURGE required") if local change exists. If true this API
+   *   will delete resource entry from LocalChangeEntity table.
    */
   suspend fun purge(type: ResourceType, id: String, forcePurge: Boolean = false)
 }
 
 /**
  * Returns a FHIR resource of type [R] with [id] from the local storage.
+ *
  * @param <R> The resource type which should be a subtype of [Resource].
  * @throws ResourceNotFoundException if the resource is not found
  */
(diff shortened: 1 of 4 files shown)


engine/src/main/java/com/google/android/fhir/FhirEngine.kt
```diff
@@ -34,6 +34,12 @@ interface FhirEngine {
    */
   suspend fun create(vararg resource: Resource): List<String>
 
+  /**
+   * Creates one or more remote FHIR [resource]s in the local storage without flagging as local
+   * changes.
+   */
+  suspend fun createRemote(vararg resource: Resource)
+
```
aditya-07, line 42:
```suggestion
  /**
   * Creates one or more FHIR [resource]s in the local storage.
   *
   * @param isLocalOnly - Setting the value to [true] instructs engine that the resource and its subsequent updates should never be synced to the server.
   * @return the logical IDs of the newly created resources.
   */
    suspend fun create(vararg resource: Resource, isLocalOnly: Boolean = false): List<String> 
```

The default value can be provided in the interface i.e [FhirEngine.kt](https://github.com/google/android-fhir/pull/2178/files#diff-c31e3e53f14056cbc724b023efe0b83d19a7d82d9c9332197e1bb1bd99e08bb1) and the [FhirEngineImpl.kt](https://github.com/google/android-fhir/pull/2178/files#diff-5b081449c3e8ee8c1459f4f7b52966b1a63fe92d79beaee1b8ed0a4b77a14626) would be called with that value if no value is provided by the caller.

engine/src/main/java/com/google/android/fhir/impl/FhirEngineImpl.kt
```diff
@@ -44,6 +44,10 @@ internal class FhirEngineImpl(private val database: Database, private val contex
     return database.insert(*resource)
   }
 
+  override suspend fun createRemote(vararg resource: Resource) {
+    return database.insertRemote(*resource)
+  }
+
```
aditya-07, line 50:
```suggestion
```
`insertRemote` would just add the resource in the db without creating a `LocalChange` counterintuitively marking the change as local only.

aditya-07: Looking at the test, createRemote is doing what it is supposed to do. But, the name `createRemote` is counterintuitive and suggests that the resource has something to do with the remote server. A more intutive name could have be `createLocal` ?

aditya-07 (changes requested):
IMO,  adding a default value to `FhirEngine.create` to specify whether the resources are meant to be local only is a better design choice.

Additionally,  we would want all the updates to the resource to be local only as well. Therefore, we should mark the resource as local only in the db itself.

ndegwamartin:
@aditya-07 could you have a second look to see if this is what you had in mind?

ndegwamartin:
@jingtang10 you had a concern with the _updates_ for the related resources processed by the implementation here, could you share that here or on the ticket #2123 ?

aditya-07 (commented):
Update and Delete changes will still create a LocalChange that maybe synced to the server.

ndegwamartin:
From [this line - ResourceDao.kt#L250](https://github.com/google/android-fhir/blob/f107597d56efcedb27e7945011186465a10c6092/engine/src/main/java/com/google/android/fhir/db/impl/dao/ResourceDao.kt#L250) it appears that the database API actually performs an upsert.

jingtang10:
> From [this line - ResourceDao.kt#L250](https://github.com/google/android-fhir/blob/f107597d56efcedb27e7945011186465a10c6092/engine/src/main/java/com/google/android/fhir/db/impl/dao/ResourceDao.kt#L250) it appears that the database API actually performs an upsert.

what do you mean? why does this matter?

i think we still need to make db change first for this pr to work properly. i.e. creating a new column to indicate local only resources.

ndegwamartin:
> From [this line - ResourceDao.kt#L250](https://github.com/google/android-fhir/blob/f107597d56efcedb27e7945011186465a10c6092/engine/src/main/java/com/google/android/fhir/db/impl/dao/ResourceDao.kt#L250) it appears that the database API actually performs an upsert.

This was a response to Aditya's comment and feedback from the discussion last time but essentially in the current implementation the API can cater for both _Creates_ and _Updates_, we'd just need to rename the method to reflect this.

Delete of course would still not be handled hence why we are exploring the db change option to capture all ops.

joiskash (changes requested):
Please review the Database Entities, one approach would be to add a boolean field to differentiate between local only and remote. Another idea might be creating a new table to contain local resources. 

1. How would existing app migrate to this version? All existing resources could be considered remote.
2. What about changes to local only resources? Can I save a resource as local only and then update it to be synced to server?
3. What would happen to the search APIs? would there be a use case where you would want to only search for local resources vs only remote vs both?

It would be good to get into the use cases and what we would like to support initially. It is not very clear from the issue #2123 as to why this might be needed. 
eg: Multiple FHIR servers, to simplify peer to peer sync strategies, save resources that are incomplete eg: a Questionnaire partially filled

ellykits:
> It would be good to get into the use cases and what we would like to support initially. It is not very clear from the issue https://github.com/google/android-fhir/issues/2123 as to why this might be needed.
eg: Multiple FHIR servers, to simplify peer to peer sync strategies, save resources that are incomplete eg: a Questionnaire partially filled

The use case @ndegwamartin was attempting to address is where we'd want to download particular resources from the server (which might have been generated from other sources) to the Android client but prevent the app from uploading any changes made on these resources to the server. For instance Questionnaires, StructureMaps, Binary resource etc as used in OpenSRP.

ndegwamartin:
Changing this to Draft for now since a workaround solution for the initial problem was put forward

ndegwamartin:
This is blocked by/Superseded by this discussion - https://github.com/google/android-fhir/discussions/2625

jingtang10:
> This is blocked by/Superseded by this discussion - #2625

why is this blocked by custom resources?

PR #2179, opened 2023-09-18
Remove left over *.json.bak file

vorburger:
Removing a file that look like it was probably committed in error.

It is not referenced anywhere.


jingtang10 (approved):
Thank you for the clean-up @vorburger!

Issue #2194, opened 2023-09-25
Set up Dependency graph for Dependencies and Dependents


vorburger:
While looking at #2185 I noticed something else technically unrelate dwhich would also be good to set-up:

* https://github.com/google/android-fhir/network/dependencies
* https://github.com/google/android-fhir/network/dependents

are empty, but can be interesting to have (for e.g. both security as well as seeing who's using us).

https://github.com/MariaDB4j/MariaDB4j/network/dependencies is how the result would look like.

I'll have a go at seeing if I can this up (it should either be not very hard, or if there is something particular I'm missing that's specific to Gradle for Android that's a big enough PITA that it's not worth pursuing it further).

@fredhersch @jingtang10 @omarismail94 @williamito  FYI

vorburger:
OK, so according to https://github.com/marketplace?query=dependency+submission+gradle, there are apparently (at least) 2-3 choices here:

1. https://github.com/marketplace/actions/gradle-buildjet-action
2. https://github.com/marketplace/actions/gradle-build-action
3. https://github.com/marketplace/actions/gradle-dependency-submission

The 1. looks, to me, like an extremely suspicious rip-off from 2. so let's stay clear of that! (Note it has 0 Stars, and the exact same README... that's kind of textbook what one would do if I tried to take over a popular GitHub Action for malicious purposes! It may also be harmless, of course; but let's not find out.)

The 2. looks great, albeit a bit.. overwhelming? It does lots of great looking stuff (optimized caching; thank you!) - but I'm not 100% convinced we could use it JUST to _Generate and Submit a GitHub Dependency Graph_ ... maybe, or maybe not; I have a hunch that until we sort out that whole _Build with Google Kokoro vs. GitHub Action_ can of worms, it could be overkill. Its doc also refers to https://github.com/gradle/github-dependency-graph-gradle-plugin, with a README that says _support coming in a future release,_ which doesn't inspire confidence in me (and each new Gradle plugin can be a PITA for future upgrades; I've been there).

The 3. looks simpler and focused only on dependency submission, and is apparently essentially based on just e.g. `./gradlew demo:dependencies` - without requiring adding any additional Gradle plugins - I like the sound of that! Hear _Gradle 7.5 or newer is required_ but that's cool because [we're on 8.0](https://github.com/google/android-fhir/blob/master/gradle/wrapper/gradle-wrapper.properties). I'll therefore further pursue this choice!

vorburger:
[...]

PR #2196, opened 2023-09-25
Create dependabot.yaml (fixes #2195)

### LINKED ISSUE #2195: Set up Dependabot
While looking at https://github.com/google/android-fhir/issues/2185 I noticed something else technically unrelate which IMHO would also be good to set-up:

https://docs.github.com/en/code-security/dependabot/dependabot-version-updates

I'll have a go at seeing if I can enable this (it should either be not very hard, or if there is something particular I'm missing that's specific to Gradle for Android that's a big enough PITA that it's not worth pursuing it further).

The expected result would be to start seeing https://github.com/MariaDB4j/MariaDB4j/pulls?q=is%3Apr++label%3Adependencies+ on this repo. (And on https://github.com/google/android-fhir/security/dependabot as well as on https://github.com/google/android-fhir/network/updates) We can tune the frequency and such things, if required (I've done it before).

@fredhersch @jingtang10 @omarismail94 @williamito FYI (please do shout here if there is any particular reason why this project would not want to benefit from automated dependency upgrades).

vorburger:
Fixes #2195

**Description**
This (should) enable https://github.com/dependabot, see https://docs.github.com/en/code-security/dependabot.

**Alternative(s) considered**

> Have you considered any alternatives? And if so, why have you chosen the approach in this PR?

Yes, always! 😺  I am aware of other such tools, but this is the best one, because it's from GitHub themselves, and thus most natively integrated. It also work great on many other open source projects where I have enabled it! 😃 

**Type**
Choose one: ~Bug fix | Feature | Documentation | Testing |~ Code health | Builds | Releases | Security

The expected result would be to start seeing [automagical Pull Requests such as these](https://github.com/MariaDB4j/MariaDB4j/pulls?q=is%3Apr++label%3Adependencies+) (from another repo) on this repo.


jingtang10:
@vorburger do we know if this works before merging the PR?

We have a centralised place to keep the dependency info: https://github.com/google/android-fhir/blob/master/buildSrc/src/main/kotlin/Dependencies.kt in the `buildSrc` folder. Is dependabot able to handle this?

vorburger:
> @vorburger do we know if this works before merging the PR?

No, not easily, unfortunately. My recommendation is to just merge this as-is; we could always `git revert` later, if it's causing problems.

> We have a centralised place to keep the dependency info: https://github.com/google/android-fhir/blob/master/buildSrc/src/main/kotlin/Dependencies.kt in the `buildSrc` folder. Is dependabot able to handle this?

Yes, I expect it will; note how, since #2208 for #2194, https://github.com/google/android-fhir/network/dependencies now shows this project's dependencies; my expectation is that when this PR gets merged, Dependabot will start proposing upgrades.

In order to easily see which upgrades proposed by Dependabot "work" , perhaps you would like to merge #2203 to fix #2197 before this?

vorburger:
> Yes, I expect it will; note how ...

Actually... I see what you mean; I'm not sure if Dependabot is smart enough to propose a correct PR, for this project as-is. We may have to do some manual work. Let's try and see? If it doesn't work, perhaps we could rework `Dependencies.kt`?

jingtang10 (approved):
happy for you to merge this but have you seen this: https://medium.com/@vladyslav.hontar/dependabot-in-action-d9b56b2be86c?

i commented in the issue too

Issue #1482, opened 2022-06-29
Espresso Test fails if Screen Size too small


omarismail94:
**Describe the bug**
In the [QuestionnaireItemDialogMultiSelectViewHolderFactoryEspressoTest](https://github.com/google/android-fhir/blob/master/datacapture/src/androidTest/java/com/google/android/fhir/datacapture/views/QuestionnaireItemDialogMultiSelectViewHolderFactoryEspressoTest.kt#L56) test class, the test involves clicking on text. Because my screen is tiny,  "Coding 3" is not fully!! within view, and the test fails with the error:
No views in hierarchy found matching: an instance of android.widget.TextView and view.getText() with or without transformation to match: is "Coding 3"

**Component**
Core library, **SDC library**, reference app, or SDC gallery app

**To Reproduce**
Setup screen with small size. 
Run: `./gradlew datacapture:connectedDebugAndroidTest`

**Expected behavior**

Regardless of screensize, test should work

**Screenshots**

**Would you like to work on the issue?**
Sure

vorburger:
In #2197 I have found that this problem appears to frequently make PRs "red", IMHO it would be good to fix this ASAP.  Until it is, IMHO it's a lot less confusing not to run that test anymore; see #2203. BTW the exact error seen there was:

_Nexus6P-24-en_US-portrait : com.google.android.fhir.datacapture.test.views.QuestionnaireItemDialogMultiSelectViewHolderFactoryEspressoTest#selectOther_shouldScrollDownToShowAddAnotherAnswer (1) [androidx.test.espresso.NoMatchingViewException: No views in hierarchy found matching: view.getId() is <2131230791/com.google.android.fhir.datacapture.test:id/add_another>](https://console.firebase.google.com/project/android-fhir-instrumeted-tests/testlab/histories/bh.96901a7d2453dacf/matrices/5012575031813075587/executions/bs.812d5bb001a62209/testcases/3)_

vorburger:
[Here](https://firebase.corp.google.com/project/android-fhir-instrumeted-tests/testlab/histories/bh.96901a7d2453dacf/matrices/5229748969591520332/executions/bs.24456ec5fee68df1/testcases/4/errors) for #2316 that `selectOther_shouldScrollDownToShowAddAnotherAnswer()` in the `QuestionnaireItemDialogMultiSelectViewHolderFactoryEspressoTest` now failed even on API Level 33, so PR #2318 proposes to entirely `@Ignore` it:

```
[...]

PR #2203, opened 2023-09-26
Ignore flaky UI Test, for now (see #1482, fixes #2197)

### LINKED ISSUE #2197: CI is broken and fails for trivial PR
**Describe the bug**
My trivial PR #2196, which only adds a `.github/dependabot.yaml` file, which IMHO is "impossible" to affect test runs and such things, failed to pass CI and is red instead of green.

This can discourage contributions (because you never clearly know if the change proposed in a PR breaks CI, or if it was "just already broken anyway"), and therefore perhaps should be fixed with top priority, over everything else?

IMHO any project should intentionally set itself a relatively high bar in this matter, and aim for CI on its main branch to always be green. Any PRs that break it and turn it red should be immediately be reverted by maintainers.

Glancing over https://github.com/google/android-fhir/commits/master this historically does not seem to be the case in this project.

**To Reproduce**
1. Raise a trivial PR e.g. adding a space to the README

**Expected behavior**
It can never say _Some checks were not successful_ for such trivial PR changes.

I am happy to help "drive" this, but suspect it will need input from more people.

/Cc @fredhersch @jingtang10 @omarismail94 @williamito FYA

vorburger:
Fixes #2197, see also #1482

**Description**
This disables a known flaky UI test. This reduces confusion on why PRs fail to build.

**Alternative(s) considered**
_Have you considered any alternatives? And if so, why have you chosen the approach in this PR?_
The alternative is to fix the flaky UI test. My Android foo isn't strong enough to know how to do that, unfortunately.

**Type**
Choose one: **Bug fix** ~Feature | Documentation | Testing | Code health | Builds | Releases | Other~


datacapture/src/androidTest/java/com/google/android/fhir/datacapture/test/views/QuestionnaireItemDialogMultiSelectViewHolderFactoryEspressoTest.kt
```diff
@@ -261,6 +262,7 @@ class QuestionnaireItemDialogMultiSelectViewHolderFactoryEspressoTest {
   }
 
   @Test
+  @Ignore // TODO https://github.com/google/android-fhir/issues/1482 FIXME
```
omarismail94, line 265: @Ignore takes a parameter where you can add the link:
```suggestion
  @Ignore("https://github.com/google/android-fhir/issues/1482")
```

jingtang10 (changes requested):
sorry for the delay... i thought i sent this out: but my main comment here is can you try to ignore this only for api level 24? this is not flaky on api level 33.

vorburger:
> sorry for the delay... i thought i sent this out: but my main comment here is can you try to ignore this only for api level 24? this is not flaky on api level 33.

I have no idea if it's possible to (easily) do that. @omarismail94 would you happen to know?

If it's not, my recommendation is to ignore it for all, until fixed, to have green CI.

omarismail94:
> > sorry for the delay... i thought i sent this out: but my main comment here is can you try to ignore this only for api level 24? this is not flaky on api level 33.
> 
> I have no idea if it's possible to (easily) do that. @omarismail94 would you happen to know?
> 
> If it's not, my recommendation is to ignore it for all, until fixed, to have green CI.

In the flaky test, you can add this block:

```
    @Test
    fun testMethodToIgnoreOnApi24() {
        // Ignore this specific test if the API level is 24
        Assume.assumeTrue(Build.VERSION.SDK_INT != Build.VERSION_CODES.N)
        
        // Your test code here...
    }
```

https://github.com/google/android-fhir/pull/2203#issuecomment-1738892206

vorburger:
An alternative suggested by @jingtang10 is https://developer.android.com/reference/androidx/test/filters/SdkSuppress.

omarismail94 (commented):
@vorburger can you also add the same thing to:
`clickAddAnotherAnswer_shouldScrollDownToShowAddAnotherAnswer`

It is the other flaky test this library

omarismail94 (approved):

jingtang10 (approved):

PR #2210, opened 2023-09-27
Add resource UUID to LocalChangeEntity, fixing changing UUID issue for the same resource 

### LINKED ISSUE #2209: Add resource UUID to LocalChangeEntity
We know that the resource ID of the LocalChangeEntity might change when the IDs are assigned by the server. In such a case linking LocalChangeEntity to ResourceEntity through resource ID will break things. Hence we need to have a linking through an ID which would not change for the resource on the device locally in its entire time of existence (till it is deleted from the ResourceEntity table). We can use ResourceEntity's resourceUuid for this.

Add resourceUuid field to the LocalChangeEntity

N/A

**Additional context**

yes

anchita-g:

Fixes #2209 

**Description**
- We need to LocalChangeEntity agnostic of the resource ID as there can be cases where the resource ID of a resource might change (POST for creation). In such cases we can use resource UUID to keep the mapping between the resource and LocalChanges intact.
- Fixing the `ResourceDatabaseMigrationTest` to use the migrated database for the particular implemented migration to run assertions
- There are cases where in download of a resource which already exists in the database leads to new resource UUID being assigned to the same resource. Hence we need to keep the same resource UUID as the previous version in such cases

**Alternative(s) considered**
One alternative in the first scenario was to just keep the resourceUuid and remove resourceId. However, when LocalChanges are uploaded to the server, we would need to look up the ResourceEntity using the resourceUuid in the LocalChangeEntity to obtain the server recognisable identifier of the resource. This would be a problem in case of deleted resources, where in for any LocalChange of a resource that has been deleted, we would not be able to find the corresponding ResourceEntity. Hence, keeping resourceId is also crucial

**Type**
Choose one:  Feature

**Checklist**

### DIFF
--- a/engine/src/androidTest/java/com/google/android/fhir/db/impl/ResourceDatabaseMigrationTest.kt
+++ b/engine/src/androidTest/java/com/google/android/fhir/db/impl/ResourceDatabaseMigrationTest.kt
@@ -16,7 +16,6 @@
 
 package com.google.android.fhir.db.impl
 
-import androidx.room.Room
 import androidx.room.testing.MigrationTestHelper
 import androidx.test.ext.junit.runners.AndroidJUnit4
 import androidx.test.platform.app.InstrumentationRegistry
@@ -30,7 +29,6 @@ import java.util.Date
 import kotlinx.coroutines.runBlocking
 import org.hl7.fhir.r4.model.HumanName
 import org.hl7.fhir.r4.model.Patient
-import org.hl7.fhir.r4.model.ResourceType
 import org.hl7.fhir.r4.model.Task
 import org.junit.Rule
 import org.junit.Test
@@ -69,12 +67,14 @@ class ResourceDatabaseMigrationTest {
 
     // Open latest version of the database. Room will validate the schema
     // once all migrations execute.
-    helper.runMigrationsAndValidate(DB_NAME, 2, true, MIGRATION_1_2)
+    val migratedDatabase = helper.runMigrationsAndValidate(DB_NAME, 2, true, MIGRATION_1_2)
 
     val readPatientJson: String?
-    getMigratedRoomDatabase().apply {
-      readPatientJson = this.resourceDao().getResource("migrate1-2-test", ResourceType.Patient)
-      openHelper.writableDatabase.close()
+    migratedDatabase.apply {
+      query("SELECT serializedResource FROM ResourceEntity").let {
+        it.moveToFirst()
+        readPatientJson = it.getString(0)
+      }
     }
 
     assertThat(readPatientJson).isEqualTo(insertedPatientJson)
@@ -101,12 +101,14 @@ class ResourceDatabaseMigrationTest {
     }
 
     // Re-open the database with version 3 and provide MIGRATION_2_3 as the migration process.
-    helper.runMigrationsAndValidate(DB_NAME, 3, true, MIGRATION_2_3)
+    val migratedDatabase = helper.runMigrationsAndValidate(DB_NAME, 3, true, MIGRATION_2_3)
 
     val retrievedTask: String?
-    getMigratedRoomDatabase().apply {
-      retrievedTask = this.resourceDao().getResource(taskId, ResourceType.Task)
-      openHelper.writableDatabase.close()
+    migratedDatabase.apply {
+      query("SELECT serializedResource FROM ResourceEntity").let {
+        it.moveToFirst()
+        retrievedTask = it.getString(0)
+      }
     }
 
     assertThat(retrievedTask).isEqualTo(bedNetTask)
@@ -133,12 +135,14 @@ class ResourceDatabaseMigrationTest {
     }
 
     // Re-open the database with version 4 and provide MIGRATION_3_4 as the migration process.
-    helper.runMigrationsAndValidate(DB_NAME, 4, true, MIGRATION_3_4)
+    val migratedDatabase = helper.runMigrationsAndValidate(DB_NAME, 4, true, MIGRATION_3_4)
 
     val retrievedTask: String?
-    getMigratedRoomDatabase().apply {
-      retrievedTask = this.resourceDao().getResource(taskId, ResourceType.Task)
-      openHelper.writableDatabase.close()
+    migratedDatabase.apply {
+      query("SELECT serializedResource FROM ResourceEntity").let {
+        it.moveToFirst()
+        retrievedTask = it.getString(0)
+      }
     }
 
     assertThat(retrievedTask).isEqualTo(bedNetTask)
@@ -164,12 +168,14 @@ class ResourceDatabaseMigrationTest {
     }
 
     // Re-open the database with version 5 and provide MIGRATION_4_5 as the migration process.
-    helper.runMigrationsAndValidate(DB_NAME, 5, true, MIGRATION_4_5)
+    val migratedDatabase = helper.runMigrationsAndValidate(DB_NAME, 5, true, MIGRATION_4_5)
 
     val retrievedTask: String?
-    getMigratedRoomDatabase().apply {
-      retrievedTask = this.resourceDao().getResource(taskId, ResourceType.Task)
-      openHelper.writableDatabase.close()
+    migratedDatabase.apply {
+      query("SELECT serializedResource FROM ResourceEntity").let {
+        it.moveToFirst()
+        retrievedTask = it.getString(0)
+      }
     }
 
     assertThat(retrievedTask).isEqualTo(bedNetTask)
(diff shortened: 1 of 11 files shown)


engine/src/main/java/com/google/android/fhir/db/impl/dao/ResourceDao.kt
```diff
@@ -183,11 +185,14 @@ internal abstract class ResourceDao {
...
   // Since the insert removes any old indexes and lastUpdatedLocal (data not contained in resource
   // itself), we extract the lastUpdatedLocal if any and then set it back again.
-  private suspend fun insertRemoteResource(resource: Resource) =
-    insertResource(
-      resource,
-      getResourceEntity(resource.logicalId, resource.resourceType)?.lastUpdatedLocal,
-    )
+  private suspend fun insertRemoteResource(resource: Resource): String {
+    val existingResourceEntity = getResourceEntity(resource.logicalId, resource.resourceType)
+    if (existingResourceEntity != null) {
```
omarismail94, line 190: is this code relevant to adding resource UUID to LocalChangeEntity? If not, can we remove it and have it in another PR?

omarismail94: I see that in the PR description:
```
There are cases where in download of a resource which already exists in the database leads to new resource UUID being assigned to the same resource. Hence we need to keep the same resource UUID as the previous version in such cases
```

did this problem exist before this change?

engine/src/main/java/com/google/android/fhir/db/impl/dao/ResourceDao.kt
```diff
@@ -76,12 +76,14 @@ internal abstract class ResourceDao {
...
-              ),
-            )
+            timeOfLocalChange?.let {
+              addDateTimeIndex(
+                createLocalLastUpdatedIndex(
+                  resource.resourceType,
+                  InstantType(Date.from(timeOfLocalChange)),
+                ),
+              )
+            }
```
omarismail94, line 86: same comment as above about `timeOfLocalChange` being nullable

engine/src/main/java/com/google/android/fhir/db/impl/DatabaseImpl.kt
```diff
@@ -115,8 +122,10 @@ internal class DatabaseImpl(
       logicalIds.addAll(
         resource.map {
           val timeOfLocalChange = Instant.now()
-          localChangeDao.addInsert(it, timeOfLocalChange)
-          resourceDao.insertLocalResource(it, timeOfLocalChange)
+          val resourceId = resourceDao.insertLocalResource(it, timeOfLocalChange)
+          val resourceEntity = selectEntity(it.resourceType, it.logicalId)
```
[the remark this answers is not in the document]
omarismail94: should we just have `insertLocalResource` return `resourceUUId`, or a `Pair<resourceId, recourceUUID>` to avoid making extra calls to the DB?

engine/src/androidTest/java/com/google/android/fhir/db/impl/ResourceDatabaseMigrationTest.kt
```diff
@@ -69,12 +67,14 @@ class ResourceDatabaseMigrationTest {
 
     // Open latest version of the database. Room will validate the schema
     // once all migrations execute.
-    helper.runMigrationsAndValidate(DB_NAME, 2, true, MIGRATION_1_2)
+    val migratedDatabase = helper.runMigrationsAndValidate(DB_NAME, 2, true, MIGRATION_1_2)
```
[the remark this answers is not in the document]
anchita-g: Yes, the migrations before did not change any existing table by adding any non-null new columns. Hence running entire migrations till the latest version and then testing only against the specific migration did not cause any problems.

engine/src/main/java/com/google/android/fhir/db/impl/dao/ResourceDao.kt
```diff
@@ -58,7 +58,7 @@ internal abstract class ResourceDao {
   lateinit var iParser: IParser
   lateinit var resourceIndexer: ResourceIndexer
 
-  open suspend fun update(resource: Resource, timeOfLocalChange: Instant) {
+  open suspend fun update(resource: Resource, timeOfLocalChange: Instant?) {
```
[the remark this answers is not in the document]
anchita-g: Yes, this is for cases where we update the resource when it is downloaded from the server. As of now we insert a new resource entity which leads to change in resource UUID. We need to instead update it and it can happen that the resource did not change at all since it was last donwloaded and hence, `timeOfLocalChange` can be nullable

engine/src/main/java/com/google/android/fhir/db/impl/dao/ResourceDao.kt
anchita-g: Addressed it in the above comment. Thanks!

engine/src/main/java/com/google/android/fhir/db/impl/dao/ResourceDao.kt
anchita-g: Yes, every time we downloaded an entity, we assigned it a new resource UUID

engine/src/main/java/com/google/android/fhir/db/impl/DatabaseImpl.kt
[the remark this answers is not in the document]
anchita-g: Makes sense. Done!

engine/src/main/java/com/google/android/fhir/db/impl/dao/ResourceDao.kt
anchita-g: The changes I added make sure that the uuid for the resource remains the same
We need the resource uuid to remain the same for its entire life it's on the device until it is deleted

aditya-07 (changes requested):

engine/src/main/java/com/google/android/fhir/db/impl/dao/ResourceDao.kt
```diff
@@ -92,7 +94,7 @@ internal abstract class ResourceDao {
       ?: throw ResourceNotFoundException(resource.resourceType.name, resource.id)
   }
 
-  open suspend fun insertAllRemote(resources: List<Resource>): List<String> {
+  open suspend fun insertAllRemote(resources: List<Resource>): List<UUID> {
```
MJ1998, line 97: Will the return List\<UUID\> be used ?
I guess List\<String\> was also not being used.

engine/src/main/java/com/google/android/fhir/db/impl/entities/LocalChangeEntity.kt
```diff
@@ -55,11 +56,18 @@ import java.time.Instant
  *
  * ] For resource that is fully synced with server this table should not have any rows.
  */
-@Entity(indices = [Index(value = ["resourceType", "resourceId"])])
+@Entity(
+  indices =
+    [
+      Index(value = ["resourceType", "resourceId"]),
+      Index(value = ["resourceType", "resourceUuid"]),
```
MJ1998, line 63: Isnt Index on resourceUuid enough ?
Why do we need to pair it with resourceType ?

MJ1998 (changes requested):
I don't understand why the alternative solution won't work.

In case of locally deleted resource, we can remove resourceId and still fetch the mapped ResourceEntity via LocalChangeEntity.resourceUuid.

aditya-07:
> One alternative in the first scenario was to just keep the resourceUuid and remove resourceId. However, when LocalChanges are uploaded to the server, we would need to look up the ResourceEntity using the resourceUuid in the LocalChangeEntity to obtain the server recognisable identifier of the resource. This would be a problem in case of deleted resources, where in for any LocalChange of a resource that has been deleted, we would not be able to find the corresponding ResourceEntity. Hence, keeping resourceId is also crucial

For the DELETE resource event, engine removes its entry from `ResourceEntity` and adds a corresponding DELETE entry into the `LocalChangeEntity`. Hence, to sync the change to sever, `LocalChangeEntity` should have the resourceID so that it can be uploaded properly.

aditya-07:
> I don't understand why the alternative solution won't work.
> 
> In case of locally deleted resource, we can remove resourceId and still fetch the mapped ResourceEntity via LocalChangeEntity.resourceUuid.


One alternative in the first scenario was to just keep the resourceUuid and remove resourceId. However, when LocalChanges are uploaded to the server, we would need to look up the ResourceEntity using the resourceUuid in the LocalChangeEntity to obtain the server recognisable identifier of the resource. This would be a problem in case of deleted resources, where in for any LocalChange of a resource that has been deleted, we would not be able to find the corresponding ResourceEntity. Hence, keeping resourceId is also crucial

For the DELETE resource event, engine removes its entry from ResourceEntity and adds a corresponding DELETE entry into the LocalChangeEntity. Hence, to sync the change to sever, LocalChangeEntity should have the resourceID so that it can be uploaded properly.

MJ1998 (approved):
LGTM

aditya-07 (approved):
LGTM!

jingtang10:
@anchita-g have you tested the db migration on a real device? thanks!

anchita-g:
> @anchita-g have you tested the db migration on a real device? thanks!

Hello, yes I was able to test the db migration on a real device using the demo application. It works seamlessly and I was able to see the new column in the `App Inspection` database explorer

PR #1911, opened 2023-03-14
Repeated Group: Fix Definition based extraction for multiple answers

nsabale7:
**IMPORTANT: All PRs must be linked to an issue (except for extremely trivial and straightforward changes).**

Fixes #1794 

**Description**
Iterating only questionnaire response items and finding the corresponding questionnaire item, as repeated groups answers create multiple response items with same linkId.

**Alternative(s) considered**
Another alternative can be to create a `mapOf<linkId, QuestionnaireItemComponent>`  so that it can be used to find by `linkId`.
`val currentQuestionnaireItem =
        questionnaireItemList.find { it.linkId == currentQuestionnaireResponseItem.linkId }`  can be replaced by map.

**Type**
Choose one: (Bug fix | Feature | Documentation | Testing | Code health | Builds | Releases | Other)
Bug fix

**Screenshots (if applicable)**

**Checklist**
- [x] I have read and acknowledged the [Code of conduct](https://github.com/google/android-fhir/blob/master/docs/CODE_OF_CONDUCT.md).
- [x] I have read the [Contributing](https://github.com/google/android-fhir/wiki/Contributing) page.
- [x] I have signed the Google [Individual CLA](https://cla.developers.google.com/about/google-individual), or I am covered by my company's [Corporate CLA](https://cla.developers.google.com/about/google-corporate ).
- [x] I have discussed my proposed solution with code owners in the linked issue(s) and we have agreed upon the general approach.
- [x] I have run `./gradlew spotlessApply` and `./gradlew spotlessCheck` to check my code follows the style guide of this project.
- [x] I have run `./gradlew check` and `./gradlew connectedCheck` to test my changes locally.
- [x] I have built and run the demo app(s) to verify my change fixes the issue and/or does not break the demo app(s).

PR #2216, opened 2023-09-29
UnpackReapted Groups Fix for single group item

### LINKED ISSUE #2162: Mapping Failed For Questionnaire Response with repeated group Items
**Describe the bug**
I have a questionnaire with a repeated parent group item with some child items. When I open the questionnaire response for the same questionnaire in review mode and then go to submit the item list in the questionnaire response is empty.

Questionnaire:
```
{
  "item": [
    {
      "type": "group",
      "linkId": "parent-item",
      "text": "Parent Group Item",
      "repeats": true,
      "item": [
        {
          "linkId": "1",
          "type": "string",
          "text": "Personal information",
          "initial": [
            {
              "valueString": "ABC"
            }
          ]
        }
      ]
    }
  ]
}
```

QuestionnaireResponse:
```
{
  "resourceType": "QuestionnaireResponse",
  "item": [
    {
      "linkId": "parent-item",
      "text": "Parent Group Item",
      "item": [
        {
          "linkId": "1",
          "text": "Personal information",
          "answer": [
            {
              "valueString": "ABC"
            }
          ]
        }
      ]
    }
  ]
}
```
**Component**

**To Reproduce**
1. set above questionnaire and questionaireResponse in the DemoQuestionnaireFragment
2. Open the Questionnaire in review mode.
3. Click on Submit

**Expected behavior**
The expected behavior is that it should give the questionnaire response with the items present.

**Demo Video**

https://github.com/google/android-fhir/assets/102521117/98f498b9-d22a-49c1-bb17-15a3127fc028

 - Device: Infinix HOT 11S
 - Android version: 11

**Additional context**

AdityaKBhadragond14:

Fixes #2162

**Description**
Consider a scenario where there is a questionnaire with one questionnaire item whose type is group and repeats is set to true for the same.

```
{
  "item": [
    {
      "type": "group",
      "linkId": "parent-item",
      "text": "Parent Group Item",
      "repeats": true,
      "item": [
        {
          "linkId": "1",
          "type": "string",
          "text": "Personal information",
          "initial": [
            {
              "valueString": "ABC"
            }
          ]
        }
      ]
    }
  ]
}
```

In this case if only group item is answered then questionnaire response returned is with empty item list like following:

```
{
  "resourceType": "QuestionnaireResponse",
  "item": [
    {
      "linkId": "parent-item",
      "text": "Parent Group Item",
      "item": [
        {
          "linkId": "1",
          "text": "Personal information",
          "answer": [
            {
              "valueString": "ABC"
            }
          ]
        }
      ]
    }
  ]
}
```

But if there are more than one group items being answered in the questionnaire then the questionnaire response is proper.
To fix this issue if there are more than one repeated group items in the questionnaire response then do the recursive unpacking logic or else if there is only one group item in the questionnaire response then do not do the unpacking of repeated groups instead return the group item.

**Alternative(s) considered**

**Type**
Bug Fix

**Checklist**

### DIFF
--- a/datacapture/src/main/java/com/google/android/fhir/datacapture/extensions/MoreQuestionnaireResponses.kt
+++ b/datacapture/src/main/java/com/google/android/fhir/datacapture/extensions/MoreQuestionnaireResponses.kt
@@ -100,22 +100,27 @@ private fun unpackRepeatedGroups(
   questionnaireItem: Questionnaire.QuestionnaireItemComponent,
   questionnaireResponseItem: QuestionnaireResponse.QuestionnaireResponseItemComponent,
 ): List<QuestionnaireResponse.QuestionnaireResponseItemComponent> {
-  questionnaireResponseItem.item =
-    unpackRepeatedGroups(questionnaireItem.item, questionnaireResponseItem.item)
-  questionnaireResponseItem.answer.forEach {
-    it.item = unpackRepeatedGroups(questionnaireItem.item, it.item)
-  }
-  return if (
-    questionnaireItem.type == Questionnaire.QuestionnaireItemType.GROUP && questionnaireItem.repeats
-  ) {
-    questionnaireResponseItem.answer.map {
-      QuestionnaireResponse.QuestionnaireResponseItemComponent().apply {
-        linkId = questionnaireItem.linkId
-        text = questionnaireItem.localizedTextSpanned?.toString()
-        item = it.item
+  if (questionnaireResponseItem.hasAnswer()) {
+    questionnaireResponseItem.item =
+      unpackRepeatedGroups(questionnaireItem.item, questionnaireResponseItem.item)
+    questionnaireResponseItem.answer.forEach {
+      it.item = unpackRepeatedGroups(questionnaireItem.item, it.item)
+    }
+    return if (
+      questionnaireItem.type == Questionnaire.QuestionnaireItemType.GROUP &&
+        questionnaireItem.repeats
+    ) {
+      questionnaireResponseItem.answer.map {
+        QuestionnaireResponse.QuestionnaireResponseItemComponent().apply {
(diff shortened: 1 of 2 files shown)


datacapture/src/main/java/com/google/android/fhir/datacapture/extensions/MoreQuestionnaireResponses.kt
jingtang10, line 1: please add a test case in `MoreQuestionnaireResponsesTest.kt`

jingtang10 (changes requested):

datacapture/src/main/java/com/google/android/fhir/datacapture/extensions/MoreQuestionnaireResponses.kt
AdityaKBhadragond14: I have added the test case please review and let me know.

jingtang10 (commented):
after looking at the test case, i'm not entirely sure this is the right fix. the example you give is already unpacked, why do you think the unpack function is the problem?

in the questionnaire view model class, we pack repeated groups under the same questionnaire response item, and use each answer to host nested questions in each repeated group. have you checked that's been handled correctly?

AdityaKBhadragond14:
> after looking at the test case, i'm not entirely sure this is the right fix. the example you give is already unpacked, why do you think the unpack function is the problem?
> 
> in the questionnaire view model class, we pack repeated groups under the same questionnaire response item, and use each answer to host nested questions in each repeated group. have you checked that's been handled correctly?

The issue was when there is only group item in the questionnaire response. In this case the group item won't be under answer of the questionnaire response due to this the unpack function was returning the empty questionnaire response.

joiskash:
We can close this PR. #1911 has the fix for this.

jingtang10:
ok let's close this. thanks both!

PR #2217, opened 2023-09-29
Resource consolidation for resources created with POST

### LINKED ISSUE #760: Create new resources on the server using POST
We need to support [create](https://hl7.org/fhir/http.html#create) using POST in the sync API.

<s>
**Describe the bug**
Locally created Resources are assigned ID at creation/before saving to the local db. Once this Resource is up-synced, the server assigns a new ID to the resource and disregards the assigned ID  (in case of HTTP-POST). The created object is returned  in the POST response. 
Since the lD for local & remote copy are different now, need to update all the local id references with the remote id.

**Component**

**To Reproduce**

**Expected behavior**

**Screenshots**

**Additional context**

</s>

anchita-g:

One of the PRs to fix #760 

**Description**
- Added resource consolidation for cases where resources are created through `POST`.  We would need to consolidate the IDs of the newly created resources in any `LocalChange` referring to these created resources as well in any `Resource` referring to these created resources.
- Introduced a new entity called `LocalChangeResourceReferenceEntity` which maintains a mapping between a `LocalChange` and any references to a resource the contents of the `LocalChange` contains. This would help us in easily looking up all the `LocalChange`s that need to be updated when a referred resource is assigned a new ID by the server. Similarly, we can use the mapped `LocalChange` to get al the resources that refer to the said resource. These resources as well as their respective references would need to be updated as well.
- Added a migration to create the new mapping from the existing content of the `LocalChange` to the "references" present in the payload of the `LocalChange` along with a test case to test the migration.

**Alternative(s) considered**
 - An alternative to update the references in the `Resource` or the `LocalChange` of the resource referring to the created resource was to use the FHIR path that is mentioned in the `LocalChangeResourceReferenceEntity`. However, in cases where there are lists of references, FHIR path does not indicate the index in the list at which the reference exists. In such cases we would have to resolve to updating the reference value by either replacing string value by regex match or use JSON update strategy. Similar case holds for `LocalChange` where in for the `UPDATE` type `LocalChange`, updating the reference using FHIR path would not be possible.

**Type**
Choose one: Feature

**Checklist**


engine/src/main/java/com/google/android/fhir/sync/upload/consolidator/DefaultResourceConsolidator.kt
jingtang10, line 1: i think this consolidator might belong better in engine, in that we want to keep db access code behind engine APIs (i.e. no db api call directly as you do in the implementation of this interface).

but i think taht's something we can work on in the follow up pr that stitches everything together.

engine/src/main/java/com/google/android/fhir/sync/upload/consolidator/PostPerResourceUrlRequestConsolidator.kt
```diff
@@ -0,0 +1,198 @@
...
+import com.google.android.fhir.logicalId
+import com.google.android.fhir.sync.upload.request.UrlUploadRequest
+import org.hl7.fhir.r4.model.Resource
+import org.hl7.fhir.r4.model.ResourceType
+import org.json.JSONArray
+import org.json.JSONObject
+
+/**
+ * Implementation of [ResourceConsolidator] for POST mode of creation of resources. The
+ * [LocalChange]s are squashed at a resource level and each resource request is uploaded
```
jingtang10, line 35: i think the squashing is irrelevant here.

engine/src/main/java/com/google/android/fhir/sync/upload/consolidator/ResourceConsolidator.kt
```diff
@@ -0,0 +1,36 @@
...
+ * together with other components in the upload package to fulfill a specific upload strategy. After
+ * a resource is uploaded to a remote FHIR server and a response is returned, we need to consolidate
+ * any changes in the database, Examples of this would be, updating the lastUpdated timestamp field,
+ * or deleting the local change from the database, or updating the resource IDs and payloads to
+ * correspond with the server’s feedback.
+ */
+internal fun interface ResourceConsolidator {
+
+  /** Consolidates the local change token with the provided response from the FHIR server. */
+  suspend fun consolidate(localChangeToken: LocalChangeToken, response: Resource)
```
jingtang10, line 35: can we change this to taking local changes instead of the token?

@omarismail94 fyi

engine/src/main/java/com/google/android/fhir/db/impl/DatabaseImpl.kt
```diff
@@ -214,6 +221,48 @@ internal class DatabaseImpl(
     }
   }
 
+  override suspend fun getAllResourcesReferringToResourceWithPath(
+    resourceType: ResourceType,
+    resourceId: String,
+  ): List<ReferringResource> {
+    val searchForReferenceValue = "$resourceType/$resourceId"
```
jingtang10, line 228: can you take a look at the fhir spec, and figure out if reference fields may not be part of a search parameter.

we index all search parameters defined in fhir. so if something isn't defined as a search parameter, we won't reference it, and we won't update the index in those cases.

engine/src/androidTest/java/com/google/android/fhir/db/impl/DatabaseImplTest.kt
jingtang10, line 1: new db apis and changes need to have tests in this class

engine/src/main/java/com/google/android/fhir/sync/upload/consolidator/PostPerResourceUrlRequestConsolidator.kt
anchita-g: `Squashing` i.e. considering changes at a resource level is important because `ReferenceIndex` only stores the references present in the latest version of a resource. 
Consider a resource A which refers to a resource B  (hence one or few of the `LocalChange` of the resource A will contain the reference to resource B). The resource A is now modified and the reference to the resource B in this resource was removed. This will remove the index for resource B in `ReferenceIndex`. Hence we would not be able to discover resource A when we want to update the resources referring to resource B. Not updating the references in resource A's `LocalChange` would only be okay if we squash changes at a resource level and then upload the changes. Hence this consolidator would only work for scenarios where changes are squashed at resource level.

I do acknowledge that we need to handle the other case as well where in we do have the handle resources which referred to a resource in one of their older versions and update the `LocalChange` for those resources.

anchita-g: Yes, in fact I have removed this specific case of consolidator. Incorporated everything into the `DefaultResourceConsolidator`

engine/src/main/java/com/google/android/fhir/sync/upload/consolidator/ResourceConsolidator.kt
anchita-g: Done

engine/src/main/java/com/google/android/fhir/db/impl/DatabaseImpl.kt
anchita-g: Your intuition was correct. I compared the search params for some resources and found out that the search params do not cover all the references in a resource .Hence I have introduced a new entity `LocalChangeResourceReferenceEntity`  which maintains a mapping between the `LocalChange` and the resources it references. This mapping can be further used to fetch all the parent resource of the `LocalChange` to get all resource which contain the reference to a resource.

engine/src/androidTest/java/com/google/android/fhir/db/impl/DatabaseImplTest.kt
anchita-g: Added test cases for the new APIs. Also added a new test class for `LocalChangeDao`

engine/src/main/java/com/google/android/fhir/db/impl/DatabaseImpl.kt
```diff
@@ -247,6 +255,42 @@ internal class DatabaseImpl(
     localChangeDao.discardLocalChanges(resources)
   }
 
+  override suspend fun updateResourceAndId(currentResourceId: String, updatedResource: Resource) {
```
omarismail94, line 258: can we break this function into smaller, more focused functions, making the code clearer and more maintainable? something like this?

```
override suspend fun updateResourceAndId(currentResourceId: String, updatedResource: Resource) {
    val currentResourceEntity = selectEntity(updatedResource.resourceType, currentResourceId)
    val oldResource = iParser.parseResource(currentResourceEntity.serializedResource) as Resource

    db.withTransaction {
        updateResourceEntity(currentResourceEntity, updatedResource)
        val resourcesToUpdate = updateLocalChanges(oldResource, updatedResource, currentResourceEntity.resourceUuid)
        updateReferringResources(resourcesToUpdate, oldResource, updatedResource)
    }
}

private suspend fun updateResourceEntity(currentResourceEntity: ResourceEntity, updatedResource: Resource) {
    resourceDao.updateResourceWithUuid(currentResourceEntity.resourceUuid, updatedResource)
}

private suspend fun updateLocalChanges(oldResource: Resource, updatedResource: Resource, resourceUuid: String): List<String> {
    return localChangeDao.updateResourceId(resourceUuid, oldResource, updatedResource)
}

private suspend fun updateReferringResources(uuids: List<String>, oldResource: Resource, updatedResource: Resource) {
    val oldReferenceValue = "${oldResource.resourceType.name}/${oldResource.logicalId}"
    val updatedReferenceValue = "${updatedResource.resourceType.name}/${updatedResource.logicalId}"

    uuids.forEach { resourceUuid ->
        resourceDao.getResourceEntity(resourceUuid)?.let {
            val referringResource = iParser.parseResource(it.serializedResource) as Resource
            val updatedReferringResource = addUpdatedReferenceToResource(iParser, referringResource, oldReferenceValue, updatedReferenceValue)
            resourceDao.updateResourceWithUuid(resourceUuid, updatedReferringResource)
        }
    }
}
```

engine/src/main/java/com/google/android/fhir/db/impl/dao/LocalChangeDao.kt
```diff
@@ -198,14 +281,147 @@ internal abstract class LocalChangeDao {
...
+   * [LocalChangeEntity.resourceId]. Looks for [LocalChangeEntity] which refer to the updated
+   * resource through [LocalChangeResourceReferenceEntity]. For each [LocalChangeEntity] which
+   * contains reference to the updated resource in its payload, we update the payload with the
+   * reference and also update the corresponding [LocalChangeResourceReferenceEntity]. We delete the
+   * original [LocalChangeEntity] and create a new one with new
+   * [LocalChangeResourceReferenceEntity]s in its place. This method returns a list of the
+   * [ResourceEntity.resourceUuid] for all the resources whose [LocalChange] contained references to
+   * the oldResource
+   */
+  suspend fun updateResourceId(
```
omarismail94, line 343: the function seems to be doing more than just updating the resource id. Can we name it to reflect what is is doing? e.g. `updateResourceAndRelatedChanges` or `updateResourceAndReferencesInLocalChanges` or something to that effect

engine/src/main/java/com/google/android/fhir/db/impl/dao/LocalChangeDao.kt
```diff
@@ -198,14 +281,147 @@ internal abstract class LocalChangeDao {
...
+    resourceUuid: UUID,
+    oldResource: Resource,
+    updatedResource: Resource,
+  ): List<UUID> {
+    // update the resource ID in the local change entity
+    val localChanges = getLocalChanges(resourceUuid)
+    localChanges
+      .map { localChangeEntity -> localChangeEntity.copy(resourceId = updatedResource.logicalId) }
+      // Add LocalChangeEntity with replace strategy, the references need not be updated
+      .forEach { addLocalChange(it) }
```
omarismail94, line 353: can we wrap this in a function called:

 `updateLocalChangeEntities(resourceUuid, updatedResource)`

to break this function into smaller pieces and the code below it to something like:

`updateLocalChangesReferringToResource`

engine/src/main/java/com/google/android/fhir/db/impl/dao/LocalChangeDao.kt
```diff
@@ -198,14 +281,147 @@ internal abstract class LocalChangeDao {
...
+
+    localChangeIds.forEach { localChangeId ->
+      val existingLocalChangeEntity = localChangesWithReferences[localChangeId]!!
+      val updatedLocalChangeEntity =
+        replaceReferencesInLocalChangeEntity(
+            localChange = existingLocalChangeEntity,
+            oldReference = oldReferenceValue,
+            updatedReference = updatedReferenceValue,
+          )
+          .copy(id = 0)
```
omarismail94, line 370: can we have the magic number `0` refered to as a named constants for clarity?

engine/src/main/java/com/google/android/fhir/db/impl/JsonUtils.kt
```diff
@@ -0,0 +1,130 @@
...
+  resource: Resource,
+  outdatedReference: String,
+  updatedReference: String,
+): Resource {
+  val resourceJsonObject = JSONObject(iParser.encodeResourceToString(resource))
+  val updatedResource = replaceJsonValue(resourceJsonObject, outdatedReference, updatedReference)
+  return iParser.parseResource(updatedResource.toString()) as Resource
+}
+
+fun replaceJsonValue(
```
omarismail94, line 35: can we reduce code duplication by consolidating the logic for handling JSONObject and JSONArray into a single function using Kotlin's type checking? something like:

```
fun replaceJsonValue(
  jsonElement: Any,
  currentValue: String,
  newValue: String,
): Any {
  when (jsonElement) {
    is JSONObject -> {
      val iterator: Iterator<*> = jsonElement.keys()
      while (iterator.hasNext()) {
        val key = iterator.next() as String
        when (val element = jsonElement[key]) {
          is String -> if (element == currentValue) jsonElement.put(key, newValue)
          is JSONObject -> replaceJsonValue(element, currentValue, newValue)
          is JSONArray -> replaceJsonValue(element, currentValue, newValue)
        }
      }
    }
    is JSONArray -> {
      for (i in 0 until jsonElement.length()) {
        val element = jsonElement[i]
        when (element) {
          is String -> if (element == currentValue) jsonElement.put(i, newValue)
          is JSONObject, is JSONArray -> replaceJsonValue(element, currentValue, newValue)
        }
      }
    }
  }
  return jsonElement
}
```

engine/src/main/java/com/google/android/fhir/db/impl/dao/LocalChangeDao.kt
```diff
@@ -123,9 +167,38 @@ internal abstract class LocalChangeDao {
...
   }
 
+  private fun extractResourceReferences(resource: Resource) =
+    fhirTerser.getAllResourceReferences(resource).toSet()
+
+  /**
+   * Extract the difference in the [ResourceReferenceInfo] by getting all the references from both
+   * the versions of the resource and then finding out the difference in the two sets.
+   */
+  private fun extractReferencesDiff(
```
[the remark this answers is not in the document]
anchita-g: In case any new reference is added in the Resource2, we need to add that.
In case any reference was removed from the Resource1, we need to add that as well.

engine/src/main/java/com/google/android/fhir/db/impl/JsonUtils.kt
anchita-g: The function signature conveys that any type of `jsonElement` will be acceptable. With the function definitions I wanted to restrict the type of `jsonElement` that is acceptable.

engine/src/main/java/com/google/android/fhir/sync/upload/consolidator/DefaultResourceConsolidator.kt
```diff
@@ -100,6 +92,23 @@ internal class DefaultResourceConsolidator(private val database: Database) : Res
     }
   }
 
+  private suspend fun updateVersionIdAndLastUpdated(
+    localChanges: List<LocalChange>,
+    resource: Resource,
+  ) {
+    if (localChanges.first().resourceId != resource.logicalId) {
```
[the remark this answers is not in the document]
anchita-g: It could be multiple changes. 
The assumption is that either it is one change or multiple changes which belong to the same resource.
Maybe I should add this explicitly as Kdoc here.

engine/src/main/java/com/google/android/fhir/sync/upload/consolidator/DefaultResourceConsolidator.kt
```diff
@@ -60,6 +50,8 @@ internal class DefaultResourceConsolidator(private val database: Database) : Res
     }
 
   private suspend fun updateVersionIdAndLastUpdated(bundle: Bundle) {
+    // TODO: Support POST in Bundle transactions. Assumption is that only PUT operations are used in
```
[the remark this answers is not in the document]
anchita-g: We will not expose the relevant `UploadStrategy` till we support this. Would that suffice?

engine/src/main/java/com/google/android/fhir/db/ResourceNotFoundException.kt
```diff
@@ -16,10 +16,13 @@
 
 package com.google.android.fhir.db
 
+import java.util.UUID
+
 /** Thrown to indicate that the requested resource is not found. */
 class ResourceNotFoundException : Exception {
```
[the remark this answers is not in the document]
anchita-g: I am not sure if builder will let me define the message based on the constructor arguments of the Exception. I wonder if deciding the message would become more complicated with the Builder pattern

engine/src/main/java/com/google/android/fhir/db/Database.kt
```diff
@@ -114,6 +115,18 @@ internal interface Database {
...
   suspend fun deleteUpdates(resources: List<Resource>)
 
+  /**
+   * Updates the [ResourceEntity.serializedResource] and [ResourceEntity.resourceId] corresponding
+   * to the updatedResource. Updates all the [LocalChangeEntity] for this updated resource as well
+   * as all the [LocalChangeEntity] referring to this resource in their [LocalChangeEntity.payload]
+   * Updates the [ResourceEntity.serializedResource] for all the resources which refer to this
+   * updated resource.
+   */
+  suspend fun updateResourceAndId(
```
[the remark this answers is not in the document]
anchita-g: Done. Thanks!!

engine/src/main/java/com/google/android/fhir/db/impl/DatabaseImpl.kt
```diff
@@ -247,6 +255,42 @@ internal class DatabaseImpl(
...
+    val currentResourceEntity = selectEntity(updatedResource.resourceType, currentResourceId)
+    val oldResource = iParser.parseResource(currentResourceEntity.serializedResource) as Resource
+    val resourceUuid = currentResourceEntity.resourceUuid
+    val oldReferenceValue = "${oldResource.resourceType.name}/${oldResource.logicalId}"
+    val updatedReferenceValue = "${updatedResource.resourceType.name}/${updatedResource.logicalId}"
+    db.withTransaction {
+      // update the resource entity
```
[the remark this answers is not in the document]
anchita-g: Done. Thanks!!

engine/src/main/java/com/google/android/fhir/db/impl/JsonUtils.kt
```diff
@@ -0,0 +1,130 @@
...
+ */
+
+package com.google.android.fhir.db.impl
+
+import ca.uhn.fhir.parser.IParser
+import org.hl7.fhir.r4.model.Resource
+import org.json.JSONArray
+import org.json.JSONObject
+
+fun addUpdatedReferenceToResource(
```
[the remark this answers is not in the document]
anchita-g: Made the functions internal. Thanks!!

engine/src/main/java/com/google/android/fhir/db/impl/dao/LocalChangeDao.kt
```diff
@@ -156,14 +229,24 @@ internal abstract class LocalChangeDao {
...
+        WHERE LocalChangeEntity.id IN (:ids) 
+        ORDER BY timestamp ASC""",
+  )
+  abstract suspend fun getLocalChanges(ids: List<Long>): List<LocalChangeEntity>
+
+  @Query(
+    """
+        SELECT COUNT(*)
+        FROM LocalChangeEntity 
+        ORDER BY timestamp ASC
```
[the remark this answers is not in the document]
anchita-g: Removed the "ORDER BY" clause. Thanks for pointing this out!!

engine/src/main/java/com/google/android/fhir/sync/upload/consolidator/DefaultResourceConsolidator.kt
```diff
@@ -100,6 +92,23 @@ internal class DefaultResourceConsolidator(private val database: Database) : Res
     }
   }
 
+  private suspend fun updateVersionIdAndLastUpdated(
```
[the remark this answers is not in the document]
anchita-g: Renamed this. Thanks!

engine/src/main/java/com/google/android/fhir/sync/upload/consolidator/DefaultResourceConsolidator.kt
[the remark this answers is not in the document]
anchita-g: Added comment to this function

engine/src/main/java/com/google/android/fhir/db/impl/DatabaseImpl.kt
anchita-g: Done. Thanks for the suggestion

engine/src/main/java/com/google/android/fhir/db/impl/dao/LocalChangeDao.kt
anchita-g: Done, thanks for the suggestion!

engine/src/main/java/com/google/android/fhir/db/impl/dao/LocalChangeDao.kt
anchita-g: Done, thanks for the suggestion!

engine/src/main/java/com/google/android/fhir/sync/upload/consolidator/DefaultResourceConsolidator.kt
jingtang10: this will be in a follow-up.

engine/src/main/java/com/google/android/fhir/db/ResourceNotFoundException.kt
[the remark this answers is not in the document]
jingtang10: i think builder is an overkill for an exception class this small.

i personally wouldn't mind having one constructor with optional uuid, id, type, cause etc. but i take the point that we don't want caller to create this exception without any of them (and it would be weird to throw an exception whilst creating an exception, so better to enforce this at compile time).

up to you though! this really isn't the most critical thing we'll solve today :)

engine/src/main/java/com/google/android/fhir/sync/upload/consolidator/DefaultResourceConsolidator.kt
```diff
@@ -100,6 +95,17 @@ internal class DefaultResourceConsolidator(private val database: Database) : Res
     }
   }
 
+  /** The function expects [List<[LocalChange]>] which belong only to the resource. */
+  private suspend fun consolidateResourceAndItsReferences(
```
jingtang10, line 99: add more doc to explain what the resource is and what the local changes are in relation to it.

also, i think the decision to update id may come from the caller of the resource consolidator. feels strange that the resource consolidator decides that from checking the id of the first local change.

jingtang10 (dismissed):

engine/src/main/java/com/google/android/fhir/sync/upload/consolidator/DefaultResourceConsolidator.kt
anchita-g: This feels like an issue which is bigger than just the Resource Consolidation for POST.  I am reverting my changes here and shall address this in the PR where we stitch up all the Upload components.

engine/src/main/java/com/google/android/fhir/db/impl/DatabaseImpl.kt
```diff
@@ -247,6 +255,67 @@ internal class DatabaseImpl(
...
+        oldResource = oldResource,
+        updatedResource = updatedResource,
+      )
+    }
+  }
+
+  private suspend fun updateResourceEntity(resourceUuid: UUID, updatedResource: Resource) =
+    resourceDao.updateResourceWithUuid(resourceUuid, updatedResource)
+
+  private suspend fun updateLocalChangeResourceIdAndReferences(
```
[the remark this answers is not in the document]
anchita-g: Added documentation

engine/src/main/java/com/google/android/fhir/db/ResourceNotFoundException.kt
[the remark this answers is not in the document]
anchita-g: Leaving this as is as of now. We can address this later

engine/src/androidTest/java/com/google/android/fhir/db/impl/DatabaseImplTest.kt
```diff
@@ -3532,6 +3533,305 @@ class DatabaseImplTest {
       )
   }
 
+  @Test
+  fun updateResourceAndId_insertPatientAndReferringResource_shouldUpdateReferencesAndUpdateResourceId() =
```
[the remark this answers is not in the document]
anchita-g: Have broken the test cases as mentioned.

engine/src/main/java/com/google/android/fhir/db/impl/dao/LocalChangeDao.kt
```diff
@@ -123,9 +180,52 @@ internal abstract class LocalChangeDao {
...
+   * 2) A new reference is added to the resource1. This implies that the reference is present in
+   *    resource2 and not in resource1. Such differences in resource references can be extracted by
+   *    subtracting the set of references in the resource1 from the set of references in the
+   *    resource2
+   *
+   * Combining the above two types of differences would give the entire set of difference in the two
+   * versions of the resource.
+   *
+   * This method is useful to extract differences for UPDATE kind of [LocalChange]
+   */
```
omarismail94, line 207: The kdoc seems verbose and repetitive in some places. To enhance clarity and conciseness, I would recommend the following improvements:
```suggestion
/**
 * Extracts the difference in [ResourceReferenceInfo] between two versions of a resource.
 *
 * A difference between two resource versions, in terms of references, can manifest in two ways:
 * 1) A reference is present in `resource1` but not in `resource2`.
 * 2) A reference is present in `resource2` but not in `resource1`.
 *
 * This method computes both types of differences and returns the combined set of differences. 
 * It is particularly useful for extracting differences during an UPDATE operation on a [LocalChange].
 *
 * @param resource1 The first version of the resource.
 * @param resource2 The second version of the resource.
 * @return A set of [ResourceReferenceInfo] indicating the differences in references between the two resource versions.
 */
```

omarismail94 (commented):
Just two minor comments!

omarismail94:
@anchita-g can you merge the master branch and resolve any conflicts? can approve once that is done and the comments are addressed

engine/src/main/java/com/google/android/fhir/db/impl/dao/LocalChangeDao.kt
```diff
@@ -40,40 +47,80 @@ import timber.log.Timber
 
 /**
  * Dao for local changes made to a resource. One row in LocalChangeEntity corresponds to one change
- * e.g. an INSERT or UPDATE. The UPDATES (diffs) are stored as RFC 6902 JSON patches. When a
- * resource needs to be synced, all corresponding LocalChanges are 'squashed' to create a a single
- * LocalChangeEntity to sync with the server.
+ * e.g. an INSERT or UPDATE. The UPDATES (diffs) are stored as RFC 6902 JSON patches.
  */
 @Dao
+@VisibleForTesting
```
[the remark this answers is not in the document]
anchita-g: Right, I removed it. Thanks

omarismail94 (approved):

PR #2234, opened 2023-10-06
Fix broken dependabot.yaml syntax (fixes #2195)

### LINKED ISSUE #2195: Set up Dependabot

vorburger:
Fixes #2195

### DIFF
--- a/.github/dependabot.yaml
+++ b/.github/dependabot.yaml
@@ -1,10 +1,10 @@
 # https://docs.github.com/github/administering-a-repository/configuration-options-for-dependency-updates
-
 version: 2
+
+# Gotta Catch 'Em All! (i.e. don't wait and propose them only "trickled")
+open-pull-requests-limit: 99
+
 updates:
-    # Gotta Catch 'Em All! (i.e. don't wait and propose them only "trickled")
-    open-pull-requests-limit: 99
-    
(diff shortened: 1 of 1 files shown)


vorburger:
> Kokoro: Build and Device Tests — Kokoro build failed. Check logs for details.

#2235 created re. this; please ignore when reviewing this change (it is totally safe to merge this, it won't break tests).

jingtang10 (approved):

PR #2262, opened 2023-10-12
Fix issue with duplicated IDs inside the Knowledge Manager

vitorpamplona:
Search by ID in the KnowledgeManager was used to return only the first reference. If the ID was duplicated in the DB, it would not filter by type and return the wrong one. This fixes it while also enforcing the ID to be a complete word in the URL

**Type**
Bug fix

**Checklist**


omarismail94 (approved):

jingtang10 (changes requested):
1) please add tests.
2) is this definitely the right fix? trying to match part of the url is still weird to me. if there is an id that is a substring of another id, the result isn't guaranteed... i have been a little worried about this actually regarding id and name... seems really confusing in our code.

jingtang10:
let's not merge this as is - i think we still have issues to address here.

jingtang10:
closing in favour of the new pr. thx vitor!

PR #2264, opened 2023-10-14
Overrides Ids when installing an IG

vitorpamplona:
**Description**
A draft solution to make sure the searchById in the Knowledge Manager works without searching by part of the URL. 

This PR: 
1. Overrides `id` when installing Knowledge Artifacts
2. Updates files to match `id`
3. Updates test case framework to not rely in `id`, because they will change 
4. Updates FhirOperator interface to offer a way to not use the internal `id` when selecting a PlanDefinition
5. Updates test cases to match the new output `id`s 

**Type**
Bug fix

**Checklist**

### DIFF
--- a/knowledge/src/main/java/com/google/android/fhir/knowledge/KnowledgeManager.kt
+++ b/knowledge/src/main/java/com/google/android/fhir/knowledge/KnowledgeManager.kt
@@ -28,9 +28,11 @@ import com.google.android.fhir.knowledge.npm.NpmPackageDownloader
 import com.google.android.fhir.knowledge.npm.OkHttpNpmPackageDownloader
 import java.io.File
 import java.io.FileInputStream
+import java.io.FileOutputStream
 import kotlinx.coroutines.Dispatchers
 import kotlinx.coroutines.withContext
 import org.hl7.fhir.instance.model.api.IBaseResource
+import org.hl7.fhir.r4.model.IdType
 import org.hl7.fhir.r4.model.MetadataResource
 import org.hl7.fhir.r4.model.Resource
 import org.hl7.fhir.r4.model.ResourceType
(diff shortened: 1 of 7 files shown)


knowledge/src/main/java/com/google/android/fhir/knowledge/KnowledgeManager.kt
```diff
@@ -95,7 +97,13 @@ internal constructor(
...
         val resource = jsonParser.parseResource(FileInputStream(file))
         if (resource is Resource) {
-          importResource(igId, resource, file)
+          val newId = importResource(igId, resource, file)
+          resource.setId(IdType(resource.resourceType.name, newId))
+
+          // Overrides the Id in the file
+          FileOutputStream(file).use {
+            it.write(jsonParser.encodeResourceToString(resource).toByteArray())
+          }
```
jingtang10, line 106: this is actually changing something quite fundamental to the knowledge manager library - we're writing the file now and the resource files are no longer immutable!

i think this is inevitable. because there's no guarantee that we'll never import resources of the same id from different terminology servers.

this comes with questions regarding things like android permissions and performance. BUT i think we can probably go ahead with this and address these issues in a future release.

an alternative solution is to always override the logical id before returning to the clinical reasoning module - have you considered this?

knowledge/src/main/java/com/google/android/fhir/knowledge/KnowledgeManager.kt
```diff
@@ -154,11 +162,19 @@ internal constructor(
         }
       }
     when (resource) {
-      is Resource -> importResource(igId, resource, file)
+      is Resource -> {
```
jingtang10, line 165: while we're at this, can you rename this function (line 155) from `importFile` to `indexResourceFile`? we're not copying this file so i want to make our function naming a bit better.

jingtang10 (changes requested):

workflow/src/main/java/com/google/android/fhir/workflow/FhirOperator.kt
```diff
@@ -189,6 +189,11 @@ internal constructor(
     return generateCarePlan(planDefinitionId, patientId, encounterId = null)
   }
 
+  @WorkerThread
```
jingtang10, line 192: please add a comment to mark the old api as deprecated

jingtang10 (approved):

Issue #2302, opened 2023-10-25
Declare dependencies differently, so that Dependabot "sees" them (and "group" dependencies which Dependabot already sees)


vorburger:
The goal of this issue is to improve how this project declares its dependencies in Gradle.

The primary motivation for this issue is that the Dependabot which I activated in #2195 so far only [raises a limited number of "version bump PRs"](https://github.com/google/android-fhir/pulls?q=author%3Aapp%2Fdependabot+), because as-is Dependabot only sees Gradle plugins, and the few Jackson dependencies declared in [`settings.gradle.kts`](https://github.com/google/android-fhir/blob/master/settings.gradle.kts) and in `common/build.gradle.kts` and `buildSrc/build.gradle.kts`, as well as GitHub Actions - but it misses all the (more interesting) dependencies in [`buildSrc/src/main/kotlin/Dependencies.kt`](https://github.com/google/android-fhir/blob/master/buildSrc/src/main/kotlin/Dependencies.kt) (understandably, because that's "non-standard / custom Kt Gradle code").

As a "side effect" of this work, I'm also hoping to be able to better declaratively define "groups" of dependencies which "belong together", so that Dependabot automatically (instead of me manually) raises #2301 instead of its individual (and incomplete) #2283 and #2285.

vorburger:
I'll (propose to) introduce a `gradle/libs.versions.toml`

see https://docs.gradle.org/current/userguide/platforms.html

and https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/about-dependabot-version-updates#gradle

vorburger:
see also https://developer.android.com/build/migrate-to-catalogs

vorburger:
With #2306 merged, the next step here (from my side) is to wait and see if Dependabot will automagically propose a bump for Glide...

vorburger:
> wait and see if Dependabot will automagically propose a bump for Glide...

Reminder to myself, I forgot about this: Because of how https://github.com/google/android-fhir/blob/master/.github/dependabot.yaml is currently configured, any #2306-like work will only lead to new dependency update PRs at `interval: weekly` on `day: monday`... should we make it more frequent?

PR #2306, opened 2023-10-26
Introduce Version Catalog, initially intentionally only for Glide

vorburger:
see #2302 

My expectation is that if this passes CI and gets merged, Dependabot will automagically propose a bump for Glide - let's try and see.

If this works, we should then migrate to this model. Perhaps that's something that could be "crowd sourced" and fanned out, instead of me doing all migrations myself?

PS: There would be an intermediate phase with "mixed" dependency declaration; to me such a "gradual adoption" is a "feature not a bug", with testing each change, and more sensible than a single big bang all in one migration.

### DIFF
--- a/buildSrc/src/main/kotlin/Dependencies.kt
+++ b/buildSrc/src/main/kotlin/Dependencies.kt
@@ -41,10 +41,6 @@ object Dependencies {
       "org.opencds.cqf.fhir:cqf-fhir-utility:${Versions.Cql.clinicalReasoning}"
   }
 
-  object Glide {
-    const val glide = "com.github.bumptech.glide:glide:${Versions.Glide.glide}"
-  }
-
   object HapiFhir {
     const val fhirBase = "ca.uhn.hapi.fhir:hapi-fhir-base:${Versions.hapiFhir}"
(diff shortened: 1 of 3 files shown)


omarismail94:
This is actually VERY nice!! I never knew about this:
https://docs.gradle.org/current/userguide/platforms.html

Superb job!

omarismail94 (approved):

vorburger:
> Superb job!

🙇🏽  but credit where credit is due - I learnt about this after doing research [starting from this comment](https://github.com/google/android-fhir/issues/2195#issuecomment-1743606990)!

> https://docs.gradle.org/current/userguide/platforms.html

see #2302, also https://developer.android.com/build/migrate-to-catalogs, and https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/about-dependabot-version-updates#gradle for why _"My expectation is that if this passes CI and gets merged, Dependabot will automagically propose a bump for Glide - let's try and see."_

jingtang10:
this is lovely. thanks @vorburger!

jingtang10:
`If this works, we should then migrate to this model. Perhaps that's something that could be "crowd sourced" and fanned out, instead of me doing all migrations myself?` -> agreed.

Issue #2294, opened 2023-10-25
Reduce GitHub Action VM Size / Resource Consumption (Cost)


vorburger:
The GitHub "Runner" which we (me) had set up in #2197 is "large".

The goal of this issue is to explore if it can reduced - ideally without slowing down builds.

vorburger:
My analysis about this:

The https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners/about-github-hosted-runners#supported-runners-and-hardware-resources which this project originally (before my #2198 PR) used was a 2-core / 7 GB RAM / 14 GB SSD VM, and that was definitely too small, and didn't suffice.

https://github.com/google/android-fhir/actions/workflows/codeql.yaml currently takes ~4-11'.

https://github.com/google/android-fhir/actions/workflows/build.yml currently takes ~2-12'

#2295 will compare this with a much smaller VM...

vorburger:
> https://github.com/google/android-fhir/pull/2295 will compare this with a VM with 4-cores · 16 GB RAM · 150 GB SSD...

CodeQL took 9½' with this - about the same as with the huge VM.

Build took only 3' - but that's likely "wrong" (misleading), because all Gradle tasks are `UP-TO-DATE`; I probably need to make a trivial code change for this to be more realistic... or cherry-pick a random other PR to make a more sensible "real" comparison.

Kokoro failed with an unrelated problem; I've opened new #2297 about that.

> https://github.com/google/android-fhir/pull/2296 will compare this with a VM with 8-cores · 32 GB RAM · 300 GB SSD...

CodeQL took 6' with this - a bit faster, but I'm not sure how significant this is.

Build again just 2' - same "problem" as above.

vorburger:
>  or cherry-pick a random other PR to make a more sensible "real" comparison.

I'm using #2283 as a semi randomly chosen example (because it seems to have taken a bit longer to build, which is either a pure coincidence, or because the change of that dependency causes some real builds to happen) and have (temporarily) added that to the 2 PRs from above; the results are:

Build of #2283 with the very large VM as-is took ~8½' 

CodeQL (ditto) took ~5'

> https://github.com/google/android-fhir/pull/2295 will compare this with a VM with 4-cores · 16 GB RAM · 150 GB SSD...

Build with #2283 took **_TBD_**.

CodeQL (ditto) took **_TBD_**.

[...]

PR #2307, opened 2023-10-26
Merge CodeQL & Build GitHub Actions (re. #2294)

vorburger:
This is actually MORE (not less) efficient, because those 2 actions (partially) did the same thing anyway, so there really is no benefit to separately run them in parallel, other than increased resource usage, and more complexity to understand this project's CI.

Related to #2294.


jingtang10 (approved):
thanks @vorburger

vorburger:
PS: Retrying this in #2314 (but a little bit differently; hoping I now better understood the problem).

jingtang10:
this is lovely. thanks @vorburger!

Issue #2308, opened 2023-10-26
AndroidxTest JAR duplicates on testImplementation scope of `common` module


vorburger:
**Describe the bug**
My initial work on #2220 has identifies that the test scope classpath of the `common` module includes all of those `androidx.test` related classes twice, once from an e.g. `core-1.5.0-runtime.jar` and then again from some `transformed/core-1.5.0/jars/classes.jar` - which seems weird, to me.

**To Reproduce**
1. Checkout (or wait for merge of) PR TBD
2. Remove `jarPath.contains("-runtime.jar!/androidx/")` in `skipJAR(String jarPath)` of `ClasspathHellDuplicatesCheckRule`
3. Run `ClasspathHellDuplicatesCheckTest`
4. See error:

**Would you like to work on the issue?**
I could further investigate this, but perhaps someone more knowledgeable about Android would like to have a look instead?

My primary initial interest with #2220 is detecting **_runtime_** (not `testImplementation`) scope classpath duplicates in the APK, if any.

Issue #2220, opened 2023-09-29
Add a Classpath Duplicate JAR Detector Unit Test


vorburger:
http://blog2.vorburger.ch/2019/02/how-to-find-duplicate-classes-and.html

https://github.com/classgraph/classgraph/issues/256

https://javadocs.opendaylight.org/org.opendaylight.infrautils/sodium/org/opendaylight/infrautils/testutils/ClasspathHellDuplicatesCheckRule.html

https://issues.apache.org/jira/browse/FINERACT-919

@jingtang10 @omarismail94 this is to track what we chat about earlier today; I'll see if I can add this here, some time.

jingtang10:
see discussions in the above pr: https://github.com/google/android-fhir/pull/2309#issuecomment-1930250629

elicsmit:
Hi @vorburger! I'm interested in working on this. I'm a student looking to contribute to Fineract. May I take this issue?

vorburger:
@elicsmit 👋🏽 go for it, but please be aware that I'm no longer actively involved in this project, sorry. I'm not sure if this is still important. Any current maintainers, feel free to close this issue if you are cleaning up backlog - absolutely no offense taken

BTW just to make sure that you're not mixing something up, because you wrote _"to Fineract"_ ... this project/repo here (`ohs-foundation/android-fhir`) has absolutely nothing to do with https://github.com/apache/fineract, which perhaps is where you meant to engage instead? (The ONLY thing the 2 have in common is that I contributed to both in the past; and neither actively at present).

I would check for https://github.com/ohs-foundation/android-fhir/commits/master/ for current maintainers of this repo.

elicsmit:
Sorry for the confusion, @vorburger! I indeed mixed up the repositories. Thank you very much for pointing that out and for the guidance!

PR #2309, opened 2023-10-26
Introduce Classpath duplicates check test

vorburger:
Related to (and initial version of) #2220.

This introduces a (first) test which discovers, and fails in case of, duplicates in JARs on the classpath.

It already found 1 issue which seems a bit suspicious (to me), see #2308, but for now excludes that.

If this is acceptable, my idea (next step) is to extend this to the other (and more interesting) modules than just `common`.

### DIFF
--- a/common/build.gradle.kts
+++ b/common/build.gradle.kts
@@ -39,4 +39,5 @@ dependencies {
   testImplementation(Dependencies.junit)
   testImplementation(Dependencies.robolectric)
   testImplementation(Dependencies.truth)
+  testImplementation("io.github.classgraph:classgraph:4.8.163")
 }
--- a/common/src/test/java/com/google/android/fhir/ClasspathHellDuplicatesCheckRule.java
+++ b/common/src/test/java/com/google/android/fhir/ClasspathHellDuplicatesCheckRule.java
@@ -1,0 +1,182 @@
+/*
+ * Copyright 2023 Google LLC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *       http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.google.android.fhir;
+
+import junit.framework.AssertionFailedError;
+
+import org.junit.rules.TestRule;
+import org.junit.runner.Description;
+import org.junit.runners.model.Statement;
+
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.stream.Collectors;
+
+import io.github.classgraph.ClassGraph;
+import io.github.classgraph.ResourceList;
+import io.github.classgraph.ScanResult;
+
+/**
+ * JUnit Rule to run detect duplicate entries on the classpath.
+ *
+ * Such classpath duplicates often cause confusing and hard to debug problems, and
+ * can also "bloat" the file size of an application package like an  Android APK.
+ * They are caused by "unclean" dependencies, and typically solved with exclusions etc.
+ *
+ * Usage: <pre>public static {@literal @}ClassRule ClasspathHellDuplicatesCheckRule
+ *       dupes = new ClasspathHellDuplicatesCheckRule();</pre>
+ *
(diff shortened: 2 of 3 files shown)


jingtang10:
how can this be extended to other modules - the test cases themselves seem a little bit "involved". is there a ... gradle plugin we can find to do the same job?

@kevinmost @yigit for gradle fun.

vorburger:
> how can this be extended to other modules

I'm testing the water to see if you would find this acceptable... 😄 

If this is accepted and merged, then I would propose to follow-up and extend this to other modules as follows:

Copy/paste that `ClasspathHellDuplicatesCheckTest.kt`, which is very small (to me), to each other module. Maybe there is a way to make it even shorter? But we're talking about saving 2-3 lines more - not worth it (I think).

Note that the `ClasspathHellDuplicatesCheckRule.java` would NOT get copy/pasted - that stays single, in common. (I would have to spend a minute, if / after this is merged, to figure out if e.g. tests in `engine/` can depend on "test infrastructure" in `common/` in Gradle. If not, I would, in the future follow-up PR, propose moving this `ClasspathHellDuplicatesCheckRule` from `common/`  into (something like) a tiny new `common-testutil/` kind of module; perhaps there are other "test utilities" which could live there in the future.)

> is there a ... gradle plugin we can find to do the same job? @kevinmost @yigit for gradle fun.

There are 2-3 old and unmaintained ones. In my view, this approach is simpler and more future proof. In a perfect world with unlimited time one sure could write a Gradle plugin for this, but (unless @kevinmost @yigit you want to help me and do this?) I unfortunately wouldn't really have the bandwidth to do that here. Personally I also feel like Gradle plugins may be a bit overrated some times... great for code generators incl. compilers and other such tools, but in this particular case this is really just a (re-usable) test utility, which (IMHO) is great as simple JUnit test as-is here; I don't really see significant additional value in having this as a Gradle plugin. It could be "nice to have" - but then that plugin needs to be maintained for future major Gradle upgrades etc. TL;DR I'm willing to collaborate with someone wanting to do that, but not do it alone myself.

yigit:
So one problem with this approach is that you are traversing the classpath of the test, not the classpath of the library, which is different. I'm guessing what we really care is the library, not the test. 

I do also think that solving this in gradle would be nicer. 
You could write a task that will do this by resolving a classpath. But also, I don't think you even need that.

You can create a fatjar task that will bundle all dependencies and set its duplicate policy (just like in the test).

You can start in this module's build.gradle; once you have it working, you can move that into buildSrc and applly to all modules.

Unfortunately, i don't have much time to write it but [shadow plugin](https://imperceptiblethoughts.com/shadow/) is likely all you need.

vorburger:
> You can create a fatjar task that will bundle all dependencies and set its duplicate policy (just like in the test).

Oh... that's a very nice idea! You mean something like https://www.baeldung.com/gradle-fat-jar#creating-a-separate-task with a https://docs.gradle.org/current/javadoc/org/gradle/api/file/DuplicatesStrategy.html#FAIL - basically? That JAR would be "useless" of course, and just for testing this, right? Slightly inefficient, but it doesn't really matter I guess... I like this idea!

> Unfortunately, i don't have much time to write it but [shadow plugin](https://imperceptiblethoughts.com/shadow/) is likely all you need.

Can you elaborate on this approach a little more? Did you mean this instead of or in addition to using the Jar Task From the Java Plugin? From an only very quick glance at https://imperceptiblethoughts.com/shadow/introduction/, I don't see `shadow`  having a duplicate policy? Sorry I don't think I really understood what you meant here. (FYI: I **have** used `shadow` in past lives and projects, [e.g. here](https://github.com/OASIS-learn-study/minecraft-storeys-maker/blob/73d6a286cedc785cceafbc789c92a096e2a82328/web/build.gradle#L41); I just don't understand yet how you see it possibly adding any value for #2220...)

yigit:
Duplicate file strategy comes from the Jar task ([docs](https://docs.gradle.org/current/dsl/org.gradle.api.tasks.bundling.Jar.html#org.gradle.api.tasks.bundling.Jar:duplicatesStrategy))

My idea was to create a plugin that will create a Jar task from the runtime configuration. I'm actually not 100% if you need shadow or not, might just need to create a Jar task to jar it with duplication strategy of none.

I would probably try this first:
https://www.baeldung.com/gradle-fat-jar

```
// in gradle plugin
project.tasks.register("checkForDuplicatesion", Jar::class.java) {
    it.duplicatesStrategy = DuplicatesStrategy.FAIL
   // not sure about this but something like this code to find it
    it.from { configurations.named("runtimeClasspath").flatMap { it.isDirectory() ? it : zipTree(it) } }
}
```
and run it in CI. The android libraries' configuration might be a different one but you can simply list configurations to figure out which one.

Lmk if that helps.

jingtang10:
closing this pr since we're pursuing a different approach. let's continue the discussion in the issue itself.

Issue #2310, opened 2023-10-26
Build Failure:  CodeQL detected code written in Java/Kotlin but could not process any of it. 


vorburger:
E.g. https://github.com/google/android-fhir/actions/runs/6653530183/job/18079682920?pr=2309 (for #2309) failed with:

```
/opt/hostedtoolcache/CodeQL/2.15.1/x64/codeql/codeql database finalize --finalize-dataset --threads=8 --ram=29865 /home/runner/work/_temp/codeql_databases/java
  CodeQL detected code written in Java/Kotlin but could not process any of it. Review our troubleshooting guide at https://gh.io/troubleshooting-code-scanning/no-source-code-seen-during-build.
  Error: Encountered a fatal error while running "/opt/hostedtoolcache/CodeQL/2.15.1/x64/codeql/codeql database finalize --finalize-dataset --threads=8 --ram=29[86](https://github.com/google/android-fhir/actions/runs/6653530183/job/18079682920?pr=2309#step:7:87)5 /home/runner/work/_temp/codeql_databases/java". Exit code was 32 and last log line was: CodeQL detected code written in Java/Kotlin but could not process any of it. Review our troubleshooting guide at https://gh.io/troubleshooting-code-scanning/no-source-code-seen-during-build. See the logs for more details.
```

vorburger:
This seems to be caused since #2307 was just merged, and GitHub appears to get a bit confused... 😭 

...  because it still wants to run the old (now deleted) CodeQL related action (which doesn't exist anymore now).

I had hoped that a `git rebase` from master (locally, or by clicking that _Update branch_ button) would fix this...

... but apparently that is not so?! Hm.

vorburger:
I raised #2311 to re-create PR #2309 to see if that helps, but it does not - exact same problem.

Apparently GitHub is somehow "hard-coded"?! Just for the record, the exact problem is that:

1. The _Build_ Action, which WORKED in #2307, now for new PRs fails, with:

```
Run github/codeql-action/analyze@v2
/opt/hostedtoolcache/CodeQL/2.15.1/x64/codeql/codeql version --format=json
{
  "productName" : "CodeQL",
  "vendor" : "GitHub",
  "version" : "2.15.1",
[...]

PR #2311, opened 2023-10-26
Introduce Classpath duplicates check test (again)

vorburger:
[NB: This PR is a copy of #2309 due to #2310; let's see if this works better and passes CI build.]

Related to (and initial version of) #2220.

This introduces a (first) test which discovers, and fails in case of, duplicates in JARs on the classpath.

It already found 1 issue which seems a bit suspicious (to me), see #2308, but for now excludes that.

If this is acceptable, my idea (next step) is to extend this to the other (and more interesting) modules than just common.


vorburger:
Closing (un-merged) in favour of the (still relevant) original #2309.

PR #2315, opened 2023-10-26
Initial Git Tips documentation for contributors

vorburger:
@jingtang10 the other day in the meeting you suggested "sharing git know how" - here's a start, is this useful?

@omarismail94 and @fredhersch FYI

### DIFF
--- a/docs/contrib/git.md
+++ b/docs/contrib/git.md
@@ -12,11 +12,6 @@ To make simple single file changes, notably e.g. to `docs/*.md` Markdown files,
 it can be convenient to simply click the _"Pencil" button_ on GitHub's Web UI
 to _Edit in place._ For a more fully fledged IDE for docs, see [Docs](docs.md).
 
-The only minor drawback with this is that, if you have committer privileges,
-then this will create a branch on the "real" (upstream / origin) repository,
-instead of on your fork. While in general this is not recommended, it's OK
-in this case, as such branches get automatically deleted on merge of PRs.
-
 ### Locally with Git CLI
 
 Using the standard `git` CLI client, you would typically do the following:
--- a/docs/contrib/git.md
+++ b/docs/contrib/git.md
@@ -2,13 +2,15 @@
 
 This page documents [Git](https://git-scm.com) tips for contributors to this project.
 
+This project generally follows [these guidelines](https://github.com/google/fhir-data-pipes/blob/master/doc/review_process.md) from our `fhir-data-pipes` sister project.
+
 ## Usage
 
 ### Edit on GitHub Web UI
 
 To make simple single file changes, notably e.g. to `docs/*.md` Markdown files,
 it can be convenient to simply click the _"Pencil" button_ on GitHub's Web UI
-to _Edit in place._
+to _Edit in place._ For a more fully fledged IDE for docs, see [Docs](docs.md).
 
 The only minor drawback with this is that, if you have committer privileges,
 then this will create a branch on the "real" (upstream / origin) repository,
@@ -45,8 +47,10 @@ Using the standard `git` CLI client, you would typically do the following:
 
        git checkout THAT-BRANCH-NAME
        ... do some work ...
-       git commit --amend
-       git push --force YOUR-UID
+       git commit -a -m "Incorporated review feedback."
+       git push --force-with-lease YOUR-UID
+
+Note that, as per [these guidelines](https://github.com/google/fhir-data-pipes/blob/master/doc/review_process.md) we do not `commit --amend`. (Maintainers will _"squash"_ when merging PRs.)
 
 ### Locally with GitHub's CLI
 
--- a/docs/git.md
+++ b/docs/git.md
@@ -1,0 +1,74 @@
+# Git Tips
+
+This page documents [Git](https://git-scm.com) tips for contributors to this project.
+
+## Usage
+
+### Edit on GitHub Web UI
+
+To make simple single file changes, notably e.g. to `docs/*.md` Markdown files,
+it can be convenient to simply click the _"Pencil" button_ on GitHub's Web UI
+to _Edit in place._
+
+The only minor drawback with this is that, if you have committer privileges,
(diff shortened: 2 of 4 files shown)


docs/git.md
```diff
@@ -0,0 +1,74 @@
...
+### Edit on GitHub Web UI
+
+To make simple single file changes, notably e.g. to `docs/*.md` Markdown files,
+it can be convenient to simply click the _"Pencil" button_ on GitHub's Web UI
+to _Edit in place._
+
+The only minor drawback with this is that, if you have committer privileges,
+then this will create a branch on the "real" (upstream / origin) repository,
+instead of on your fork. While in general this is not recommended, it's OK
+in this case, as such branches get automatically deleted on merge of PRs.
```
jingtang10, line 16: why don't we just suggest contributors do this from the forked repo?

jingtang10 (changes requested):
@vorburger can you possibly incorporate the single-commit per review round point in bashir's doc: https://github.com/google/fhir-data-pipes/blob/master/doc/review_process.md? it's a great set of rules by the way we probably should incorporate most/all of them.

vorburger:
> @vorburger can you possibly incorporate the single-commit per review round point in bashir's doc: https://github.com/google/fhir-data-pipes/blob/master/doc/review_process.md? it's a great set of rules by the way we probably should incorporate most/all of them.

Done! Just pushed, plz have another look - better, like this? Merge?

docs/git.md
jingtang10: I just realised if you don't have write access it will actually create a change in your forked repo.

So few people have write permission in the repo I'm totally ok for us to not mention the second paragraph.

docs/contrib/git.md
```diff
@@ -0,0 +1,78 @@
...
+### Edit on GitHub Web UI
+
+To make simple single file changes, notably e.g. to `docs/*.md` Markdown files,
+it can be convenient to simply click the _"Pencil" button_ on GitHub's Web UI
+to _Edit in place._ For a more fully fledged IDE for docs, see [Docs](docs.md).
+
+The only minor drawback with this is that, if you have committer privileges,
+then this will create a branch on the "real" (upstream / origin) repository,
+instead of on your fork. While in general this is not recommended, it's OK
+in this case, as such branches get automatically deleted on merge of PRs.
```
jingtang10:
```suggestion
```

docs/git.md
vorburger: @jingtang10 sure thing; done! Merge?

docs/contrib/git.md
```diff
@@ -0,0 +1,73 @@
...
+       git checkout master
+       git pull
+       git checkout THAT-BRANCH-NAME
+       git rebase master
+
+4. When you get code review feedback and would like to make some changes:
+
+       git checkout THAT-BRANCH-NAME
+       ... do some work ...
+       git commit -a -m "Incorporated review feedback."
```
jingtang10, line 45: I was wondering about this

```
Please try to have a single commit per each review round. When you want to send your PR, if you have multiple commits, [squash them](https://stackoverflow.com/questions/5189560/how-to-squash-my-last-x-commits-together). Also, when you are responding to review comments, make sure to create exactly one commit to address all required changes. You can use [--amend](https://git-scm.com/docs/git-commit#Documentation/git-commit.txt---amend) feature of git commit. This is to help your reviewer see all of the changes since last review round in a single commit.
```

In https://github.com/google/fhir-data-pipes/blob/master/doc/review_process.md which I particularly like. I think having clear review history is a boost to productivity. do you think we should recommend the contributor to create 1 commit for each round of review?

docs/contrib/git.md
vorburger: @jingtang10 I think we may just have "crossed"? I've (just) rm this, in a34e8a5036eae0d026834eed3df2efaeeb0b783d, did you see? Please _Resolve Conversation_ if OK for you?

docs/contrib/git.md
jingtang10: @bashir2 fyi

I'm ok to copy some of the tips bashir wrote actually.

jingtang10 (approved):
thanks for writing these tips Michael! I recognise this is of huge value to developers - and I'm also ok for us to iterate on this instead of dragging this out.

approved and fix/merge as you see fit. thanks!

docs/contrib/git.md
vorburger: Yeah I do agree with that recommendation! I was hoping the link(s, twice) to https://github.com/google/fhir-data-pipes/blob/master/doc/review_process.md  which I have added (did you see?) and the change I made to the recommended Git CLI commands (it originally used to recommend the `git commit --amend` flag, which I have removed now; did you notice?) the instructions now do lead to having a single commit per each review round. Do you think it's not clear enough as-is? I was trying to avoid copy/pasting from https://github.com/google/fhir-data-pipes/blob/master/doc/review_process.md ...

docs/contrib/git.md
jingtang10: resolving

docs/contrib/git.md
jingtang10: i think the point is that whilst you might have many local commits - when you push you should squash them? or, whilst you do git commit -a -m the first time, you should git commit --amend in the following commits before replying to the review comments?

vorburger: Oh, I see... tell you what, I'll resolve this comment thread so that we can merge this already, to have at least a first version of Git guidelines for contributors in. Then I'll follow-up with a new PR to refine this, and we can further discuss there. Hope that works for you!

vorburger: ==> 2617

bashir2: Thanks @vorburger for creating this; and yes @jingtang10 the point of that recommendation is that after the review starts, there should exactly be one commit per review round. Otherwise it is very hard to follow what changes are done to address each comment (and sometimes even finding previous comments).

Issue #2297, opened 2023-10-25
Kokoro Check build failure due to NodeJS installation failure


vorburger:
#2295 (related to #2294) failed the Kokoro Check build:

https://source.cloud.google.com/results/invocations/80bddf42-e6cc-4be7-862b-c625032e4abb/targets/openhealthstack%2Fandroid-fhir%2Fgcp_ubuntu%2Fpresubmit/log

```
[ID: 9480442] Executing command via SSH:
export KOKORO_BUILD_NUMBER="3818"
export KOKORO_JOB_NAME="openhealthstack/android-fhir/gcp_ubuntu/presubmit"
source /tmpfs/kokoro-env_vars.sh; cd /tmpfs/src/ ;  chmod 755 github/android-fhir/kokoro/gcp_ubuntu/kokoro_build.sh ;  PYTHON_3_VERSION="$(pyenv which python3 2> /dev/null || which python3)" ; PYTHON_2_VERSION="$(pyenv which python2 2> /dev/null || which python2)" ; if "$PYTHON_3_VERSION" -c "import psutil" ; then KOKORO_PYTHON_COMMAND="$PYTHON_3_VERSION" ; else KOKORO_PYTHON_COMMAND="$PYTHON_2_VERSION" ; fi > /dev/null 2>&1 ; echo "export KOKORO_PYTHON_COMMAND="$KOKORO_PYTHON_COMMAND"" > "$HOME/.kokoro_python_vars" ; nohup bash -c "( rm -f /tmpfs/kokoro_build_exit_code ; github/android-fhir/kokoro/gcp_ubuntu/kokoro_build.sh   ; echo \${PIPESTATUS[0]} > /tmpfs/kokoro_build_exit_code ) > /tmpfs/kokoro_build.log 2>&1" > /dev/null 2>&1 & echo $! > /tmpfs/kokoro_build.pid ; source "$HOME/.kokoro_python_vars" ; "$KOKORO_PYTHON_COMMAND" /tmpfs/kokoro_log_reader.py /tmpfs/kokoro_build.log /tmpfs/kokoro_build_exit_code /tmpfs/kokoro_build.pid /tmpfs/kokoro_log_reader.pid --start_byte 0

npm WARN using --force I sure hope you know what you are doing.
/usr/local/bin/n -> /usr/local/lib/node_modules/n/bin/n
+ n@9.2.0
updated 1 package in 1.022s
  installing : node-v16.18.0
[...]

PR #2316, opened 2023-10-26
Replace styfle/cancel-workflow-action with new GitHub `concurrency:` standard

vorburger:
See https://github.com/styfle/cancel-workflow-action/blob/main/README.md (and go/github-actions#cancel-workflows)

I've intentionally raised this separately from #2313 (which is included here) and #2314 (which builds on this) to make it easier to review, and revert in case it's needed because this unexpectedly breaks anything (which it should not).

This should be rebased after #2313 is merged.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -35,8 +35,11 @@ env:
   GRADLE_BUILD_ACTION_CACHE_KEY_PREFIX: "fhir" # change this to invalidate cache
 
 concurrency:
-  group: ${{ github.workflow }}-${{ github.head_ref || github.ref || github.run_id }}
-  cancel-in-progress: true
+  # github.head_ref uniquely identifies Pull Requests (but is not available when building branches like main or master)
+  # github.ref is the fallback used when building for workflows triggered by push
+  # Note that || are fallback values (not "concatenations")
+  group: ${{ github.workflow }}-${{ github.head_ref || github.ref }}
+  cancel-in-progress: true # Use e.g. ${{ github.ref != 'refs/heads/main' }} (or master, until #2180) to only cancel for PRs not on branch
 
   # A workflow run is made up of one or more jobs that can run sequentially or in parallel
 jobs:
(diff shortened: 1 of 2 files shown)


vorburger:
> Kokoro: Build and Device Tests — Kokoro build failed. Check logs for details.

due to #2297

vorburger:
> Kokoro: Build and Device Tests — Kokoro build failed. Check logs for details.

OMG, it's #1482 again... how about we just do #2318 at this point?

.github/workflows/build.yml
```diff
@@ -34,6 +34,10 @@ env:
   GRADLE_BUILD_ACTION_CACHE_DEBUG_ENABLED: true
   GRADLE_BUILD_ACTION_CACHE_KEY_PREFIX: "fhir" # change this to invalidate cache
 
+concurrency:
+  group: ${{ github.workflow }}-${{ github.head_ref || github.ref || github.run_id }}
```
jingtang10, line 38: if the run_id is included here in the group id, does this not mean that each run will have a separate group id? and does that mean previous runs will not be cancelled?

actually, if the head_ref is part of this, how would cancellation even help anything? shouldn't we cancel jobs that are in the same branch? i.e. if you push a new commit we should cancel the previous job?

jingtang10: thanks michael for the explanation.

this is clear to me now.

up to you about the run_id. but either way can we add a single line of comment to explain what's happening here? doesn't have to go through the length of your comment above, but just noting that we're using the branch/pr ref etc so possible to cancel previous runs.

jingtang10 (approved):

.github/workflows/build.yml
jingtang10: once addressed please feel free to merge.

vorburger: Removed `github.run_id` and added inline documentation comments; I will self merge when build is green.

vorburger:
FTR: I "tested" this by clicking the _Update Branch_ button, and seeing https://github.com/google/android-fhir/actions/runs/6696513296 got cancelled.

I'm going to merge this now, without waiting for Kokoro, because this is a GitHub-Action-only change, which cannot impact Firebase Test Lab runs (so Kokoro's result is irrelevant in this particular case).

Issue #2239, opened 2023-10-09
Allow modification of Questionnaire submit action button text via configuration


ellykits:
**Is your feature request related to a problem? Please describe.**
As part of the enhancement, we'd like to set the text for the submit action button via configuration.

**Describe the solution you'd like**
Expose a function/property etc. to allow modification of the submit action button text.

**Describe alternatives you've considered**
For the set save button text, we can `findTheViewById` and apply the text as desired but this will be subject to code changes.

**Additional context**
In order to filter the content of our registers (list of resources) we opted to re-use the UI rendered via Questionnaires instead of implementing custom UI to capture the content for the filter. As part of the requirement a configuration to change the text for the submit action to "APPLY FILTERS" is needed.  

See the screenshot below, the Questionnaire is the second screen in the pic.

![image (1)](https://github.com/google/android-fhir/assets/26865532/0287a73d-6efd-4383-a199-e868f6aa0eac)

**Would you like to work on the issue?**
Yes

MJ1998:
The app bar is not controlled by the QuestionnaireFragment. You can probably achieve this by customizing app bar in your application for this particular fragment where you are hosting the QuestionnaireFragment as its child fragment. Example is [this older version of ScreenerFragment](https://github.com/google/android-fhir/blob/225ff1dadc853fd9868086509e6657855691d7b6/reference/src/main/java/com/google/android/fhir/reference/ScreenerFragment.kt#L43-L68)

Or am I missing something ?

DebbieArita:
@MJ1998 For this issue, we're trying to make the submit action button text on the QuestionnaireFragment configurable. Reference the second screen whose action button reads 'APPLY FILTERS'. There are cases where I want to be able to override the default 'SUBMIT' text and have any other text on the action button. The app bar issue will be addressed [here](https://github.com/google/android-fhir/issues/2221). I hope that clarifies.

f-odhiambo:
@DebbieArita has an open PR on this

PR #2319, opened 2023-10-27
Allow modification of Questionnaire submit action button text via configuration

DebbieArita:
…figuration

Fixes https://github.com/google/android-fhir/issues/2239

**Description**

**Alternative(s) considered**

**Type**

**Checklist**


ellykits (changes requested):

MJ1998 (approved):
LGTM

DebbieArita:
@santosh-pingle @MJ1998 could you help me resolve the Kokoro check and maybe point out anything else I need to close on this?

MJ1998:
@DebbieArita 
Kokoro is running right now. Let's wait for the result.

DebbieArita:
@santosh-pingle @MJ1998 could you help me resolve the Kokoro check and maybe point out anything else I need to close on this? 

> @DebbieArita Kokoro is running right now. Let's wait for the result.

Alright, thank you.

jingtang10 (changes requested):
Please try to add this line:

```
    <string name="submit_questionnaire">New Submit String</string>
```

to the resource files in your app, and I think it'll override the string value in the library resources.

and if you really need to do this at run-time (which I'm not too keen on actually), you can do it this way: https://blog.kiprosh.com/how-to-replace-strings-xml-dynamically-in-android/

Please try this and see if this works better for you. My concern here is that if we provide an API / configuration for each string we'll end up with a ton of new API surfaces... and I worry that we'll struggle to manage.

ellykits:
@jingtang10 Changing the text via the `strings.xml` file would be the easiest however, this would change the text through the application which is not desired in our current use case. We'd want to change the texts only for particular questionnaires. 

cc @DebbieArita

ellykits:
> and if you really need to do this at run-time (which I'm not too keen on actually), you can do it this way: https://blog.kiprosh.com/how-to-replace-strings-xml-dynamically-in-android/
> 
> Please try this and see if this works better for you. My concern here is that if we provide an API / configuration for each string we'll end up with a ton of new API surfaces... and I worry that we'll struggle to manage.

I missed reading this part of your comment @jingtang10. We only required the update for the questionnaire submission text. We'll try the suggestion.

jingtang10 (approved):
i think this change alone is fine. but let's be careful with introducing too many arguments in the future.

Issue #2304, opened 2023-10-26
Run Flank in GitHub Action


vorburger:
I would like to explore if it's possible to run (trigger?) Flank (https://flank.github.io/flank/) in the GitHub Action (which now works fine otherwise),

i.e.  `./gradlew :engine:runFlank` etc.

as in https://github.com/google/android-fhir/blob/master/kokoro/gcp_ubuntu/kokoro_build.sh

If this works, then I'm not sure this project still needs the Google internal only Kokoro infrastructure? (Thus "fixing" e.g. #2300 or #2297 et al.)

This would, to me, lead to a simplification of this project's (to me, seemingly) "complex" CI infra.

omarismail94:
`runFlank` runs the Firebase tests. We need a way to authenticate with Firebase if we run things in GitHub, and the only way I can think of is using a JSON key. Exporting a JSON key is not allowed and Im not sure how difficult it is to get an exception. Why this works in Kokoro is becasue it runs on GCP, which uses IAM roles to authenticate as opposed to JSON keys

vorburger:
> Exporting a JSON key is not allowed and Im not sure how difficult it is to get an exception. 

I'll reach out privately because I would like to learn more about this...

vorburger:
OK, I'm learning more about this, so there are 1-2 problems, here:

1. This projects' Firebase GCP project has Org restrictions which make it impossible to export Service Account credentials as a JSON file (which I would then securely store in a GitHub Secret). There is a process around it; I'll try to find out more about if this may be feasible.

2. Firebase doesn't seem to have an (only) "Test Lab Runner" kinda role?! That's... quite surprising!

vorburger:
Note to future self: The following is just a _Warning_ and it already always appears and is no problem:

```
$ ./gradlew :engine:runFlank

> Task :engine:execFlank
version: v21.11.0
revision: ...
session id: ...

Warning: Failed to make bucket for android-fhir-instrumeted-tests
Cause: The requested bucket name is not available. The bucket namespace is shared by all users of the system. Please select a different name and try again.
```

This is NOT related to and not NOT the root cause of problems such as the following:

```
ftl.run.exception.PermissionDenied: Caused by: com.google.api.client.googleapis.json.GoogleJsonResponseException: 403 Forbidden
GET https://testing.googleapis.com/v1/testEnvironmentCatalog/ios?projectId=android-fhir-instrumeted-tests
{
  "code": 403,
  "errors": [
    {
[...]

PR #2322, opened 2023-10-27
Switch Firebase Test Lab Artifacts to another GCP GCS Bucket

vorburger:
Related to #2304

Motivated by https://github.com/google/android-fhir/issues/2304#issuecomment-1783526569.

Not technically strictly required, but it (having the GCS bucket in the same single GCP project) seems simpler, to me.

Happy to abandon this if I'm missing something.


omarismail94:
Kokoro uploads build artifacts to the GCS bucket, `android-fhir-build-artifacts`, which required some work to set up. If/once we do get rid of Kokoro, we would just delete the whole kokoro_build.sh script. Until then, i will close this PR

PR #2329, opened 2023-11-08
Api to clear all answers of questionnaire response.

### LINKED ISSUE #2221: Clear all action on Questionnaire
We'd like to reset all the answered fields in the questionnaires via a `Clear all` action.

The "Clear all" function will remove all the answer items for the "current" QuestionnaireResponse when the questionnaire is being used actively.

None

**Additional context**
In order to filter the content of our registers (list of resources) we opted to re-use the UI rendered via Questionnaires instead of implementing custom UI to capture the content for the filter. As part of the requirement, a "Clear all" action is required to reset all the data filter fields".  

See the screenshot below, the Questionnaire is the second screen in the pic.

image (1)

Yes

santosh-pingle:

Fixes #2221 

**Description**
Provide api to clear all answers present in the questionnaire response.

**Alternative(s) considered**

**Type**
Choose one: Feature 

**Checklist**


datacapture/src/main/java/com/google/android/fhir/datacapture/QuestionnaireViewModel.kt
```diff
@@ -421,6 +421,27 @@ internal class QuestionnaireViewModel(application: Application, state: SavedStat
...
+   * Recursively clears the answer components of a questionnaire response item by iterating through
+   * each child item.
+   *
+   * @param item The item component of the questionnaire response to clear.
+   */
+  private fun clearItemAnswerComponent(item: QuestionnaireResponseItemComponent) {
+    for (child in item.item) {
+      clearItemAnswerComponent(child)
+    }
+    item.answer = listOf()
```
MJ1998, line 442: You can use `QuestionnaireResponse.allItems` defined in `MoreQuestionnaireResponse`
```suggestion
    questionnaireResponse.allItems().forEach { it.answer = emptyList() }
    modificationCount.update { it + 1 }
```

MJ1998 (changes requested):

datacapture/src/main/java/com/google/android/fhir/datacapture/QuestionnaireViewModel.kt
santosh-pingle: done.

MJ1998 (commented):
What about draftAnswerMap, modifiedQuestionnaireResponseItemSet, responseItemToAnswersMapForDisabledQuestionnaireItem ,etc ?
Do they need to be cleared as well ?

santosh-pingle:
> What about draftAnswerMap, modifiedQuestionnaireResponseItemSet, responseItemToAnswersMapForDisabledQuestionnaireItem ,etc ? Do they need to be cleared as well ?

I am trying to understand why it is necessary to clear these data structures individually. Is there any use case that requires attention? Wouldn't calling this function modificationCount.update { it + 1 } be sufficient?

MJ1998:
I am suspecting that draftAnswers can still show up. 

Notice the `answersChangedCallback`. Whenever any answer changes we updated draftAnswer, modifiedQuestionnaireResponseItemSet and then do `modificationCount.update { it + 1 }`.
So I think draftAnswer and modifiedQuestionnaireResponseItemSet definitely needs to be cleared otherwise the obsolete values will still show. Try a testcase where you can have draftAnswers, eg, edittext.

Again, the cached answers in `responseItemToAnswersMapForDisabledQuestionnaireItem` should  also be cleared even though I don't think it will effect anything but no point keeping obsolete answers.

santosh-pingle:
> I am suspecting that draftAnswers can still show up.
> 
> Notice the `answersChangedCallback`. Whenever any answer changes we updated draftAnswer, modifiedQuestionnaireResponseItemSet and then do `modificationCount.update { it + 1 }`. So I think draftAnswer and modifiedQuestionnaireResponseItemSet definitely needs to be cleared otherwise the obsolete values will still show. Try a testcase where you can have draftAnswers, eg, edittext.
> 
> Again, the cached answers in `responseItemToAnswersMapForDisabledQuestionnaireItem` should also be cleared even though I don't think it will effect anything but no point keeping obsolete answers.

Ohhh!, thats correct, it requires to clear draftAnswer or else it keep showing partial answer in the textfields, e.g partial date answer in the datePicker. Basically, clears dataStructures which get modified by answerChangeCallback.

PR #2370, opened 2023-12-12
Handle op:add for same path in patch

### LINKED ISSUE #2342: Error with Json Patch
**Describe the bug**
Right now, we have a problem adding multiple activities to the CarePlan. The issue lies in the Json Patch. It's giving us the same path for each activity in the patch file. When the changes are merged, the system compares them based on the path [in this file](https://github.com/google/android-fhir/blob/704d753daeec67f05dbeea9e138cd9ada686043b/engine/src/main/java/com/google/android/fhir/sync/upload/patch/PerResourcePatchGenerator.kt#L116) `.associateBy { it.optString("op") to it.optString("path") }`. This causes it to pick only one activity that's being updated, resulting in the server returning a CarePlan with just one added activity.

After investigating the Json Patch further, it seems there are problems with it, causing it to return the wrong patch. [json-patch](https://github.com/java-json-tools/json-patch/issues)

**Component**
Fhir Engine

**To Reproduce**
1. Open an existing CarePlan in app (already in sync with server)
2. Update it to add 2 or more Task activities to the CarePlan, to generate LocalChange for the CarePlan
3. Sync to server, only the last task activity in the Careplan gets synced

**Additional context**
Patch File - LocalChange payload
```
[
  {
    "op": "replace",
    "path": "/activity/6/detail/status",
    "value": "in-progress"
  },
  {
    "op": "add",
    "path": "/activity/-",
    "value": {
      "outcomeReference": [
        {
          "reference": "Task/4f26da79-4571-4442-911a-c3a707e3e662",
          "display": "Viral Load Collection"
        }
      ],
      "detail": {
        "kind": "Task",
        "code": {
          "coding": [
            {
              "system": "https://d-tree.org/",
              "code": "art-client-viral-load-collection"
            }
          ]
        },
        "status": "in-progress",
        "scheduledPeriod": {
          "start": "2023-11-15T11:45:40+02:00",
          "end": "2023-12-14T11:10:34.227+03:00"
        },
        "performer": [
          {
            "reference": "Practitioner/311015",
            "display": "John Doe"
          }
        ],
        "description": "Viral Load Collection"
      }
    }
  },
  {
    "op": "add",
    "path": "/activity/-",
    "value": {
      "outcomeReference": [
        {
          "reference": "Task/6f6e6672-ef2d-4e57-9854-0a04cd0fb001",
          "display": "Index Case Testing"
        }
      ],
      "detail": {
        "kind": "Task",
        "code": {
          "coding": [
            {
              "system": "https://d-tree.org/",
              "code": "art-client-index-case-testing"
            }
          ]
        },
        "status": "in-progress",
        "scheduledPeriod": {
          "start": "2023-11-15T11:45:40+02:00",
          "end": "2023-12-14T11:10:34.227+03:00"
        },
        "performer": [
          {
            "reference": "Practitioner/311015",
            "display": "John Doe"
          }
        ],
        "description": "Index Case Testing"
      }
    }
  }
]
```

Local CarePlan.
[local-careplan.json](https://github.com/google/android-fhir/files/13484582/local-careplan.json)

Remote CarePlan
[remote-careplan.json](https://github.com/google/android-fhir/files/13484589/remote-careplan.json)
 
yes, @LZRS and @evance-mose 

cc @KhumboLihonga

omarismail94:

Fixes #2342

**Description**
When merging JSON patches:
    *  "replace" and "remove" operations from the second patch will overwrite any existing operations for the same path.
    *  "add" operations from the second patch will be added to the list of operations for that path, even if operations already exist for that path.

**Alternative(s) considered**

**Type**
Choose one: (**Bug fix** | Feature | Documentation | Testing | Code health | Builds | Releases | Other)

**Checklist**


kelvin-ngure:
@omarismail94 since mergePatches is being called in a reduce function, the first time it runs, the parameters will receive strings withthis sort of structure (the below are patch strings)

[{"op":"add","path":"\/member\/-","value":{"entity":{"reference":"Patient\/dbe7936d-a1bc-4593-b219-025f34a8a8bf"}}}]
[{"op":"add","path":"\/member\/-","value":{"entity":{"reference":"Patient\/e4ebe735-cbf5-4365-b8ad-0169466d51r9"}}}]

The output of the mergePatches function has this structure since it is create from an Array Node (I use the word Patch String in place of the above objects)

[
    Patch string,
    Patch String,
]

Therefore, the next time the merge patches function is run (e.g when there are more than 2 LocalChange), the parameter for firstPatch will receive the correct data type but not the same structure. The first time, it was like this => [{}] but subsequent times it will be like this => [[{}], [{}]]. 

Will the op.get("path") still work? I assume that due to the nesting, path won't be recognized as a key. Thoughts?

engine/src/main/java/com/google/android/fhir/sync/upload/patch/PerResourcePatchGenerator.kt
```diff
@@ -96,24 +95,40 @@ internal object PerResourcePatchGenerator : PatchGenerator {
...
-   * Creates a mutable map from operation type (e.g. add/remove) + property path to the entire
-   * operation containing the updated value. Two such maps can be merged using `Map.putAll()` to
-   * yield a minimal set of operations equivalent to individual patches.
-   */
-  private fun JSONArray.patchMergeMap(): MutableMap<Pair<String, String>, JSONObject> {
-    return (0 until this.length())
-      .map { this.optJSONObject(it) }
-      .associateBy { it.optString("op") to it.optString("path") }
-      .toMutableMap()
+    patchNode1.forEach { op ->
```
kelvin-ngure, line 116: @omarismail94 

could we rename this op to patch or patchNode or something patch related? Calling it op can cause confusion and even forced us to name the actual op in the patch as opType

kelvin-ngure (changes requested):

engine/src/main/java/com/google/android/fhir/sync/upload/patch/PerResourcePatchGenerator.kt
omarismail94: will do!

omarismail94:
> Edit: Below the *** was a concern had but I was wrong since the code is extracting the object from the list. And at the end, only the objects are inserted into the output array. So the structure during the reduction stays the same.
> 
> @omarismail94 since mergePatches is being called in a reduce function, the first time it runs, the parameters will receive strings with this sort of structure (the below are patch strings)
> 
> [{"op":"add","path":"/member/-","value":{"entity":{"reference":"Patient/dbe7936d-a1bc-4593-b219-025f34a8a8bf"}}}] [{"op":"add","path":"/member/-","value":{"entity":{"reference":"Patient/e4ebe735-cbf5-4365-b8ad-0169466d51r9"}}}]
> 
> The output of the mergePatches function has the below structure since it is created from an Array Node (I use the word Patch String in place of the above objects)
> 
> [ Patch string, Patch String, ]
> 
> Therefore, the next time the merge patches function is run (e.g when there are more than 2 LocalChange), the parameter for firstPatch will receive the correct data type but not the same structure. The first time, it was like this => [{}] but subsequent times it will be like this => [[{}], [{}]].
> 
> Will the op.get("path") still work? I assume that due to the nesting, path won't be recognized as a key. Thoughts?

@kelvin-ngure thanks for clarifying!  i assume the change fixes the issue you have then?

kelvin-ngure:
@omarismail94 yes it looks like it will work for my highlited usecase

MJ1998 (commented):
Have you tried using latest version of jsonpatch ?

I think for each activity the path generated should be unique and not `activity/-`

We are using 1.13 but 1.9 is also available. I haven't checked if they have solved this problem or not but worth verifying.

engine/src/main/java/com/google/android/fhir/sync/upload/patch/PerResourcePatchGenerator.kt
```diff
@@ -96,24 +95,40 @@ internal object PerResourcePatchGenerator : PatchGenerator {
...
   private fun mergePatches(firstPatch: String, secondPatch: String): String {
     // TODO: validate patches are RFC 6902 compliant JSON patches
-    val firstMap = JSONArray(firstPatch).patchMergeMap()
-    val secondMap = JSONArray(secondPatch).patchMergeMap()
-    firstMap.putAll(secondMap)
-    return JSONArray(firstMap.values).toString()
-  }
+    val objectMapper = ObjectMapper()
+    val firstPatchNode: JsonNode = JsonLoader.fromString(firstPatch)
+    val secondPatchNode: JsonNode = JsonLoader.fromString(secondPatch)
```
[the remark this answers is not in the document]
jingtang10: i think node is fine

jingtang10 (approved):

PR #2380, opened 2023-12-18
Dependency versions fix

vitorpamplona:
Moves the version selection from resolutionStrategy to dependencyConstraints to make sure it is exported to maven correctly on publishing.

**Type**
Bug fix

**Checklist**


vitorpamplona:
@jingtang10 👆

MJ1998 (commented):
This is interesting. 

One question: Can we add this constraint to project level build.gradle since we are adding it to all the libraries ?

vorburger:
@vitorpamplona how about moving more dependencies to the #2306 approach instead?

vitorpamplona:
@vorburger I looked at Version Catalogs but I didn't find a way to explicitly force strict versions for our dependencies. 

@MJ1998 The constraint block to specify versions seems to be required every time one of the modules declares the dependency on HAPI individually. It then places the correct `<dependencyManagement>` tag on the maven side. I think your idea would work if gradle exported hierarchical `pom.xml`s.

vorburger:
> @vorburger I looked at Version Catalogs but I didn't find a way to explicitly force strict versions for our dependencies.

How do you mean? In https://github.com/google/android-fhir/blob/977ecfac11a5fc7dc283b6f1911cb800af9193fb/gradle/libs.versions.toml#L4 there is a fixed version, right?

But I'm possibly missing the full problem of what this proposes to fix.

vitorpamplona:
I just found that gradle does offer a `!!` notation for strict. 

```
    // or...
    implementation('org.slf4j:slf4j-api:[1.7, 1.8[!!1.7.25')
    // is equivalent to
    implementation('org.slf4j:slf4j-api') {
        version {
           strictly '[1.7, 1.8['
           prefer '1.7.25'
        }
    }
```

It's kinda hidden in the specs (maybe outdated?), but we can test this later on with the TOML syntax. I just don't know how to declare a proper `constraints` block in TOML so that it becomes a `<dependencyManagement>` block in maven.

jingtang10:
@vorburger the problem we're trying to solve is to stop applications using our libraries from having to declare resolution strategies like in this file: https://github.com/google/android-fhir/pull/2380/files#diff-a645a2c44f3874493dc92c37e4dbbcd33cb5042edd2de9d9ce6259ffdbf33010

This is a pretty strong functional requirement. I would place this as higher priority than migrating to toml files.

So what I would suggest is to complete this fix first, and continue to migrate to toml wherever we can (it appears a partial migration is possible). And hopefully complete the migration soon with a way to declare `constraints` as @vitorpamplona mentioned in the above comment.

buildSrc/src/main/kotlin/Dependencies.kt
```diff
@@ -79,13 +99,22 @@ object Dependencies {
...
+      "$moduleGroup:jackson-module-jaxb-annotations:${Versions.jackson}"
+    const val jsr310Base = "$datatypeGroup:jackson-datatype-jsr310:${Versions.jackson}"
+
+    const val annotations = "$annotationsBase:${Versions.jackson}"
+    const val bom = "$bomBase:${Versions.jackson}"
+    const val core = "$coreBase:${Versions.jacksonCore}"
+    const val databind = "$databindBase:${Versions.jackson}"
+    const val dataformatXml = "$dataformatXmlBase:${Versions.jackson}"
+    const val jaxbAnnotations = "$jaxbAnnotationsBase:${Versions.jackson}"
+    const val jsr310 = "$jsr310Base:${Versions.jackson}"
```
jingtang10, line 117:
```suggestion
```

this whole block seems unused? can we delete?

datacapture/build.gradle.kts
```diff
@@ -108,6 +104,13 @@ dependencies {
   testImplementation(Dependencies.mockitoKotlin)
   testImplementation(Dependencies.robolectric)
   testImplementation(Dependencies.truth)
+
+  constraints {
+    Dependencies.hapiFhirConstraints().forEach { (libName, constraints) ->
+      api(libName, constraints)
+      implementation(libName, constraints)
```
jingtang10, line 111: I find it really weird that whilst we're going through the constraints for all the libraries, we'd have to declare them for `api` and `implemention` and `compileOnly` separately. comparing this flie with `common/build.gradle.kts` and `workflow-testing/build.gradle.kts` it makes me slightly confused...

isn't there a better way of writing this code so it's clearer? surely we only need to call `api` here because above on line 81 in this file we're using an `api` dependency right? does that mean we don't actually need to declare the constraint on all these libraries?

jingtang10 (approved):
thanks @vitorpamplona for fixing this - huge improvement.

a couple of suggestions. would be great to fix in this pr but can be in an immediate follow-up too.

datacapture/build.gradle.kts
vitorpamplona: The issue on this one is that Maven doesn't know the difference between `api` and `implementation`. But gradle does. So, on pure gradle builds, the version constraint must match how the library, OR a library that depends on a contained library, was added (`api`, `implementation`, `compile`, etc). On the workflow build, the evaluator's dependency stack also tries to add HAPI FHIR, so the constraint must be on both.

buildSrc/src/main/kotlin/Dependencies.kt
vitorpamplona: Nice catch! I just removed them.

vorburger:
> It's kinda hidden in the specs (maybe outdated?), but we can test this later on with the TOML syntax. I just don't know how to declare a proper constraints block in TOML so that it becomes a <dependencyManagement> block in maven.

I don't suppose those `strictly` and `prefer` (and `reject` and `rejectAll`) keywords [documented here](https://docs.gradle.org/current/userguide/platforms.html#sub::toml-dependencies-format) are what you are looking for?

PR #2383, opened 2023-12-19
Add API call options to the FhirOperator for OpenSRP

vitorpamplona:
Offers additional options to evaluate measures and plan def apply in the FhirOperator

**Type**
Feature

**Checklist**


workflow/src/main/java/com/google/android/fhir/workflow/FhirOperator.kt
```diff
@@ -162,20 +167,33 @@ internal constructor(
...
-    )
+
+    val report = measureProcessor.evaluateMeasure(measure, start, end, reportType, subjects, null)
+
+    // add subject reference for non-individual reportTypes
+    if (report.type.name == MeasureReportType.SUMMARY.name) {
+      if (!practitioner.isNullOrBlank()) {
+        report.setSubject(Reference(practitioner))
+      } else if (!subjectId.isNullOrBlank()) {
+        report.setSubject(Reference(subjectId))
```
MJ1998, line 193: practitioner and subjectId should mandatorily be in either of the following formats for `Reference`:-
(e.g. "Patient/123" or "http://example.com/Patient/123")

From comment above we can use `practitionerWithType` and `subjectWithType` here!

workflow/src/main/java/com/google/android/fhir/workflow/FhirOperator.kt
```diff
@@ -162,20 +167,33 @@ internal constructor(
     start: String,
     end: String,
     reportType: String,
-    subjectId: String?,
-    practitioner: String?,
+    subjectId: String? = null,
+    practitioner: String? = null,
```
MJ1998, line 171: Not sure how `MeasureProcessor` evaluates when both practitioner and subject Id is given.
If it assumes `MeasureEvalType.SUBJECT` even when practitioner is given then is it wrong ?

Consider dividing this API into 2 : `evaluateMeasureForSubject` & `evaluateMeasureForPopulation` ?

MJ1998 (changes requested):
In generateCarePlan APIs,
can we add options for the third API as well:-

```
fun generateCarePlan(
    planDefinitionId: String,
    subject: String,
    encounterId: String?,
  ): IBaseResource {
  ```

**Comment by MJ1998** [comment]
Thanks @vitorpamplona for the PR. This looks necessary!

Requesting to create an issue and tag it here as well! Thanks.

File: workflow/src/main/java/com/google/android/fhir/workflow/FhirOperator.kt (hunk 2 above)
Comment (vitorpamplona) [reply workflow/src/main/java/com/google/android/fhir/workflow/FhirOperator.kt:171]: Yep, that seems to be a pattern in most of the server tests I have seen. Practitioner overrides subject.

**Comment by vitorpamplona** [comment]
The 

```
fun generateCarePlan(
    planDefinitionId: String,
    subject: String,
    encounterId: String?,
  ): IBaseResource {
```

signature is deprecated. I am not sure if we should change it at this point.

File: workflow/src/main/java/com/google/android/fhir/workflow/FhirOperator.kt (hunk 2 above)
Comment (MJ1998) [reply workflow/src/main/java/com/google/android/fhir/workflow/FhirOperator.kt:170]: I think SUBJECT overrides POPULATION (Practitioner)

```
MeasureEvalType evalType = (MeasureEvalType)MeasureEvalType.fromCode(reportType)
      .orElse(subjectIds.get(0) != null && subjectIds != null && !subjectIds.isEmpty() 
        ? MeasureEvalType.SUBJECT : MeasureEvalType.POPULATION);
```
Infact now I think it's always SUBJECT even when we give practitioner in subjectIds.
[PS: The check `subjectIds.get(0) != null` should be at last]

File: workflow/src/main/java/com/google/android/fhir/workflow/FhirOperator.kt [hunk 3]
```diff
@@ -162,20 +166,34 @@ internal constructor(
...
+        /* periodStart = */ start,
+        /* periodEnd = */ end,
+        /* reportType = */ reportType,
+        /* subjectIds = */ listOf(subject),
+        /* additionalData = */ null,
+      )
+
+    // add subject reference for non-individual reportTypes
+    if (report.type.name == MeasureReportType.SUMMARY.name && !subject.isNullOrBlank()) {
+      report.setSubject(Reference(subject))
```
MJ1998, line 194: I saw this function `evalTypeToReportType` in R4MeasureProcessor. Seems like MeasureEvalType.POPULATION is mapped to MeasureReportType.SUMMARY but for POPULATION, the subject is null.

Not sure if this function is even being used because I am not able to find any usage.

Also there are other non-individual types like SUBJECTLIST, PATIENTLIST as well .

What is intended here ?

workflow/src/main/java/com/google/android/fhir/workflow/FhirOperator.kt
vitorpamplona: Sure, having a subject will make the report a Subject type. That subject can then be a Patient, Group or the Practitioner (just another type of group). 

But if you have a practitioner AND a patient at the same time in this API, the report will be about the practitioner.

workflow/src/main/java/com/google/android/fhir/workflow/FhirOperator.kt
```diff
@@ -162,20 +166,34 @@ internal constructor(
...
+      }
+
+    val report =
+      measureProcessor.evaluateMeasure(
+        /* measure = */ Eithers.forLeft3(CanonicalType(measureUrl)),
+        /* periodStart = */ start,
+        /* periodEnd = */ end,
+        /* reportType = */ reportType,
+        /* subjectIds = */ listOf(subject),
+        /* additionalData = */ null,
```
MJ1998, line 189: Since now we changing the API, might as well add `additionalData` as an optional parameter in the API ?

vitorpamplona: My initial impression is that this should be given by the Engine module, but I am not sure if we will need it in the SDK.

workflow/src/main/java/com/google/android/fhir/workflow/FhirOperator.kt
vitorpamplona: > Seems like MeasureEvalType.POPULATION is mapped to MeasureReportType.SUMMARY but for POPULATION, the subject is null.

Correct, then the server-side folks set the reference in a separate class: 

https://github.com/cqframework/clinical-reasoning/blob/6ee71b792b1dc32ba9e077831dd89a5462077200/cqf-fhir-cr/src/main/java/org/opencds/cqf/fhir/cr/measure/r4/R4MeasureService.java

PATIENTLIST should not be a thing. If it is, it's some non-specified case. Details here: https://hapifhir.io/hapi-fhir/docs/clinical_reasoning/measures.html

workflow/src/main/java/com/google/android/fhir/workflow/FhirOperator.kt
MJ1998: Hence the practitioner check first! Cool.

workflow/src/main/java/com/google/android/fhir/workflow/FhirOperator.kt
MJ1998: My point is that report.type will be SUMMARY when subject is null. 
This if condition will never satisfy, right ?

workflow/src/main/java/com/google/android/fhir/workflow/FhirOperator.kt
MJ1998: My guess would be if this API will be available to be used independent of fhir engine then `additionalData` should also be an API parameter.

vitorpamplona: Cool, I just added it.

workflow/src/main/java/com/google/android/fhir/workflow/FhirOperator.kt
vitorpamplona: In theory yes, in practice users can force the `reportType` from params. I am not sure why they would do this, but it is certainly an option.

MJ1998: I think we should handle it!
Maybe just adding another require statement that both reportType and subjectId cannot be null ?

vitorpamplona: But both can be null at the same time.

MJ1998: oh yeah my bad. this looks good!

workflow/src/main/java/com/google/android/fhir/workflow/FhirOperator.kt
```diff
@@ -239,29 +253,83 @@ internal constructor(
...
+      /* parameters = */ parameters,
+      /* useServerData = */ useServerData,
+      /* bundle = */ bundle,
+      /* prefetchData = */ prefetchData,
       libraryProcessor,
     ) as IBaseResource
   }
 
+  @WorkerThread
+  fun generateCarePlan(
```
MJ1998, line 291: can we not have just 1 API to accept all three: id, canonical and plandefinition?

internally the plandefinitionprocessor does a "resolvePlanDefinition"

vitorpamplona: We could, but the knowledge manager is changing IDs on inserting so it's not ideal to call these functions by id (that's why it is deprecated). We could have one function with the other two. But this seems more explicit to the user that he/she has to choose one of them. It's a similar issue to the other params, but here we know how it should operate.

MJ1998: Fair enough.

MJ1998 (approved):
LGTM

MJ1998:
Also create an issue (if it does not exist) and tag it.

jingtang10 (approved):
thanks vitor!

vitorpamplona:
@jingtang10 can you enable auto-merge again here?

PR #2385, opened 2023-12-21
Synchronize sync workers

### LINKED ISSUE #2361: Concurrent Sync Worker execution can lead to data inconsistencies
**Describe the bug**
Recent changes in PR #2053 allow two sync workers to run concurrently.
This can lead to inconsistencies if a manual one-time sync and a periodic sync overlap.
The problem arises because both workers might attempt to upload resources in a conflicting order, potentially causing:
1. Upload of older resources before newer ones
2. Overwriting of data

**Component**
Fhir Engine Library

**To Reproduce**
Not sure how to reproduce as it requires concurrent running of the workers to sync resources that can cause inconsistency.

**Expected behavior**

**Screenshots**

**Additional context**

MJ1998:

Fixes #2361 

**Description**
We use the recommended way for mutual exclusion: [link](https://kotlinlang.org/docs/shared-mutable-state-and-concurrency.html#mutual-exclusion)

Solution: Wrap the `FhirSynchronizer#synchronize` with `mutex.withLock`

mutex.withLock will suspend the successive invocations of `FhirSynchronizer#synchronize`, without blocking the thread, until the lock is released. Hence all work requests will be processed - and in FIFO order.

**Alternative(s) considered**
Alternative: One alternative involves using a ReentrantLock within a try-catch-finally block. This approach essentially replicates mutex-like behavior to manage concurrent access to shared resources. It would require maintaining a job queue to handle work requests sequentially. Without a queue, workers would need to return Result.retry, leading to repeated attempts every 30 seconds.

Another alternative was to wrap `FhirSyncWorker#doWork` with mutex.withLock but based on [this](https://github.com/google/android-fhir/pull/2385#discussion_r1471034556) comment and also for the fact that FhirSynchronizer#synchronize should run sequentially irrespective of whether FhirSyncWorker uses it or not.

**Type**
Bug fix

**Checklist**


aditya-07 (changes requested):
Please add a test to check the change.

MJ1998:
I am not sure if automation test for this is a good idea. Because it takes lot of execution time to be able to verify the synchronicity of any concurrent jobs.  See how concurrent behavior is verified [here](https://medium.com/mobile-app-development-publication/mutex-for-coroutines-5f4a4ca60763).

Such long execution time, however, is very likely in case of long sync time in an actual app

@aditya-07

engine/src/main/java/com/google/android/fhir/sync/FhirSyncWorker.kt
```diff
@@ -60,74 +62,76 @@ abstract class FhirSyncWorker(appContext: Context, workerParams: WorkerParameter
...
-            else -> {
-              setProgress(buildWorkData(syncJobStatus))
             }
           }
         }
-      }
 
-    val result = synchronizer.synchronize()
-    val output = buildWorkData(result)
+      val result = synchronizer.synchronize()
```
jingtang10, line 117: i actually think the mutex can be just inside the synchronizer's synchronize function instead. in general we want to keep the locked code path as minimum as possible. and i think it's unnecessary to lock the rest of this function.

jingtang10 (approved):
please feel free to merge once tested manually again.

aditya-07 (approved):

MJ1998:
I figured how to test the mutex locking.
@aditya-07 @jingtang10

engine/src/main/java/com/google/android/fhir/sync/FhirSyncWorker.kt
MJ1998: Moved. Also added test.

engine/src/test/java/com/google/android/fhir/sync/FhirSynchronizerTest.kt
```diff
@@ -142,4 +146,36 @@ class FhirSynchronizerTest {
...
+              assertThat(e).isInstanceOf(IllegalStateException::class.java)
+              assertThat(e.localizedMessage).isNotNull()
+              assertThat(e.localizedMessage)
+                .isEqualTo("This mutex is already locked by the specified owner: $mutexOwner")
+            }
+          }
+        }
+      }
+
+      val result = fhirSynchronizer.synchronize()
```
aditya-07, line 179: The test should call `synchronize()` on 2 different `FhirSynchronizer` objects to simulate two workers getting started concurrently. 
Also, you should be able to provide Test downloader / uploader (with longer delay in first synchronizer) instead of mock to simulate a real scenario and use the `SyncJobStatus` maybe to assert that the second synchronize job doesn't start before the end of first synchronize job.

engine/src/test/java/com/google/android/fhir/sync/FhirSynchronizerTest.kt
```diff
@@ -142,4 +146,36 @@ class FhirSynchronizerTest {
...
+        fhirSynchronizer.syncState.collect {
+          if (it is SyncJobStatus.Started) {
+            try {
+              fhirSynchronizer.synchronize()
+            } catch (e: Exception) {
+              assertThat(e).isInstanceOf(IllegalStateException::class.java)
+              assertThat(e.localizedMessage).isNotNull()
+              assertThat(e.localizedMessage)
+                .isEqualTo("This mutex is already locked by the specified owner: $mutexOwner")
+            }
```
aditya-07, line 174: We shouldn't depend on the exception message as the verbiage may change in subsequent versions (which comes from kotlin framework) causing the test to fail.

aditya-07 (changes requested):

engine/src/test/java/com/google/android/fhir/sync/FhirSynchronizerTest.kt
MJ1998: Done.

engine/src/test/java/com/google/android/fhir/sync/FhirSynchronizerTest.kt
MJ1998: Changed the test.

aditya-07 (approved):

engine/src/main/java/com/google/android/fhir/sync/FhirSynchronizer.kt
```diff
@@ -79,18 +81,20 @@ internal class FhirSynchronizer(
...
+        .let {
+          if (it.isEmpty()) {
+            setSyncState(SyncResult.Success())
+          } else {
+            setSyncState(SyncResult.Error(it))
+          }
         }
-      }
+    }
   }
```
aditya-07, line 98: @MJ1998 Forgot to add this earlier, but could you please add doc to the function explaining its locking.

MJ1998: Added in #2417

PR #2395, opened 2024-01-01
Quantity unit options with initial quantity value

### LINKED ISSUE #2157: Date picker widget not working w calculated expression
**Describe the bug**
There is 2 issue to this:
1. The behavior_calculated_expression.json is not complete, missing unitOptions and the initial.valueQuantity.value, please use the JSON below to solve this first point
2. The Date picker widget in behavior_calculated_expression.json doesn't show the right value after it's calculated-expression was evaluated

After investigation, this code is the problem for point 2, it involves date formatting:
https://github.com/google/android-fhir/pull/1897/files#r1319542076

**Component**
SDC library and catalog app

**To Reproduce**
1. Catalog app
2. Click behavior tab
3. Click calculated expression
4. Fill in the Age years quantity widget
5. The date picker widget won't get updated

**Expected behavior**
The Date picker widget should be updated if the Age years quantity widget is filled

**Demo**

[Screen_recording_20230908_160436.webm](https://github.com/google/android-fhir/assets/62053304/f61189f2-e248-4f77-a56c-462fa773c83f)

**Additional context**
- I was attempting to fix this, but there's actually more to the issue (I'm referring to point 2):
- More context to the ticket https://github.com/google/android-fhir/pull/2132#discussion_r1319542407
- Use this complete behavior_calculated_expression.json
```
{
  "item": [
    {
      "linkId": "a-birthdate",
      "text": "Birth Date",
      "type": "date",
      "extension": [
        {
          "url": "http://hl7.org/fhir/uv/sdc/StructureDefinition/sdc-questionnaire-calculatedExpression",
          "valueExpression": {
            "language": "text/fhirpath",
            "expression": "%resource.repeat(item).where(linkId='a-age-years' and answer.empty().not()).select(today() - answer.value)"
          }
        }
      ]
    },
    {
      "linkId": "a-age-years",
      "text": "Age years",
      "type": "quantity",
      "extension": [
        {
          "url": "http://hl7.org/fhir/StructureDefinition/questionnaire-unitOption",
          "valueCoding": {
            "system": "http://unitsofmeasure.org",
            "code": "years",
            "display": "years"
          }
        }
      ],
      "initial": [
        {
          "valueQuantity": {
            "value": 0,
            "unit": "years",
            "system": "http://unitsofmeasure.org",
            "code": "years"
          }
        }
      ]
    }
  ]
}
```

I think @PallaviGanorkar is the right person for this

maimoonak:
Fixes #2157 

**Description**
As per fhir specs https://build.fhir.org/ig/HL7/sdc/behavior.html#initial an initial without decimal part of quantity type can be used for default unit reference purpose. The QuantityViewHolderFactory is now handling the unitoption with initial quantity used to specify default unit.

**Type**
Choose one: (Bug fix)

**Checklist**

### DIFF
--- a/catalog/src/main/assets/behavior_calculated_expression.json
+++ b/catalog/src/main/assets/behavior_calculated_expression.json
@@ -19,13 +19,34 @@
       "linkId": "a-age-years",
       "text": "Age years",
       "type": "quantity",
-      "initial": [{
-        "valueQuantity": {
-          "unit": "years",
-          "system": "http://unitsofmeasure.org",
-          "code": "years"
+      "extension": [
+        {
+          "url": "http://hl7.org/fhir/StructureDefinition/questionnaire-unitOption",
+          "valueCoding": {
+            "system": "http://unitsofmeasure.org",
+            "code": "years",
+            "display": "years"
+          }
+        },
+        {
+          "url": "http://hl7.org/fhir/StructureDefinition/questionnaire-unitOption",
+          "valueCoding": {
+            "system": "http://unitsofmeasure.org",
+            "code": "months",
+            "display": "months"
+          }
         }
-      }]
+      ],
+      "initial": [
+        {
+          "valueQuantity": {
+            "value": 1,
+            "unit": "years",
+            "system": "http://unitsofmeasure.org",
+            "code": "months"
+          }
+        }
+      ]
     }
   ]
 }
--- a/datacapture/src/main/java/com/google/android/fhir/datacapture/extensions/MoreQuestionnaireItemComponents.kt
+++ b/datacapture/src/main/java/com/google/android/fhir/datacapture/extensions/MoreQuestionnaireItemComponents.kt
@@ -1,5 +1,5 @@
 /*
- * Copyright 2023 Google LLC
+ * Copyright 2023-2024 Google LLC
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -581,6 +581,15 @@ internal val Questionnaire.QuestionnaireItemComponent.unitOption: List<Coding>
     return this.extension
       .filter { it.url == EXTENSION_QUESTIONNAIRE_UNIT_OPTION_URL }
       .map { it.value as Coding }
+      .ifEmpty {
+        // https://build.fhir.org/ig/HL7/sdc/behavior.html#initial
+        // quantity given as initial without value is for default unit reference purpose
+        if (this.hasInitial()) {
+          listOf(this.initialFirstRep.valueQuantity.toCoding())
+        } else {
+          listOf()
+        }
+      }
   }
 
 // ********************************************************************************************** //
--- a/datacapture/src/main/java/com/google/android/fhir/datacapture/extensions/MoreTypes.kt
+++ b/datacapture/src/main/java/com/google/android/fhir/datacapture/extensions/MoreTypes.kt
@@ -1,5 +1,5 @@
 /*
- * Copyright 2023 Google LLC
+ * Copyright 2023-2024 Google LLC
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -122,6 +122,11 @@ internal fun StringType.toIdType(): IdType {
 /** Converts Coding to CodeType. */
 internal fun Coding.toCodeType(): CodeType {
   return CodeType(code)
+}
+
+/** Converts Quantity to Coding. */
+internal fun Quantity.toCoding(): Coding {
+  return Coding(this.system, this.code, this.unit)
 }
 
 fun Type.valueOrCalculateValue(): Type {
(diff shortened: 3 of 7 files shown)


catalog/src/main/assets/behavior_calculated_expression.json
jingtang10, line 1: As I commented in the issue itself, is it possible to update this questionnaire so that we're calculating age from date of birth? I find it a little bit unintuitive to calculate date of birth from age... (perhaps I'm not familiar with the use case)

jingtang10: Or, perhaps it's easier just to use the example in the sdc ig: https://build.fhir.org/ig/HL7/sdc/examples.html#calculatedexpression

i quite like this example.

datacapture/src/main/java/com/google/android/fhir/datacapture/extensions/MoreQuestionnaireItemComponents.kt
```diff
@@ -581,6 +581,15 @@ internal val Questionnaire.QuestionnaireItemComponent.unitOption: List<Coding>
     return this.extension
       .filter { it.url == EXTENSION_QUESTIONNAIRE_UNIT_OPTION_URL }
       .map { it.value as Coding }
+      .ifEmpty {
+        // https://build.fhir.org/ig/HL7/sdc/behavior.html#initial
```
jingtang10, line 585: i think there're two things in the specification that are open to interpretation:

1. do we consider multiple intial values? what if multiple initial values have the same unit? i think in this case we should deduplicate. please do that in the code.
2. the text in sdc ig only talks about the case where the initial values don't have the numerical value. but what happens if the initial values do have numerical values as well as unit? in this case, it does seem to make sense to use the units as well so that at least these initial values can be represented.
3. what if the initial values include units that do not appear in the unit option extension? again, i feel like maybe in this case the easiest thing to do is to include the units in the initial values. yes, we can throw an exception, but that seems very unforgiving.

==> if we consider point 2 and 3 above together, perhaps an easier way to write the code is just to join the unit options with all the units in the initial values and then deduplicate them. can you do that?

datacapture/src/main/java/com/google/android/fhir/datacapture/extensions/MoreTypes.kt
```diff
@@ -124,6 +124,11 @@ internal fun Coding.toCodeType(): CodeType {
   return CodeType(code)
 }
 
+/** Converts Quantity to Coding. */
```
jingtang10, line 127: i think this is an example of how comments can be made more useful :)

as it is, i would actually remove the comment. "converts quantity to coding" doesn't actually give me any more information than the function signature `Quantity.toCoding()` which is very straightforward and clear already.

but there's more to this function, it is not a lossless conversion - it's losing information as we're losing the value field of the quantity value. so warning developers that this is not really a conversion but rather getting the unit from the Quantity value would be useful.

actually now having thought about this, i feel we should rename the function to `getUnit`. what do you think?

datacapture/src/main/java/com/google/android/fhir/datacapture/views/factories/QuantityViewHolderFactory.kt
```diff
@@ -189,6 +191,11 @@ internal object QuantityViewHolderFactory :
             Coding(it.system, it.code, it.unit)
```
jingtang10, line 191: should this use your new function?

datacapture/src/main/java/com/google/android/fhir/datacapture/views/factories/QuantityViewHolderFactory.kt
```diff
@@ -189,6 +191,11 @@ internal object QuantityViewHolderFactory :
...
           }
             ?: questionnaireViewItem.draftAnswer?.let { if (it is Coding) it else null }
+              ?: questionnaireViewItem.questionnaireItem
+              .takeIf { it.hasInitial() }
+              ?.initialFirstRep
```
jingtang10, line 196: shouldn't this just be 

```suggestion
              .initial
              ?.firstOrNull()
```

jingtang10: or shouldn't you just call `questionnaireViewItem.questionnaireItem.unitOption.firstOrNull()`

jingtang10 (approved):

jingtang10 (changes requested):
oops, approved by mistake. please address previous comments!

thanks @maimoonak

catalog/src/main/assets/behavior_calculated_expression.json
maimoonak: The calculation of Birthdate from Age is a very common usecase of underdeveloped countries projects, when mother or patient itself donot remember their birthdate, but system needs an estimated birthdate so age is used to get an estimated birthdate. I have changed the text a bit to make it more clear. 

I tried the example as well but it throws exception with power and round function
```
 org.hl7.fhir.exceptions.PathEngineException: Error evaluating FHIRPath expression: focus for 0 can only have one value, but has 0 values (@char 1)
                  at org.hl7.fhir.r4.utils.FHIRPathEngine.makeExceptionPlural(FHIRPathEngine.java:629)
                  at org.hl7.fhir.r4.utils.FHIRPathEngine.funcPower(FHIRPathEngine.java:3866)
                  at org.hl7.fhir.r4.utils.FHIRPathEngine.evaluateFunction(FHIRPathEngine.java:3692)
                  at org.hl7.fhir.r4.utils.FHIRPathEngine.execute(FHIRPathEngine.java:1488)
                  
```

jingtang10 (approved):

PR #2396, opened 2024-01-01
Questionnaire response reference for questionnaire id or url validation

### LINKED ISSUE #1713: QuestionnaireResponse Validator should not throw exception when questionnaire url is null
**Describe the bug**
QuestionnaireResponse validator [asserts condition](https://github.com/google/android-fhir/blob/master/datacapture/src/main/java/com/google/android/fhir/datacapture/validation/QuestionnaireResponseValidator.kt#L61) where it is expected that questionnaire.url = questionnaireResonse.questionnaire. This condition is not covering all use cases and throws exception needlessly where this is not true. A questionnaire.url can be null (most of questionnaire have do not have any url), it could be something like http://my-fhir-server/fhir.com/Questionnaire/1234, Or it could be simply Questionnaire/12345, etc. The validation should not force this condition unless all cases are covered.

**Component**
SDC library

**To Reproduce**
1. Any questionnaire without url
2. Set QuestionnaireResonse.questionnaire = Questionnaire/q-id
3. Submit
4. IllegalArgument exception as questionnaire.url is null but QuestionnaireResonse.questionnaire has a valid value

**Expected behavior**
Should not throw exception if questionnaire.url is nul

maimoonak:
Fixes #1713 

**Description**
The questionnaire-response references questionnaire for which it is being filled. The value of this field must either be null OR match the questionnaire id or url property. 

**Type**
Choose one: (Bug fix)

**Checklist**


datacapture/src/main/java/com/google/android/fhir/datacapture/validation/QuestionnaireResponseValidator.kt
```diff
@@ -63,9 +63,11 @@ object QuestionnaireResponseValidator {
   ): Map<String, List<ValidationResult>> {
     require(
       questionnaireResponse.questionnaire == null ||
-        questionnaire.url == questionnaireResponse.questionnaire,
+        questionnaireResponse.questionnaire == questionnaire.url ||
+        questionnaireResponse.questionnaire == questionnaire.id ||
+        questionnaireResponse.questionnaire.endsWith("/${questionnaire.id}"),
```
jingtang10, line 68: i'm not sure about this last condition, shouldn't that be covered by the url case?

if we're not sure about this can we clarify this one chat.fhir.org?

jingtang10 (changes requested):

jingtang10:
Reading this again: https://hl7.org/fhir/R4/questionnaire-definitions.html#Questionnaire.url I'm really not sure if this PR is correct in the first place. Please refer to the discussion in the linked issue https://github.com/google/android-fhir/issues/1713#issuecomment-1892447824

PR #2442, opened 2024-02-15
Updated PerResourcePatchGenerator  to return ordered PatchMapping to avoid referential integrity issues.

aditya-07:

Fixes #[issue number]

**Description**
1. Create a Directed Graph based on all the outgoing references for each Patch. Since the referential integrity comes to picture if the resource is not created yet, the edges of the graph are only the outgoing references to resources that are getting created during this upload. This may helps avoid some cycles as well.
2. The references are taken from `LocalChangeResourceReferenceEntity` which stores each outgoing references associated with a particular `LocalChange` as new entry.
3. Order the Patches so that if the resource A has outgoing references {B,C} (CREATE) and {D} (UPDATE), then B,C needs to go before the resource A so that referential integrity is retained. Order of D shouldn't matter for the purpose of referential integrity.

**Alternative(s) considered**
1. Splitting changes to CREATE first and subsequent UPDATE, but this would also require ordering the the create list incase it can overflow to multiple Bundles.  This would be similar to the current approach but, would require multiple PUSH for the same resource incase the resource was Created and then Updated on the device before being pushed to the server.

**Type**
Choose one: Bug fix 

**Checklist**


MJ1998 (changes requested):
I feel we can improve complexity of this solution by using LocalChangeReferenceEntity table to fetch all the outgoing references.

engine/src/main/java/com/google/android/fhir/sync/upload/patch/PerResourcePatchGenerator.kt
```diff
@@ -139,3 +143,97 @@ internal object PerResourcePatchGenerator : PatchGenerator {
...
+    )
+  return createTopologicalOrderedList(adjacencyList).mapNotNull { resourceIdToPatchMapping[it] }
+}
+
+/**
+ * @return A map of [PatchMapping] to all the outgoing references to the other [PatchMapping]s of
+ *   type [Patch.Type.INSERT] .
+ */
+@androidx.annotation.VisibleForTesting
+internal fun List<PatchMapping>.createReferenceAdjacencyList(
```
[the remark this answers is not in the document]
aditya-07: Done

engine/src/main/java/com/google/android/fhir/sync/upload/patch/PerResourcePatchGenerator.kt
```diff
@@ -139,3 +143,97 @@ internal object PerResourcePatchGenerator : PatchGenerator {
...
+// to it. This may make the sub graphs smaller and also help avoid cyclic dependencies.
+@androidx.annotation.VisibleForTesting
+internal fun PatchMapping.findOutgoingReferences(): List<Node> {
+  val references = mutableListOf<Node>()
+  when (generatedPatch.type) {
+    Patch.Type.INSERT -> {
+      JsonLoader.fromString(generatedPatch.payload).search(references)
+    }
+    Patch.Type.UPDATE -> {
+      JsonLoader.fromString(generatedPatch.payload).forEach {
```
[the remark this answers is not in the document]
aditya-07: Yes, I have removed this code .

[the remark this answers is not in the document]
aditya-07: N.A

MJ1998 (commented):
Nice job stitching the `LocalChangeResourceReferenceEntity` here. 

Although I dont like PatchGenerator having access to database which is from another package. I think we should abstract this out in a separate interface which is parent of the two packages: db and sync. 
Maybe FhirSyncDbInteractor works well in this case as well. Will raise a PR soon so that this can be imagined better.

engine/src/main/java/com/google/android/fhir/sync/upload/patch/PatchGenerator.kt
```diff
@@ -34,16 +35,17 @@ internal interface PatchGenerator {
...
 
 internal object PatchGeneratorFactory {
   fun byMode(
     mode: PatchGeneratorMode,
+    database: Database,
   ): PatchGenerator =
     when (mode) {
       is PatchGeneratorMode.PerChange -> PerChangePatchGenerator
-      is PatchGeneratorMode.PerResource -> PerResourcePatchGenerator
+      is PatchGeneratorMode.PerResource -> PerResourcePatchGenerator(database)
```
MJ1998, line 48: This will create a new instance every time as contrary to earlier where the object was a singleton. And singleton is what I would prefer because the upload strategy wont change.

MJ1998 (changes requested):

engine/src/main/java/com/google/android/fhir/sync/upload/patch/PerResourcePatchGenerator.kt
```diff
@@ -139,3 +145,100 @@ internal object PerResourcePatchGenerator : PatchGenerator {
...
+
+private typealias Node = String
+
+/**
+ * @return - A ordered list of the [PatchMapping]s based on the references to other [PatchMapping]
+ *   if the mappings are acyclic
+ * - throws [IllegalStateException] otherwise
+ */
+@androidx.annotation.VisibleForTesting
+internal suspend fun List<PatchMapping>.orderByReferences(database: Database): List<PatchMapping> {
```
jingtang10, line 157: if you add a reference in a resource that you later remove, by this method you'd have a dependency. but in reality there isn't after squashing... how do you handle that?

jingtang10 (changes requested):

engine/src/main/java/com/google/android/fhir/sync/upload/patch/PerResourcePatchGenerator.kt
aditya-07: This function is called on the squashed `PatchMapping` and as such only finds the references that are present in it.

engine/src/main/java/com/google/android/fhir/sync/upload/patch/PatchOrdering.kt
```diff
@@ -0,0 +1,124 @@
...
+   * - throws [IllegalStateException] otherwise
+   */
+  @VisibleForTesting(otherwise = VisibleForTesting.PROTECTED)
+  internal suspend fun List<PatchMapping>.orderByReferences(
+    database: Database,
+  ): List<PatchMapping> {
+    // if the resource A has outgoing references (B,C) and these referenced resources are getting
+    // created on device,
+    // then these referenced resources (B,C) needs to go before the resource A so that referential
+    // integrity is retained.
```
jingtang10, line 40: this needs to go into the function doc

engine/src/main/java/com/google/android/fhir/sync/upload/patch/PatchOrdering.kt
```diff
@@ -0,0 +1,124 @@
...
+
+package com.google.android.fhir.sync.upload.patch
+
+import androidx.annotation.VisibleForTesting
+import com.google.android.fhir.db.Database
+import com.google.android.fhir.db.LocalChangeResourceReference
+
+private typealias Node = String
+
+/** Orders the [PatchMapping]s to maintain referential integrity during upload. */
```
jingtang10, line 25: needs examples

engine/src/main/java/com/google/android/fhir/sync/upload/patch/PatchOrdering.kt
```diff
@@ -0,0 +1,124 @@
...
+    return createTopologicalOrderedList(adjacencyList).mapNotNull { resourceIdToPatchMapping[it] }
+  }
+
+  /**
+   * @return A map of [PatchMapping] to all the outgoing references to the other [PatchMapping]s of
+   *   type [Patch.Type.INSERT] .
+   */
+  @VisibleForTesting
+  internal fun List<PatchMapping>.createAdjacencyListForCreateReferences(
+    insertResourceIds: Set<String>,
```
jingtang10, line 67: can you try to remove this as an argument - and just calculate this inside this function? does it make it look better?

engine/src/main/java/com/google/android/fhir/sync/upload/patch/PatchOrdering.kt
```diff
@@ -0,0 +1,124 @@
...
+      )
+    return createTopologicalOrderedList(adjacencyList).mapNotNull { resourceIdToPatchMapping[it] }
+  }
+
+  /**
+   * @return A map of [PatchMapping] to all the outgoing references to the other [PatchMapping]s of
+   *   type [Patch.Type.INSERT] .
+   */
+  @VisibleForTesting
+  internal fun List<PatchMapping>.createAdjacencyListForCreateReferences(
```
jingtang10, line 66: i wonder if it looks better without this being an extension function - please just give it a try.

engine/src/test/java/com/google/android/fhir/sync/upload/patch/PatchOrderingTest.kt
```diff
@@ -0,0 +1,825 @@
...
+            type = Group.GroupType.PERSON
+          }
+          .also { localChanges.add(createInsertLocalChange(it, Random.nextLong())) }
+
+      // pa-01
+      Patient()
+        .apply { id = "patient-1" }
+        .also {
+          localChanges.add(
+            createUpdateLocalChange(
```
jingtang10, line 303: maybe make this a bit more prominent - feel like this is the crux of this test case.

engine/src/test/java/com/google/android/fhir/sync/upload/patch/PatchOrderingTest.kt
jingtang10, line 1: can you consider simplifying the test cases a little bit? are there shorter test cases that you can write that would give you the same coverage?

engine/src/test/java/com/google/android/fhir/sync/upload/patch/PatchOrderingTest.kt
```diff
@@ -0,0 +1,825 @@
...
+        "patient-1",
+        "patient-2",
+        "patient-3",
+        "group-1",
+        "encounter-1",
+        "observation-1",
+        "encounter-2",
+        "observation-2",
+        "encounter-3",
+        "observation-3",
```
jingtang10, line 740: i'm fine with this. but maybe add a comment to say that this order is so only because we know the implementation of the topological sort - it's entirely possible to generate a different order here which is acceptable/correct, should we have a different implementation of the topological sort.

jingtang10 (approved):
thanks aditya! great job!

engine/src/main/java/com/google/android/fhir/sync/upload/patch/PatchOrdering.kt
```diff
@@ -0,0 +1,124 @@
...
+        patchMapping.findOutgoingReferences(localChangeIdToReferenceMap).filter {
+          insertResourceIds.contains(it)
+        }
+    }
+    return adjacencyList
+  }
+
+  // if the outgoing reference is to a resource that's just an update and not create, then don't
+  // link
+  // to it. This may make the sub graphs smaller and also help avoid cyclic dependencies.
```
MJ1998, line 84: NIT: Encapsulate this comment within /** */

engine/src/main/java/com/google/android/fhir/sync/upload/patch/PerResourcePatchGenerator.kt
```diff
@@ -138,4 +145,20 @@ internal object PerResourcePatchGenerator : PatchGenerator {
     mergedOperations.values.flatten().forEach(mergedNode::add)
     return objectMapper.writeValueAsString(mergedNode)
   }
+
+  companion object {
+
+    private lateinit var _instance: PerResourcePatchGenerator
```
MJ1998, line 151: I think this should be marked @Volatile
[source](https://www.baeldung.com/kotlin/singleton-classes#1-companion-object)

aditya-07: Done

engine/src/main/java/com/google/android/fhir/sync/upload/patch/PatchOrdering.kt
aditya-07: done

engine/src/test/java/com/google/android/fhir/sync/upload/patch/PatchOrderingTest.kt
aditya-07: Done

engine/src/test/java/com/google/android/fhir/sync/upload/patch/PatchOrderingTest.kt
aditya-07: Added a Utility to help create the LocalChange and LocalChangeResourceReference for various resources and operations (Create & Update). This has simplified reading the test.

engine/src/test/java/com/google/android/fhir/sync/upload/patch/PatchOrderingTest.kt
aditya-07: Done

engine/src/main/java/com/google/android/fhir/sync/upload/patch/PatchOrdering.kt
aditya-07: Kept it as extension but removed the extra function argument.

engine/src/main/java/com/google/android/fhir/sync/upload/patch/PatchOrdering.kt
aditya-07: Done

engine/src/main/java/com/google/android/fhir/sync/upload/patch/PatchOrdering.kt
aditya-07: Done

engine/src/main/java/com/google/android/fhir/sync/upload/patch/PatchOrdering.kt
aditya-07: Done

engine/src/test/java/com/google/android/fhir/sync/upload/patch/PatchOrderingTest.kt
```diff
@@ -0,0 +1,825 @@
...
+      val localChanges = LinkedList<LocalChange>()
+      val localChangeResourceReferences = mutableListOf<LocalChangeResourceReference>()
+
+      var group =
+        Group()
+          .apply {
+            id = "group-1"
+            type = Group.GroupType.PERSON
+          }
+          .also { localChanges.add(createInsertLocalChange(it, Random.nextLong())) }
```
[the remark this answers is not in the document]
aditya-07: Done

engine/src/main/java/com/google/android/fhir/sync/upload/patch/PatchOrdering.kt
```diff
@@ -0,0 +1,124 @@
...
+package com.google.android.fhir.sync.upload.patch
+
+import androidx.annotation.VisibleForTesting
+import com.google.android.fhir.db.Database
+import com.google.android.fhir.db.LocalChangeResourceReference
+
+private typealias Node = String
+
+/** Orders the [PatchMapping]s to maintain referential integrity during upload. */
+object PatchOrdering {
```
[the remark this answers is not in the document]
aditya-07: Done

engine/src/main/java/com/google/android/fhir/sync/upload/patch/PatchOrdering.kt
```diff
@@ -0,0 +1,124 @@
...
+
+  /**
+   * @return - A ordered list of the [PatchMapping]s based on the references to other [PatchMapping]
+   *   if the mappings are acyclic
+   * - throws [IllegalStateException] otherwise
+   */
+  @VisibleForTesting(otherwise = VisibleForTesting.PROTECTED)
```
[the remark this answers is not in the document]
aditya-07: Done

MJ1998 (commented):
I think we can have a better design:-
Crux: Instead of fetching LocalChanges we will fetch LocalChange.id as we don't need full LocalChange objects to create a toposort of Patches.

Steps:-

1. **fetchLocalChangeIds**(): List\<LocalChangeId\>
2. **orderByResourceTypeAndId**(): Map\<ResourceTypeWithId, List\<LocalChangeId\>\>
[ referencing ResourceTypeWithId as RTWI, and  List\<LocalChangeId\> as IDPatch ]
So orderByResourceTypeAndId(): Map\<RTWI, IDPatch\>
3. **buildTopoSort**(): List\<Pair\<RTWI, IDPatch\>\> (use database.getLocalChangeResourceReferenceIds)
4. **Process** this list based on UploadStrategy. For "single-resource" strategy process one by one, for "all-local-changes" process all of them at once. process = fetch whole of local changes: Flow\<List\<LocalChange\>\>
5. **patchGenerator**(List\<LocalChange\>\>): List\<Patch\>
6. **uploadRequestGenerator**(List\<Patch\>): List\<UploadRequest\>

Pros:-
This will significantly reduce the memory consumed by delaying the fetching of full LocalChange objects.
Supports both upload strategies.

Con:-
One caveat could be double database fetching. But fetching just the LocalChangeIds should not take much time as its LocalChangeEntity's PrimaryKey.

I think given we want to support offline, memory is a more concerning issue than time complexity to sync here. But without profiling the two designs agains a db it's difficult to say.

Also, maybe we can take this up as a new issue and a new PR.

jingtang10:
> I think we can have a better design:- Crux: Instead of fetching LocalChanges we will fetch LocalChange.id as we don't need full LocalChange objects to create a toposort of Patches.
> 
> Steps:-
> 
> 1. fetchLocalChangeIds(): List<LocalChangeId>
> 2. orderByResourceTypeAndId(): Map<ResourceTypeWithId, List<LocalChangeId>>
>    set alias RTWI = ResourceTypeWithId
>    set alias IDPatch = List<LocalChangeId>
>    So orderByResourceTypeAndId(): Map<RTWI, IDPatch>
> 3. buildTopoSort(): List<Pair<RTWI, IDPatch>> (use database.getLocalChangeResourceReferenceIds)
> 4. Process this list based on UploadStrategy. For "single-resource" strategy process one by one, for "all-local-changes" process all of them at once. process = fetch whole of local changes: Flow<List<LocalChange>>
>    Now our old approach
> 5. patchGenerator(List<LocalChange>>): List<Patch>
> 6. uploadRequestGenerator(List<Patch>): List<UploadRequest>
> 
> Pros:- This will significantly reduce the memory consumed by delaying the fetching of full LocalChange objects. Supports both upload strategies.
> 
> Con:- One caveat could be double database fetching. But fetching just the LocalChangeIds should not take much time as its LocalChangeEntity's PrimaryKey.
> 
> I think given we want to support offline, memory is a more concerning issue than time complexity to sync here. But without profiling the two designs agains a db it's difficult to say.
> 
> Also, maybe we can take this up as a new issue and a new PR.

thanks for this suggestion jajoo! i'd like to understand a bit better the exact savings of this approach. but yes we should do this in a separate issue.

jingtang10 (approved):

jingtang10:
looks good - if jajoo is ok with this you can merge this aditya!

thanks!

MJ1998 (commented):
LGTM

MJ1998 (approved):

engine/src/main/java/com/google/android/fhir/db/impl/dao/LocalChangeDao.kt
```diff
@@ -348,6 +348,17 @@ internal abstract class LocalChangeDao {
...
+    """
+        SELECT *
+        FROM LocalChangeResourceReferenceEntity
+        WHERE localChangeId = (:localChangeId)
+    """,
+  )
+  abstract suspend fun getReferencesForLocalChanges(
+    localChangeId: List<Long>,
+  ): List<LocalChangeResourceReferenceEntity>
+
```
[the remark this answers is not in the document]
santosh-pingle: Same here.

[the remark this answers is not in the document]
santosh-pingle: I unblock myself by updating given change to 
```
SELECT *
        FROM LocalChangeResourceReferenceEntity
        WHERE localChangeId IN (:localChangeId)
```

PR #2462, opened 2024-03-05
Update the purge API to accept a list of resource IDs

### LINKED ISSUE #2439: Update the purge API to accept a list of resource IDs
- We would like to be able to purge resources in bulk selectively.  This would help implement functionality to delete data tied to certain jurisdictions if a user stops being responsible for that jurisdiction. 
- Example: - A practitioner is assigned to collect data for Nairobi but is then transferred to Mombasa. We would like to have the ability to purge the Nairobi "patient" data. 
- The current purge API expects 1 resource ID at a time. This ticket is aimed at updating that to expect a list of IDs

- [ ] The suggested solution by @aditya-07  is to combine the Search and Purge API i.e. search for the resources to delete using the relevant resource search parameters then pass the result result resource IDs to the purge API for deletion.

ndegwamartin:

Fixes #2439 

**Description**
1. Overloaded the `FhirEngine.purge` and modified `Database.purge` since has access modifier `internal` which is reused by the FhirEngine overloads.
2. In the implementation of `Database.purge`, wrapped the processing of the resource ids list in a transaction. Either they all pass or it fails.

**Alternative(s) considered**

**Type**
Feature

N/A

**Checklist**


vorburger:
@ndegwamartin Thank You for requesting a review from me!

As I only "volunteer" and do not "work for" this project, unfortunately I do not have the "free cycles" to help here.

I trust that some of my colleagues who are working on this project will review this.

Thank You for your understanding!

engine/src/main/java/com/google/android/fhir/impl/FhirEngineImpl.kt
```diff
@@ -81,7 +81,11 @@ internal class FhirEngineImpl(private val database: Database, private val contex
   }
 
   override suspend fun purge(type: ResourceType, id: String, forcePurge: Boolean) {
-    database.purge(type, id, forcePurge)
+    database.purge(type, setOf(id), forcePurge)
+  }
+
+  override suspend fun purge(type: ResourceType, ids: Set<String>, forcePurge: Boolean) {
+    database.purge(type, ids, forcePurge)
```
MJ1998, line 88: Should this fail, I think we should report to the application "why".
Maybe propagate the transaction failure with a proper message ?

engine/src/main/java/com/google/android/fhir/FhirEngine.kt
```diff
@@ -112,12 +112,25 @@ interface FhirEngine {
...
+   * Purges a resource from the database based on resource type and id without any deletion of data
+   * from the server.
+   *
+   * @param type The [ResourceType]
+   * @param ids The resource ids [Set]<[Resource.id]>
+   * @param forcePurge default value is false, here resources will not be deleted from
+   *   LocalChangeEntity table but it will throw IllegalStateException("Resource has local changes
+   *   either sync with server or FORCE_PURGE required") if local changes exists. If true this API
+   *   will delete resource entries from LocalChangeEntity table.
+   */
```
MJ1998, line 132: Add kdoc to the base function about what happens when purging of the whole list fails.

MJ1998 (changes requested):

engine/src/main/java/com/google/android/fhir/impl/FhirEngineImpl.kt
ndegwamartin: This can fail for a number of reasons:
1. The Resource in question is not found in the DB 
2. The Resource has ref. entries in the local changes table and FORCE_PURGE is false
3. Some other (system) issue

For the first two we throw the exceptions explicitly for each resource processed

ndegwamartin: After the conversation today we've agreed to wrap the whole batch list processing logic in a transaction. Thus if any item fails it throws an `IllegalStateException`. 

Left [the same comment on ticket #2439](https://github.com/google/android-fhir/issues/2439#issuecomment-2036718710) as well

engine/src/main/java/com/google/android/fhir/FhirEngine.kt
```diff
@@ -192,8 +192,26 @@ interface FhirEngine {
...
    *   Otherwise, an [IllegalStateException] will be thrown if local changes exist. Defaults to
    *   `false`.
+   *
+   *   If you need to purge resources in bulk use the method
+   *   [FhirEngine.purge(type: ResourceType, ids: Set<String>, forcePurge: Boolean = false)]
    */
   suspend fun purge(type: ResourceType, id: String, forcePurge: Boolean = false)
+
+  /**
+   * Purges a resource from the database based on resource type and id without any deletion of data
```
MJ1998, line 202: Can you modify it to suit set of ids. Maybe:-
`Purges resources of the specified type, identified by their IDs.`

engine/src/main/java/com/google/android/fhir/db/Database.kt
```diff
@@ -174,17 +174,17 @@ internal interface Database {
   suspend fun getLocalChanges(resourceUuid: UUID): List<LocalChange>
 
   /**
-   * Purge resource from database based on resource type and id without any deletion of data from
+   * Purge resources from database based on resource type and id without any deletion of data from
```
MJ1998, line 177: Similarly here.

MJ1998 (approved):
There is one scenario which we never covered. What happens in case we purge a resource with forcePurge set to true but there is same resource reference in another LocalChange - referential integrity broken. 
For example when application does a purge Patient of given id(s) but not Observation. 

We can either:-
1. Support reference resource recursive purging with another flag.
2. Throw an exception when there are references which were not purged ?

This can be taken up in a new PR. Could you create an issue for the same?  thanks.

Nit comments. 
LGTM.

ndegwamartin:
@MJ1998 ticket created there https://github.com/google/android-fhir/issues/2510

ndegwamartin:
@MJ1998 when you get a chance please assist in merging this

Issue #2481, opened 2024-03-15
Support GitHub Codespaces


vorburger:
**Is your feature request related to a problem? Please describe.**
In order to be able to test e.g. #2463 for #2237, as well as to be able to work on #2232, it would be much easier, at least for me, if I didn't need to have a computer with the Android Studio with the Android SDK etc. set up. (I originally used to have it set up, but changed computers since.)

**Describe the solution you'd like**
A great solution for this are GitHub's Codespaces, based on Dev Containers.

**Describe alternatives you've considered**
There are none.

**Would you like to work on the issue?**
Yes, I'll explore how hard it would be to make at least a very basic initial version available.

zashab:
Tiktoken
https://learnxinyminutes.com/docs/yaml/

PR #2482, opened 2024-03-15
Initial GitHub Codespaces (AKA Dev Container) configuration (see #2481)

vorburger:
see #2481

@jingtang10 & @santosh-pingle how about this?


jingtang10 (approved):
whilst i think this is a great change, i think we need to document a little better in our codebase why we're using this and how. especially how other developers (and developers in our wider community) can benefit from this. @vorburger please feel free to merge this. but can you think about adding a bit more documentation either in here or as an immediate follow-up? thanks :)

PR #2483, opened 2024-03-15
Generate API docs site during build

vorburger:
Part of #2232.

Enabled by #2463, this generates the Kotlin API docs on https://google.github.io/android-fhir/ during the CI build. (They were "static" before, and just deployed from what was in the `docs/` directory, with this, they are "built". This is the 1st step towards generating more documentation during the build in a future next step.)

Originally raised as #2483, but that PR for some strange reason didn't update even though I (force) pushed new commit.

@williamito or @santosh-pingle or @MJ1998 or @jingtang10  OK for you?

~Includes #2463, but just for testing; when this works, I'll merge that, and then rebase this to not include that anymore.~

~I'm marking this one as draft as I'm still testing this - please do not merge as long as it's in Draft.~

Issue #2232, opened 2023-10-06
Move Technical Documentation from Wiki to MD in /docs/ and publish new GitHub Pages based Site by CI with Actions


vorburger:

vorburger:
@jingtang10 with #2232 merged (TX), I've taken the liberty of already adding the (current) site from /Docs/ to the repo homepage (on the right, under the About, above the tags).

vorburger:
![Screenshot_20231008-184506~2](https://github.com/google/android-fhir/assets/298598/e6456cd6-4cc4-40c5-9521-012ce9a0a188)

PR #2484, opened 2024-03-16
Remove docs/ API doc

vorburger:
It's removed because this will be generated during the build with #2483, as part of #2232.

@jingtang10 OK for you?


vorburger:
@jingtang10 or am I misunderstanding and you would want to keep these API docs from older releases? We could also do that, but if so, I need to move them.

williamito (approved):

jingtang10 (approved):
thanks so much for this!

jingtang10:
> @jingtang10 or am I misunderstanding and you would want to keep these API docs from older releases? We could also do that, but if so, I need to move them.

no this is great

Issue #2485, opened 2024-03-16
Flaky QuestionnaireUiEspressoTest$dateTimePicker_shouldSetAnswerWhenDateAndTimeAreFilled has 12/24h problem


vorburger:
**Describe the bug**

A new flaky test is slowing down development velocity on unrelated work; [this](https://firebase.corp.google.com/u/0/project/android-fhir-instrumeted-tests/testlab/histories/bh.96901a7d2453dacf/matrices/6289670456786294467/executions/bs.6edd147739902de3/testcases/7) and [this](https://firebase.corp.google.com/u/0/project/android-fhir-instrumeted-tests/testlab/histories/bh.96901a7d2453dacf/matrices/6289670456786294467/executions/bs.f9011cdb2a7c5fc8/testcases/8)
 ran into:

`
expected: 2005-01-05T06:10
but was : 2005-01-05T18:10
at com.google.android.fhir.datacapture.test.QuestionnaireUiEspressoTest$dateTimePicker_shouldSetAnswerWhenDateAndTimeAreFilled$1.invokeSuspend(QuestionnaireUiEspressoTest.kt:255)
`

This has been noticed e.g. in #2471, and it's now also blocking merging #2463 to fix #2237, which in turn is holding up #2483.

**Component**
Datacapture

**To Reproduce**
Raise a PR and it Kokoro fails due to something unrelated to what you are working on .

**Expected behavior**
Tests which are unrelated to the current change must always pass.

**Would you like to work on the issue?**
I'll see if I propose a PR to disable (ignore) that test.

I don't want to work on fixing that test - but perhaps someone else is interesting in picking this up?

jingtang10:
thanks for raising this @vorburger

PR #2486, opened 2024-03-16
Ignore flaky dateTimePicker_shouldSetAnswerWhenDateAndTimeAreFilled

vorburger:
re. #2485


vorburger:
#2487 may be better than this - if that builds.

vorburger:
#2489 was a better Fix for that problem than this, so closing this in favour of that one (which I've just merged).

PR #2511, opened 2024-04-16
Update currentSyncJobStatus for oneTimeSync when syncJobStatus is null

### LINKED ISSUE #2472: Subsequent OneTime sync triggers always start with CurrentSyncJobStatus.SUCCEDED status
**Describe the bug**
When One time sync is triggered again, the state is first retrieved from the `syncJobStatusFromWorkManager`(screenshot). A null check is performed and if the state returned is null the [state is retrieved from the sync job status datastore](https://github.com/google/android-fhir/blob/9697a8a01283f38a9c2b20e8e3608cfc00bd9e7e/engine/src/main/java/com/google/android/fhir/sync/Sync.kt#L276). This will hold the status of the previous one time sync.

This means that on subsequent one time syncs, we have no way of capturing the state that a new sync has started. This is important so that for instance we can show a UI that shows _Sync process is starting_. The main issue with the UX is that every time we trigger it subsequently, it first  shows a **Sync complete** dialog then it proceds to do the actual sync after which it shows the _sync complete dialog_ again. 

**Component**
Core library

**To Reproduce**
- Trigger a one time sync multiple times

**Expected behavior**
We should always return a job status like `Running.Started` or `Enqueued` for every fresh onetime sync trigger in order to update UX accordingly.

**Screenshots**
Screenshot 2024-03-12 at 12 15 46

N/A

**Additional context**

Sure, however @santosh-pingle recently worked on the original feature

santosh-pingle:

Fixes #2472

**Description**
WorkManager returns null status for ENQUEUED, RUNNING, and CANCELLED states. Now, if syncJobStatus is null, CurrentSyncJobStatus is created based on the WorkInfo state. If the WorkInfo state is SUCCEEDED or FAILED, then the sync job status is fetched from the FHIR data store.

**Alternative(s) considered**

**Type**
Choose one: Bug fix

**Checklist**


ndegwamartin:
@santosh-pingle tested out the implementation, it works fine on our app and solves the issue we reported.

joiskash (approved):
Looks good, just left a small correction in documentation

ndegwamartin:
@jingtang10 anything left for this PR or can we merge it?

engine/src/main/java/com/google/android/fhir/sync/Sync.kt
jingtang10, line 1: can you remove the function:
```
private fun handleNullWorkManagerStatusForPeriodicSync(
    workInfoState: WorkInfo.State,
  ): CurrentSyncJobStatus
``` 

which is no longer used?

jingtang10, line 1: upon reading this PR i got really confused what `getWorkerInfo` is returning. the doc for that function is actually really really thin. can you add more details and add some examples?

I asked gemini to create an example:

```
/**
 * Retrieves the detailed status and progress information for a specific [FhirSyncWorker] job.
 *
 * This function monitors the [WorkManager] for the specified work name and provides a reactive flow of updates.
 * Each update includes the current work state (e.g., ENQUEUED, RUNNING, SUCCEEDED, etc.) along with the 
 * parsed job-specific status if available.
 *
 * @param context The application context to access the WorkManager.
 * @param workName The unique identifier of the worker job to track.
 * 
 * @return A Flow of [Pair] objects where:
 *   * First element: The current [WorkInfo.State] of the job.
 *   * Second element (optional): A [SyncJobStatus] object deserialized from the worker's progress data.
 *     The specific type of [SyncJobStatus] depends on the "StateType" key in the progress data.
 *     If the progress data is empty, missing the necessary keys, or an error occurs during deserialization,
 *     this element will be null.
 * 
 * @throws JsonParseException If the Gson deserialization of the 'State' data fails.
 * @throws ClassNotFoundException If the class specified in 'StateType' is not found.
 */
```

please use this as an inspiration only

jingtang10 (changes requested):

engine/src/androidTest/java/com/google/android/fhir/sync/SyncInstrumentedTest.kt
```diff
@@ -104,12 +105,49 @@ class SyncInstrumentedTest {
     assertThat(states.last()).isInstanceOf(CurrentSyncJobStatus.Succeeded::class.java)
   }
 
+  @Test
+  fun oneTime_worker_nextExecutionAfterSucceeded() {
```
jingtang10, line 109: can you rename this to say what the expectation is for this function? is it to test that the next execution should start with running?

also - can you double check if you revert your change in Sync.kt, this test case would fail?

santosh-pingle:
Working on review comments.

engine/src/androidTest/java/com/google/android/fhir/sync/SyncInstrumentedTest.kt
santosh-pingle: actually, test is not failing after reverting the change, it looks like syncJobStatus value in workManager api is not null.

santosh-pingle: Not sure how to set up condition WorkManager returns null status for ENQUEUED, RUNNING, and CANCELLED states for this specific change.

engine/src/main/java/com/google/android/fhir/sync/Sync.kt
santosh-pingle: added kotlin doc.

engine/src/androidTest/java/com/google/android/fhir/sync/SyncInstrumentedTest.kt
santosh-pingle: acknowledged.

PR #2513, opened 2024-04-17
Add initial new website (see #2232)

vorburger:
see #2232


vorburger:
I've finally had a moment to have a closer look into why as-is this "works for me" (local), but initially fails the CI build:

It's because `build-docs.bash` uses a `pipenv` - which is not available by default in GitHub Runner VM images, apparently.

I could just add https://pipenv.pypa.io/en/latest/installation.html to `build-docs.bash` ... or is there a ready-made action?

Looking around a bit, discarding a few repos with 0 stars, I have zoomed in on:

1. https://github.com/tiagovrtr/actions-pipenv
1. https://github.com/VaultVulp/action-pipenv
1. https://github.com/palewire/install-python-pipenv-pipfile

Of these, the last one (@palewire's) does both Python & PipEnv, and seem to do some caching; so I'll try to add that...

vorburger:
>  I'm trying https://github.com/tiagovrtr/actions-pipenv instead, let's see...

That seems to work better, but `pipenv run mkdocs build` now fails with:

```
Warning: Python 3.11 was not found on your system...
Neither 'pyenv' nor 'asdf' could be found to install Python.
You can specify specific versions of Python with:
$ pipenv --python path/to/python
Error: Process completed with exit code 1.
```

so trying Python 3.11 instead of latest 3.12 which I've initially selected (whatever works).

vorburger:
> so trying Python 3.11 instead of latest 3.12 which I've initially selected (whatever works).

That helped! Plus adding a missing install - I must have done that locally, but CI needs to explicitly do it as well, of course.

With that, this is ready for review! https://google.github.io/android-fhir/ should look very different after this is merged...

.github/workflows/build.yml
```diff
@@ -99,6 +99,16 @@ jobs:
       # This deploys content based on docs/ which ends up in site/
       # to https://google.github.io/android-fhir/ using GitHub Pages.
 
+      - name: Install Python
+        uses: actions/setup-python@v5
+        with:
+          python-version: 3.11
+          cache: "pipenv"
+      - name: Install PipEnv
```
MJ1998, line 107: Should we cache the pipenv enviroment when the pipfile.lock is unchanged ?

.vscode/settings.json
```diff
@@ -0,0 +1,14 @@
+{
+  // https://squidfunk.github.io/mkdocs-material/creating-your-site/
+  "yaml.schemas": {
+    "https://squidfunk.github.io/mkdocs-material/schema.json": "mkdocs.yml"
```
MJ1998, line 4: NIT: Spell check. "mkdocs.yaml" 
```suggestion
    "https://squidfunk.github.io/mkdocs-material/schema.json": "mkdocs.yaml"
```

Or if `yml` is preffered then modify the file extension

Pipfile
```diff
@@ -0,0 +1,12 @@
+[[source]]
+url = "https://pypi.org/simple"
+verify_ssl = true
+name = "pypi"
+
+[packages]
+mkdocs-material = "*"
```
MJ1998, line 7: I think we should pin a version so that there are no potential issues with a breaking change in the latest versions.

MJ1998, line 7: From Gemini, should we also add following plugins:-
`mkdocs-minify-plugin`
`mkdocs-git-revision-date-plugin`

.github/workflows/build.yml
vorburger: That's a great point! But my reading of https://github.com/tiagovrtr/actions-pipenv was that the `cache: "pipenv"` just does this correctly already, as-is. I can see some cachine related logs [here](https://github.com/google/android-fhir/actions/runs/8927162902/job/24519899802?pr=2513) - so to me it looked like this was working? (If I misunderstand what you mean, perhaps you would like to suggest a specific change? But perhaps that would be best first documented on https://github.com/tiagovrtr/actions-pipenv ...)

Pipfile
vorburger: That's a great idea! Thank You for suggesting this. I have read up about and learnt about the difference between fixing versions in `Pipfile.lock` (which IS reproducible on CI, BUT changes on OTHER updates) VS `Pipfile` Done; resolving.

.vscode/settings.json
vorburger: Well spotted! Thank You for noticing this oversight. Done / fixed; resolving.

Pipfile
vorburger: Gemini, huh? 🔮 Sure thing, you got it - Done; resolving. (But I've opted to use https://github.com/timvink/mkdocs-git-revision-date-localized-plugin instead of https://github.com/zhaoterryy/mkdocs-git-revision-date-plugin.)

.github/workflows/build.yml
MJ1998: I dont fully understand it but Gemini suggests this:-
```
- name: Cache Pipenv Environment
  uses: actions/cache@v3
  with:
    path: ~/.local/share/virtualenvs/*
    key: ${{ runner.os }}-pipenv-${{ hashFiles('Pipfile.lock') }}
    restore-keys: |
      ${{ runner.os }}-pipenv-
```

MJ1998 (approved):
LGTM

.github/workflows/build.yml
MJ1998: Resolving as this a minor build optimization and can be taken up later (if we notice slow builds)

PR #2514, opened 2024-04-18
Date Picker Constraint Issue Fix.

### LINKED ISSUE #2508: Date Picker Constraint Not Enforced on 'Ok' Button Click.
**Describe the bug**
In the date picker if there is some constraint which does not allow to select the current date and the user still clicks on "OK" button without selecting any date it still selects the current date as the answer as it is the default to it.

[Screen_recording_20240404_153456.webm](https://github.com/google/android-fhir/assets/102521117/dc99a15b-a44e-4acb-88c0-dbe59565ca0e)

**Component**
SDC Gallery app

**To Reproduce**

**Expected behavior**
There can be two approaches to solving this:
1. Disable the "OK" button if the date selected is invalid enable only when the selected date is valid.
2. Set the default date to the minimum valid date. This way we can reduce the number of clicks the user has to do in order to select a valid date.

**Screenshots**

**Additional context**

Yes.

AdityaKBhadragond14:

Fixes #2508

**Description**
The MaterialDatePicker Builder has been modified such that if the questionnaire item has the minimum date extension added then the MaterialDatePicker.setSelection() will have that minimum date from extension else it will get the local date. This way it will select the first valid date based on the constraints added for date.

**Alternative(s) considered**

**Type**
Bug Fix

**Checklist**

### DIFF
--- a/datacapture/src/main/java/com/google/android/fhir/datacapture/views/factories/DatePickerViewHolderFactory.kt
+++ b/datacapture/src/main/java/com/google/android/fhir/datacapture/views/factories/DatePickerViewHolderFactory.kt
@@ -48,6 +48,7 @@ import com.google.android.material.textfield.TextInputLayout
 import java.text.ParseException
 import java.time.Instant
 import java.time.LocalDate
+import java.time.LocalDateTime
 import java.time.ZoneId
 import java.time.format.DateTimeParseException
 import java.util.Date
@@ -154,21 +155,31 @@ internal object DatePickerViewHolderFactory :
       }
 
       private fun buildMaterialDatePicker(localDate: LocalDate?): MaterialDatePicker<Long> {
-        val selectedDateMillis =
-          localDate?.atStartOfDay(ZONE_ID_UTC)?.toInstant()?.toEpochMilli()
-            ?: MaterialDatePicker.todayInUtcMilliseconds()
+        val min =
+          (questionnaireViewItem.minAnswerValue as? DateType)?.value?.time
+            ?: localDate?.atStartOfDay(ZONE_ID_UTC)?.toInstant()?.toEpochMilli()
+              ?: MaterialDatePicker.todayInUtcMilliseconds()
+        val minLocalDateTime =
+          LocalDateTime.ofInstant(
+            Instant.ofEpochMilli(min),
+            ZoneId.of(
+              ZONE_ID_UTC.id,
+            ),
+          )
+        val newMinLocalDateTime = minLocalDateTime.plusDays(1)
+        val selectedDateTimeMillis =
+          newMinLocalDateTime.atZone(ZoneId.of(ZONE_ID_UTC.id)).toInstant().toEpochMilli()
+        val max = (questionnaireViewItem.maxAnswerValue as? DateType)?.value?.time
+        val calendarConstraints = getCalenderConstraint(min, max)
 
         return MaterialDatePicker.Builder.datePicker()
           .setTitleText(R.string.select_date)
-          .setSelection(selectedDateMillis)
-          .setCalendarConstraints(getCalenderConstraint())
+          .setSelection(selectedDateTimeMillis)
+          .setCalendarConstraints(calendarConstraints)
           .build()
       }
 
-      private fun getCalenderConstraint(): CalendarConstraints {
-        val min = (questionnaireViewItem.minAnswerValue as? DateType)?.value?.time
-        val max = (questionnaireViewItem.maxAnswerValue as? DateType)?.value?.time
-
+      private fun getCalenderConstraint(min: Long?, max: Long?): CalendarConstraints {
         if (min != null && max != null && min > max) {
           throw IllegalArgumentException("minValue cannot be greater than maxValue")
         }
--- a/datacapture/src/androidTest/java/com/google/android/fhir/datacapture/views/factories/DatePickerViewHolderFactoryEspressoTest.kt
+++ b/datacapture/src/androidTest/java/com/google/android/fhir/datacapture/views/factories/DatePickerViewHolderFactoryEspressoTest.kt
@@ -1,0 +1,106 @@
+/*
+ * Copyright 2024 Google LLC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *       http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.google.android.fhir.datacapture.views.factories
+
+import android.view.View
+import android.widget.FrameLayout
+import androidx.test.espresso.Espresso.onView
+import androidx.test.espresso.action.ViewActions.click
+import androidx.test.espresso.assertion.ViewAssertions.matches
+import androidx.test.espresso.matcher.RootMatchers.isDialog
+import androidx.test.espresso.matcher.ViewMatchers.isDisplayed
+import androidx.test.espresso.matcher.ViewMatchers.withId
+import androidx.test.espresso.matcher.ViewMatchers.withText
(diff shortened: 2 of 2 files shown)


datacapture/src/main/java/com/google/android/fhir/datacapture/views/factories/DatePickerViewHolderFactory.kt
```diff
@@ -154,21 +155,31 @@ internal object DatePickerViewHolderFactory :
...
+            ?: localDate?.atStartOfDay(ZONE_ID_UTC)?.toInstant()?.toEpochMilli()
+              ?: MaterialDatePicker.todayInUtcMilliseconds()
+        val minLocalDateTime =
+          LocalDateTime.ofInstant(
+            Instant.ofEpochMilli(min),
+            ZoneId.of(
+              ZONE_ID_UTC.id,
+            ),
+          )
+        val newMinLocalDateTime = minLocalDateTime.plusDays(1)
```
MJ1998, line 169: Why is this needed - specifically "plusDays" ?
Once we get minDateInMillis we can set that directly right ?

datacapture/src/main/java/com/google/android/fhir/datacapture/views/factories/DatePickerViewHolderFactory.kt
```diff
@@ -154,21 +155,31 @@ internal object DatePickerViewHolderFactory :
       }
 
       private fun buildMaterialDatePicker(localDate: LocalDate?): MaterialDatePicker<Long> {
-        val selectedDateMillis =
```
MJ1998, line 157: NIT: name this variable properly. Maybe `minDateInMillis` ?

MJ1998 (commented):
Thanks Aditya for this PR. Can you address the comments ?

datacapture/src/main/java/com/google/android/fhir/datacapture/views/factories/DatePickerViewHolderFactory.kt
AdityaKBhadragond14: Oh my bad. Actually the questionnaireViewItem.minAnswerValue gets the GMT date instead of local date that is the reason why I was using "plusDays" to get the correct minimum date. I will change it to get the local date.

AdityaKBhadragond14:
> Thanks Aditya for this PR. Can you address the comments ?

I have addressed the comments. Please do a re-review. Thanks.

datacapture/src/main/java/com/google/android/fhir/datacapture/views/factories/DatePickerViewHolderFactory.kt
```diff
@@ -154,21 +154,28 @@ internal object DatePickerViewHolderFactory :
...
+            ?.value
+            ?.localDate
+            ?.atStartOfDay(
+              ZONE_ID_UTC,
+            )
+            ?.toInstant()
+            ?.toEpochMilli()
+            ?: localDate?.atStartOfDay(ZONE_ID_UTC)?.toInstant()?.toEpochMilli()
+              ?: MaterialDatePicker.todayInUtcMilliseconds()
+        val max = (questionnaireViewItem.maxAnswerValue as? DateType)?.value?.time
```
MJ1998, line 168: Nit: Rename this too to `maxDateInMillis` ?

MJ1998 (approved):
LGTM

MJ1998:
Hi @AdityaKBhadragond14 

It seems the build is failing on few tests in QuestionnaireUiEspressoTest. It's not available via kokoro logs here and we are working on it via this ticket [Issue/2515](https://github.com/google/android-fhir/issues/2515).

In the meantime these are the errors:-
Screenshot 2024-06-03 at 6 11 16 PM

This should be an easy fix. Requesting you to fix these and upload a commit. It will automatically merge the PR.
Thanks and apologies for the late response on this.

AdityaKBhadragond14:
> Hi @AdityaKBhadragond14
> 
> It seems the build is failing on few tests in QuestionnaireUiEspressoTest. It's not available via kokoro logs here and we are working on it via this ticket [Issue/2515](https://github.com/google/android-fhir/issues/2515).
> 
> In the meantime these are the errors:- Screenshot 2024-06-03 at 6 11 16 PM
> 
> This should be an easy fix. Requesting you to fix these and upload a commit. It will automatically merge the PR. Thanks and apologies for the late response on this.

Sure @MJ1998 . I will do it. Thanks.

datacapture/src/main/java/com/google/android/fhir/datacapture/views/factories/DatePickerViewHolderFactory.kt
```diff
@@ -154,21 +154,28 @@ internal object DatePickerViewHolderFactory :
...
+          (questionnaireViewItem.minAnswerValue as? DateType)
+            ?.value
+            ?.localDate
+            ?.atStartOfDay(
+              ZONE_ID_UTC,
+            )
+            ?.toInstant()
+            ?.toEpochMilli()
+            ?: localDate?.atStartOfDay(ZONE_ID_UTC)?.toInstant()?.toEpochMilli()
+              ?: MaterialDatePicker.todayInUtcMilliseconds()
```
jingtang10, line 167: some of this can probably be simplified, the conversion from local date to epoch milli is the same for the min value and the local date input, so you can simplify this code a bit by rewriting that.

but the bigger question is why this logic is written as such - you're reading from the questionnaire constraint's min value, and if it's missing, then you take the input value (which could have been an input from before, and then if that's also missing, you take today's date. and then you set this as the selection. and then you use this for two things: 1) the selection for the date picker and 2) the min date for the calendar constraint. i don't really understand why this is done this way.

to me, a perhaps much simpler fix is to simply not set the selection below:

```
.setSelection(selectedDateMillis)
```

if the selection will not make any sense. have you tried removing this line?

in that case i'd imagine the behaviour of the date picker will be that you cannot click ok since there simply isn't any possible date to choose from that would satisfy the calendar constraints.

have you tried that? i feel that's a more logical fix.

jingtang10 (changes requested):

datacapture/src/main/java/com/google/android/fhir/datacapture/views/factories/DatePickerViewHolderFactory.kt
AdityaKBhadragond14: Hi @jingtang10 , this makes sense. Removing the .setSelection(selectedDateMillis) will disable the "ok" button and enable only when selected a date. I think this could be more better fix than what I had proposed. Thanks for pointing it out.

AdityaKBhadragond14:
@jingtang10 I have made the changes please have a look. Thanks.

datacapture/src/main/java/com/google/android/fhir/datacapture/views/factories/DatePickerViewHolderFactory.kt
jingtang10: you still to set the selection if there is an existing answer right? why are you removing it completely?

jingtang10 (changes requested):

AdityaKBhadragond14:
Ohh I am sorry, I misunderstood it. You mean the existing answer will come from the Questionnaire? @jingtang10

datacapture/src/main/java/com/google/android/fhir/datacapture/views/factories/DatePickerViewHolderFactory.kt
MJ1998: The idea here is to streamline user experience, Jing. This is what I had in mind:-
**Invalid Date Range:** If minDate is later than maxDate, disable the date picker and display a clear error message.
**Existing Answer:** If the user has already selected a date (i.e., localDate is not null), set the date picker to that value.
**No Answer, Valid Range:** If there's no previous answer but the date range is valid:
1. Prioritize Today's Date: Since users are likely to choose today's date, set the date picker to the today's date if it falls within the valid range (minDate to maxDate).
2. Fallback to minDate: If today's date is outside the valid range, use minDate as the default selection to avoid excessive scrolling or clicking for the user.

Notes:-
1. Once user selects a valid date, ie, localDate is non null, then we should set the date picker to that value only.
2. minDateInMillis should be calculated solely based on minValueAnswer, not influenced by the user's current selection or the current date. 

So I guess something like following:-

```
val dateToSelect = when { 
    localDate != null -> localDate
    currentDate between minDate..maxDate -> currentDate 
    else -> minDate // If not in range, minDate fallback
}
```

PR #2524, opened 2024-04-23
Ordering of patches with cycles during upload

### LINKED ISSUE #2500: Sync upload local changes reordering patch issue
**Describe the bug**
After the latest updates to Sync, an exception is thrown on the [code block/line here](https://github.com/google/android-fhir/blob/a5db185c60035c4b9a1f1c287a3d4f0c2dbba766/engine/src/main/java/com/google/android/fhir/sync/upload/patch/PatchOrdering.kt#L131) :

```kotlin
private fun createTopologicalOrderedList(adjacencyList: Map<Node, List<Node>>): List<Node> {
    val stack = ArrayDeque<String>()
    val visited = mutableSetOf<String>()
    val currentPath = mutableSetOf<String>()

    fun dfs(key: String) {
      check(currentPath.add(key)) { "Detected a cycle." }
      if (visited.add(key)) {

```
The check throws an exception if it detects a cycle, however there could be valid cycles . 

See screenshot (Shows breakpoint debugging of [code block/line above](https://github.com/google/android-fhir/blob/a5db185c60035c4b9a1f1c287a3d4f0c2dbba766/engine/src/main/java/com/google/android/fhir/sync/upload/patch/PatchOrdering.kt#L131) )

**Component**
Core library - Engine

**To Reproduce**
1. Trigger sync upload after a workflow that creates resources with a dependency tree as shown on the screenshot.

**Expected behavior**
Sync upload should be successful

**Screenshots**
Screenshot 2024-03-26 at 15 14 50

[Screenshot shows breakpoint debugging of line here](https://github.com/google/android-fhir/blob/a5db185c60035c4b9a1f1c287a3d4f0c2dbba766/engine/src/main/java/com/google/android/fhir/sync/upload/patch/PatchOrdering.kt#L131) 

N/A

**Additional context**
N/A

@aditya-07 worked on the original issue

aditya-07:

Fixes #2500

**Description**

~~1. Find subgraphs (weakly connected) of Resources with cycles in them .
2. Pack the subgraphs in  Bundles such that a single subgraph is not split between two Bundles (as it may cause server error).
3. Order the remaining resources and add them to the Bundle (in remaining space of or new Bundles) .~~

1. Find the ordered strongly connected components of the Resource Graph.
2. Pack the sccs in Bundles such that a single scc is not split between two Bundles.

**Alternative(s) considered**

**Type**
Choose one: Bug fix

**Checklist**


MJ1998:
Is this ready for review ?

MJ1998 (changes requested):
What if we use Strongly Connected Components (SCC) ? 
1. Identify the SCCs
2. Consolidate all nodes in each SCC into a supernode.
3. Then apply the same topological sorting

So for following graph:-
1 -> 2, 3
2 -> 4, 5
4 -> 1
5 -> 6
7 -> 8, 9
8 -> 7

SCC = [{1,2,4}, {7, 8}]
We can have the topological sort as - [{3}, {6}, {5}, {1, 2, 4}, {9}, {7, 8}]

Current solution is also right but the above solution has more improvements:-
1. Time complexity is one-fourth (I guess).
2. In current approach the weakly-connected-components will go in one bundle if there is a cycle. Whereas in the above approach only the SCC will go in one bundle -> Reduces the size of the bundle a lot.

MJ1998:
It's the Tarjan's Algorithm: https://www.geeksforgeeks.org/tarjan-algorithm-find-strongly-connected-components/

engine/src/main/java/com/google/android/fhir/sync/upload/patch/PatchGenerator.kt
```diff
@@ -67,3 +67,20 @@ internal data class PatchMapping(
...
+
+  /**
+   * [IndividualMapping] doesn't have a cyclic dependency on other [IndividualMapping] /
+   * [PatchMapping]. It may however have an acyclic dependency on other [IndividualMapping]s /
+   * [PatchMapping]s.
+   */
+  data class IndividualMapping(val patchMapping: PatchMapping) : OrderedMapping
+
+  /**
+   * [CombinedMapping] contains weakly connected [PatchMapping]s where one or more [PatchMapping]s
```
jingtang10, line 82:
```suggestion
   * [CombinedMapping] contains strongly connected [PatchMapping]s where one or more [PatchMapping]s
```

engine/src/main/java/com/google/android/fhir/sync/upload/patch/PatchGenerator.kt
```diff
@@ -67,3 +67,20 @@ internal data class PatchMapping(
   val localChanges: List<LocalChange>,
   val generatedPatch: Patch,
 )
+
+/** Structure to help describe the cyclic nature of ordered [PatchMapping]. */
+internal sealed interface OrderedMapping {
```
jingtang10, line 72: i think there is no order in this structure itself - so maybe patch mapping group is a better name - or something similar?

engine/src/main/java/com/google/android/fhir/sync/upload/patch/PatchOrdering.kt
```diff
@@ -139,4 +156,138 @@ internal object PatchOrdering {
     adjacencyList.keys.forEach { dfs(it) }
     return stack.reversed()
   }
+
+  private fun stronglyConnectedComponents(diGraph: Graph, nodesCount: Int): List<List<Node>> {
```
jingtang10, line 160: optionally - you can separate all the logic related to scc to a separate class - and try to write it as an internal api.

engine/src/main/java/com/google/android/fhir/sync/upload/patch/PatchOrdering.kt
```diff
@@ -139,4 +156,138 @@ internal object PatchOrdering {
...
+    oldGraph: Graph,
+    nodeMapping: Pair<Map<Node, List<Node>>, Map<Node, Node>>,
+  ): Graph {
+    val newGraph = mutableMapOf<Node, List<Node>>()
+    val (superNodeToNodes, nodeToSuperNode) = nodeMapping
+    // Remove any cyclic dependency from connected components.
+    // reduce them to a single node
+    // replace the ssc nodes with super node
+    oldGraph.forEach {
+      //      println(it.key)
```
jingtang10, line 189:
```suggestion
```

engine/src/main/java/com/google/android/fhir/sync/upload/patch/PatchOrdering.kt
```diff
@@ -139,4 +156,138 @@ internal object PatchOrdering {
...
+
+  private fun transformGraph(
+    oldGraph: Graph,
+    nodeMapping: Pair<Map<Node, List<Node>>, Map<Node, Node>>,
+  ): Graph {
+    val newGraph = mutableMapOf<Node, List<Node>>()
+    val (superNodeToNodes, nodeToSuperNode) = nodeMapping
+    // Remove any cyclic dependency from connected components.
+    // reduce them to a single node
+    // replace the ssc nodes with super node
```
jingtang10, line 187: reformat this please (capitalise, punctuation etc)

engine/src/main/java/com/google/android/fhir/sync/upload/request/TransactionBundleGenerator.kt
jingtang10, line 1: please udpate the unit tests here

engine/src/main/java/com/google/android/fhir/sync/upload/request/UrlRequestGenerator.kt
```diff
@@ -30,15 +31,37 @@ internal class UrlRequestGenerator(
   private val getUrlRequestForPatch: (patch: Patch) -> UrlUploadRequest,
 ) : UploadRequestGenerator {
 
+  /**
+   * Since a [UrlUploadRequest] can only handle a single resource request, the
+   * [OrderedMapping.CombinedMapping.patchMappings] are flattened and handled as
+   * [OrderedMapping.IndividualMapping] mapping to generate [UrlUploadRequestMapping] for each
+   * [PatchMapping].
```
jingtang10, line 38: this is good. but can you also add that we could potentially make this illegal in the future and explicitly introduce configurations for referential integrity on the server side.

and maybe just add a little bit to explain that single resource requests won't be able to handle resources referencing each other if the server is strict about referential integrity.

jingtang10 (changes requested):

engine/src/main/java/com/google/android/fhir/sync/upload/patch/PatchGenerator.kt
aditya-07: Done

engine/src/main/java/com/google/android/fhir/sync/upload/patch/PatchOrdering.kt
aditya-07: done

engine/src/main/java/com/google/android/fhir/sync/upload/request/TransactionBundleGenerator.kt
aditya-07: Done

engine/src/main/java/com/google/android/fhir/sync/upload/request/UrlRequestGenerator.kt
aditya-07: Done

engine/src/main/java/com/google/android/fhir/sync/upload/patch/PatchOrdering.kt
aditya-07: Done

engine/src/main/java/com/google/android/fhir/sync/upload/patch/StronglyConnectedPatches.kt
```diff
@@ -0,0 +1,171 @@
...
+   * Converts a graph that has ssc to a graph with super nodes as per the provided [nodeToSuperNode]
+   * mapping.
+   *
+   * **Input**
+   * * oldGraph `[A : B, C], [B : A, C], [C : A, B, F], [D : E], [E : D], [F : G]`
+   * * nodeToSuperNode `[A:1, B:1, C:1, D:2, E:2, F:3, G:4]`
+   *
+   * **Output** `[1 : 3], [2], [3: 4]`
+   */
+  private fun transformGraphWithSSCtoGraphWithSuperNodes(
```
jingtang10, line 142: what this function actually does is: generating edges in the graph G' of super nodes from the edges in the original graph G.

so if there's an edge a->b in G, then there is an edge from S(a)->S(b) in G' where S(x) is the super node of x and S(a)!=S(b).

maybe you can add this to the description as at the moment it's not as detailed.

engine/src/main/java/com/google/android/fhir/sync/upload/patch/PatchOrdering.kt
```diff
@@ -71,7 +74,17 @@ internal object PatchOrdering {
         .groupBy { it.localChangeId }
 
     val adjacencyList = createAdjacencyListForCreateReferences(localChangeIdToResourceReferenceMap)
-    return createTopologicalOrderedList(adjacencyList).mapNotNull { resourceIdToPatchMapping[it] }
+
+    return StronglyConnectedPatches(adjacencyList, resourceIdToPatchMapping.size)
+      .sscOrdered { createTopologicalOrderedList(it) }
```
jingtang10, line 79: i think you can do

```
.sscOrdered(::createTopologicalOrderedList)
```

engine/src/main/java/com/google/android/fhir/sync/upload/patch/StronglyConnectedPatches.kt
```diff
@@ -0,0 +1,171 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.google.android.fhir.sync.upload.patch
+
+import kotlin.math.min
+
+internal class StronglyConnectedPatches(private val diGraph: Graph, private val nodesCount: Int) {
```
jingtang10, line 21: this needs some tests. especially simple cases e.g. graph of size 1, graph of size 0. graph of size 2 with or without a single link.

engine/src/main/java/com/google/android/fhir/sync/upload/patch/PatchGenerator.kt
```diff
@@ -67,3 +67,20 @@ internal data class PatchMapping(
   val localChanges: List<LocalChange>,
   val generatedPatch: Patch,
 )
+
+/** Structure to help describe the cyclic nature of ordered [PatchMapping]. */
+internal sealed interface PatchMappingGroup {
```
jingtang10, line 72: i wonder if we can define a single structure which is a collection of patch mappings that should be uploaded together. in (possibly most) some cases, they should just be individual patch mappings, but in the case of cycles, they would be bundled together as a result of your SCC calculation.

engine/src/main/java/com/google/android/fhir/sync/upload/patch/PatchOrdering.kt
```diff
@@ -20,7 +20,9 @@ import androidx.annotation.VisibleForTesting
 import com.google.android.fhir.db.Database
 import com.google.android.fhir.db.LocalChangeResourceReference
 
-private typealias Node = String
+typealias Node = String
+
+typealias Graph = Map<Node, List<Node>>
```
jingtang10, line 25: think about this alternative definition: 

```
data class Node(private id: String, private out: List<Node>)
typealias Graph = List<Node>
```

and potentially this *may* simplify your code a bit. because you can define things like

```
data class SuperNode(val out: List<SuperNode>, val nodes: List<Nodes>)
data class Node(val id: String, val out: List<Node>, val superNode: SuperNode)
```

and another thing is that - if we use this, we might be able to do away with copying the strings.

engine/src/main/java/com/google/android/fhir/sync/upload/patch/StronglyConnectedPatches.kt
```diff
@@ -0,0 +1,171 @@
...
+    val orderedNodes = ordering(updatedGraph)
+    return superNodeToSSC(orderedNodes, superNodeToConnectedComponents)
+  }
+
+  private fun sscTarzan(): List<List<Node>> {
+    // Code inspired from
+    // https://github.com/williamfiset/Algorithms/blob/master/src/main/java/com/williamfiset/algorithms/graphtheory/TarjanSccSolverAdjacencyList.java
+    var counter = -1
+    val intToNode = mutableMapOf<Int, String>()
+    val nodeToInt = mutableMapOf<String, Int>()
```
jingtang10, line 40: add comments here.

engine/src/main/java/com/google/android/fhir/sync/upload/patch/StronglyConnectedPatches.kt
```diff
@@ -0,0 +1,171 @@
...
+
+    var id = 0
+    var sccCount = 0
+    val visited = BooleanArray(graph.size)
+    val ids = IntArray(graph.size) { -1 }
+    val low = IntArray(graph.size)
+    val sccs = IntArray(graph.size)
+    val stack = ArrayDeque<Int>()
+
+    fun dfs(at: Int) {
```
jingtang10, line 73: add comment here

jingtang10 (changes requested):

engine/src/main/java/com/google/android/fhir/sync/upload/patch/PatchGenerator.kt
aditya-07: Done

engine/src/main/java/com/google/android/fhir/sync/upload/patch/StronglyConnectedPatches.kt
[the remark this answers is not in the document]
aditya-07: Done. Converted it into object and added docs to the class and public function.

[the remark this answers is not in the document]
aditya-07: Done. Removed the [nodeCount] param to remove the changes of wrong value being sent.

engine/src/main/java/com/google/android/fhir/sync/upload/patch/PatchOrdering.kt
aditya-07: No need based on the recent refactoring.

engine/src/main/java/com/google/android/fhir/sync/upload/patch/StronglyConnectedPatches.kt
aditya-07: Done. Added tests for various use cases with 1, 2 and multiple nodes.

engine/src/main/java/com/google/android/fhir/sync/upload/patch/PatchGenerator.kt
aditya-07: Done. PatchMappingGroup is now a simple data class.

engine/src/main/java/com/google/android/fhir/sync/upload/patch/PatchOrdering.kt
aditya-07: Tarjan computes the strongly connected components in topological order. So removed the code for various manipulations including generating super nodes.
I don't think we need this structure change now.

engine/src/main/java/com/google/android/fhir/sync/upload/patch/StronglyConnectedPatches.kt
aditya-07: Obsolete code based on recent refactoring. Removed.

engine/src/main/java/com/google/android/fhir/sync/upload/patch/StronglyConnectedPatches.kt
aditya-07: Done

engine/src/main/java/com/google/android/fhir/sync/upload/patch/StronglyConnectedPatches.kt
aditya-07: Done

engine/src/main/java/com/google/android/fhir/sync/upload/patch/PatchGenerator.kt
```diff
@@ -67,3 +67,10 @@ internal data class PatchMapping(
   val localChanges: List<LocalChange>,
   val generatedPatch: Patch,
 )
+
+/**
+ * Structure to describe the cyclic nature of [PatchMapping].
+ * - A single value in [patchMappings] signifies the acyclic nature of the node.
+ * - Multiple values in [patchMappings] signifies the cyclic nature of the nodes among themselves.
```
jingtang10, line 74: can you add a bit about how they are used? - they are used to make sure patches that have to be uploaded in the same request are uploaded in the same request...

engine/src/main/java/com/google/android/fhir/sync/upload/patch/PatchOrdering.kt
jingtang10, line 25: internal?

jingtang10, line 25: add examples for nodes and graphs.

jingtang10 (approved):
thank you @aditya-07 great job!

engine/src/main/java/com/google/android/fhir/sync/upload/patch/PatchOrdering.kt
aditya-07: Done

aditya-07: Done

engine/src/main/java/com/google/android/fhir/sync/upload/patch/PatchGenerator.kt
aditya-07: Done

engine/src/main/java/com/google/android/fhir/sync/upload/patch/StronglyConnectedPatches.kt
```diff
@@ -0,0 +1,82 @@
...
+   */
+  fun scc(directedGraph: Graph): List<List<Node>> {
+    return findSCCWithTarjan(directedGraph)
+  }
+
+  private fun findSCCWithTarjan(diGraph: Graph): List<List<Node>> {
+    val connectedComponents = mutableListOf<List<Node>>()
+    val currentLowValueForEachNode = mutableMapOf<Node, Int>()
+    var exploringCounter = 0
+    val assignedLowValueForEachNode = mutableMapOf<Node, Int>()
```
MJ1998, line 38: nit
```suggestion
    val lowLinks = mutableMapOf<Node, Int>()
```

`lowLinks` is the terminology used in tarjan's algorithm

engine/src/main/java/com/google/android/fhir/sync/upload/patch/StronglyConnectedPatches.kt
```diff
@@ -0,0 +1,82 @@
...
+   *   topologically ordered which may change based on the ordering algorithm and the [Node]s inside
+   *   a SSC may be ordered randomly depending on the path taken by algorithm to discover the nodes.
+   */
+  fun scc(directedGraph: Graph): List<List<Node>> {
+    return findSCCWithTarjan(directedGraph)
+  }
+
+  private fun findSCCWithTarjan(diGraph: Graph): List<List<Node>> {
+    val connectedComponents = mutableListOf<List<Node>>()
+    val currentLowValueForEachNode = mutableMapOf<Node, Int>()
```
MJ1998, line 36: nit: 
```suggestion
    val discoveryTimes = mutableMapOf<Node, Int>()
```

Just trying to use the same terminology in Tarjan's algorithm

engine/src/main/java/com/google/android/fhir/sync/upload/patch/StronglyConnectedPatches.kt
```diff
@@ -0,0 +1,82 @@
...
+    var exploringCounter = 0
+    val assignedLowValueForEachNode = mutableMapOf<Node, Int>()
+    val nodesCurrentlyInStack = mutableSetOf<Node>()
```
MJ1998, line 39: This not make the complexity O(nLogn).

If there is a way to map a String of format "<ResourceType>/<Id>" to an integer in O(1) we can reduce the complexity to O(n).

For the least you could use a HashSet ? With number of nodes in thousands the likelihood of collision is very low.
The complexity will almost always will be O(n)

engine/src/main/java/com/google/android/fhir/sync/upload/patch/PatchOrdering.kt
```diff
@@ -71,7 +91,10 @@ internal object PatchOrdering {
         .groupBy { it.localChangeId }
 
     val adjacencyList = createAdjacencyListForCreateReferences(localChangeIdToResourceReferenceMap)
-    return createTopologicalOrderedList(adjacencyList).mapNotNull { resourceIdToPatchMapping[it] }
+
+    return StronglyConnectedPatches.scc(adjacencyList).map {
+      PatchMappingGroup(it.mapNotNull { resourceIdToPatchMapping[it] })
```
MJ1998, line 96: Oh this step will anyway take O(nlogn) - See a comment below.
I guess we can associate each resourceTypeAndId to an Int and hence derive the `resourceIdToPatchMapping`
Then we can create an adjacencyList of Ints and the algorithm would run in O(n).
Finally do the reverse mapping.

MJ1998 (changes requested):

engine/src/main/java/com/google/android/fhir/sync/upload/patch/PatchGenerator.kt
```diff
@@ -67,3 +67,13 @@ internal data class PatchMapping(
...
+ *
+ * [PatchMappingGroup] is used by the engine to make sure that related resources get uploaded to the
+ * server in the same request to maintain the referential integrity of resources during creation.
+ */
+internal data class PatchMappingGroup(val patchMappings: List<PatchMapping>)
```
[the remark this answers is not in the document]
MJ1998: Also update kdoc to something like:-
```
/**
 * Represents a strongly connected component (SCC) within a directed graph of [PatchMapping] instances.
 *
 * A strongly connected component is a subgraph where every [PatchMapping] node is reachable from every
 * other node within the same subgraph. In the context of patch mappings, this signifies a group of patch
 * mappings that have cyclic dependencies among themselves.
 *
 * This class is used to group related [PatchMapping] instances together for processing, ensuring that
 * resources with dependencies are handled as a unit.
 */
 ```

File: engine/src/main/java/com/google/android/fhir/sync/upload/patch/StronglyConnectedPatches.kt (hunk 17 above)
Comment (MJ1998) [reply engine/src/main/java/com/google/android/fhir/sync/upload/patch/StronglyConnectedPatches.kt:39]: Resolving this as mutableSetOf() is a LinkedHashSet indeed.

File: engine/src/main/java/com/google/android/fhir/sync/upload/patch/PatchGenerator.kt (hunk 19 above)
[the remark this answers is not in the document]
Comment (aditya-07) [reply engine/src/main/java/com/google/android/fhir/sync/upload/patch/PatchGenerator.kt]: Done.

File: engine/src/main/java/com/google/android/fhir/sync/upload/patch/PatchOrdering.kt [hunk 20]
```diff
@@ -53,15 +72,16 @@ internal object PatchOrdering {
...
-   * @return - A ordered list of the [PatchMapping]s based on the references to other [PatchMapping]
-   *   if the mappings are acyclic
-   * - throws [IllegalStateException] otherwise
+   * @return A ordered list of the [PatchMappingGroup] containing:
+   * - [PatchMappingGroup] with single value for the [PatchMapping] based on the references to other
+   *   [PatchMapping] if the mappings are acyclic
+   * - [PatchMappingGroup] with multiple values for [PatchMapping]s based on the references to other
+   *   [PatchMapping]s if the mappings are cyclic.
    */
   suspend fun List<PatchMapping>.orderByReferences(
```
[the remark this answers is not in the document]
aditya-07: Done

engine/src/main/java/com/google/android/fhir/sync/upload/patch/PatchOrdering.kt
```diff
@@ -53,15 +72,16 @@ internal object PatchOrdering {
...
    * retained. Order of D shouldn't matter for the purpose of referential integrity.
    *
-   * @return - A ordered list of the [PatchMapping]s based on the references to other [PatchMapping]
-   *   if the mappings are acyclic
-   * - throws [IllegalStateException] otherwise
+   * @return A ordered list of the [PatchMappingGroup] containing:
+   * - [PatchMappingGroup] with single value for the [PatchMapping] based on the references to other
+   *   [PatchMapping] if the mappings are acyclic
+   * - [PatchMappingGroup] with multiple values for [PatchMapping]s based on the references to other
+   *   [PatchMapping]s if the mappings are cyclic.
```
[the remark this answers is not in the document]
aditya-07: Done. I haven't added the link here, but at the `StronglyConnectedPatches.findSCCWithTarjan`.

engine/src/main/java/com/google/android/fhir/sync/upload/patch/StronglyConnectedPatches.kt
```diff
@@ -0,0 +1,82 @@
...
+   * @return An ordered List of strongly connected components of the [directedGraph]. The SCCs are
+   *   topologically ordered which may change based on the ordering algorithm and the [Node]s inside
+   *   a SSC may be ordered randomly depending on the path taken by algorithm to discover the nodes.
+   */
+  fun scc(directedGraph: Graph): List<List<Node>> {
+    return findSCCWithTarjan(directedGraph)
+  }
+
+  private fun findSCCWithTarjan(diGraph: Graph): List<List<Node>> {
+    val connectedComponents = mutableListOf<List<Node>>()
```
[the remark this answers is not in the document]
aditya-07: Done

engine/src/main/java/com/google/android/fhir/sync/upload/patch/StronglyConnectedPatches.kt
```diff
@@ -0,0 +1,82 @@
...
+
+      // We have found the head node in the scc.
+      if (currentLowValueForEachNode[at] == assignedLowValueForEachNode[at]) {
+        val connected = mutableListOf<Node>()
+        while (true) {
+          val node = stack.removeFirst()
+          connected.add(node)
+          nodesCurrentlyInStack.remove(node)
+          if (node == at || stack.isEmpty()) break
+        }
```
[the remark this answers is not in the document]
aditya-07: Done

engine/src/main/java/com/google/android/fhir/sync/upload/patch/StronglyConnectedPatches.kt
aditya-07: Done

engine/src/main/java/com/google/android/fhir/sync/upload/patch/StronglyConnectedPatches.kt
aditya-07: Done

engine/src/main/java/com/google/android/fhir/sync/upload/patch/StronglyConnectedPatches.kt
aditya-07: To keep separation of the actual algorithm and the generated graph, I think its fine if we keep `nodesCurrentlyInStack` and `resourceIdToPatchMapping` as two separate values in two separate places/ routines.

`visitedNodes` and `nodesCurrentlyInStack` are now boolean arrays.

engine/src/main/java/com/google/android/fhir/sync/upload/patch/PatchOrdering.kt
aditya-07: It should be O(n) as we are using a hashmap.

engine/src/main/java/com/google/android/fhir/sync/upload/patch/StronglyConnectedPatches.kt
```diff
@@ -0,0 +1,96 @@
...
+  fun scc(directedGraph: Graph): List<List<Node>> {
+    return findSCCWithTarjan(directedGraph)
+  }
+
+  /**
+   * Finds strongly connected components in topological order. See
+   * https://en.wikipedia.org/wiki/Tarjan%27s_strongly_connected_components_algorithm.
+   */
+  private fun findSCCWithTarjan(diGraph: Graph): List<List<Node>> {
+    val nodeCount = (diGraph.keys + diGraph.values.flatten().toSet()).size
```
MJ1998, line 39: Should'nt this be?
```suggestion
    val nodeCount = (diGraph.keys + diGraph.values.flatten()).toSet().size
```

Or actually the diGraph.keys will have all the keys already. Look at `createAdjacencyListForCreateReferences` where for each patchMapping we add a List<Node> which could also be empty.

engine/src/main/java/com/google/android/fhir/sync/upload/patch/StronglyConnectedPatches.kt
MJ1998: We can still create another mapping from resourceIdToPatchMapping where we map "resourceTypeAndId" to an Int.
The algorithm is still separate and will work on Node which is an alias of Int instead of String now.

MJ1998 (approved):
LGTM only. This is a great PR!

engine/src/main/java/com/google/android/fhir/sync/upload/patch/StronglyConnectedPatches.kt
aditya-07: I am trying to avoid intermediary mappings and this makes code more readable.

engine/src/main/java/com/google/android/fhir/sync/upload/patch/StronglyConnectedPatches.kt
aditya-07: `diGraph.keys` is already a set. Its just to make sure that if ever the code is called with a graph with just the edge added, it doesn't fail.

engine/src/main/java/com/google/android/fhir/sync/upload/patch/StronglyConnectedPatches.kt
aditya-07: Done

PR #2526, opened 2024-04-23
Configurable drop down answer option.

### LINKED ISSUE #2505: Configurable hyphen("-") answer option in dropdown menu.
**Describe the bug**
In the dropdown menu the hyphen ("-") answer option is being added if the user does not want to select any answer from the dropdown which is confusing as to what the hyphen states.

**Component**
SDC gallery app

**To Reproduce**

**Expected behavior**
 I suggest the hyphen option should be configurable such that it can be configured to be replaced by any other answer option like "None".

**Screenshots**
Screenshot_20240401_181445

**Additional context**

Yes.

AdityaKBhadragond14:

Fixes #2505

**Description**
In the DropDownViewHolderFactory the hyphen("-") is being set as the default answer option to the drop down. Instead of setting it a fixed answer I have made it configurable. To change the default answer option the questionnaire item representing the drop down item will have the initial property which will have a valueString. The default answer option will be filled with the initial property value if available else it will be filled hyphen("-"). 
This approach ensures that the configuration will be done via questionnaire itself and the code need be changed everytime to change a default answer option.

**Alternative(s) considered**

**Type**
Bug Fix.

**Checklist**

### DIFF
--- a/datacapture/src/main/java/com/google/android/fhir/datacapture/views/factories/DropDownViewHolderFactory.kt
+++ b/datacapture/src/main/java/com/google/android/fhir/datacapture/views/factories/DropDownViewHolderFactory.kt
@@ -67,6 +67,12 @@ internal object DropDownViewHolderFactory :
           hint = questionnaireViewItem.enabledDisplayItems.localizedFlyoverSpanned
           helperText = getRequiredOrOptionalText(questionnaireViewItem, context)
         }
+        val initialAnswerString =
+          if (questionnaireViewItem.questionnaireItem.hasInitial()) {
+            questionnaireViewItem.questionnaireItem.initial.first().valueStringType.valueAsString
+          } else {
+            context.getString(R.string.hyphen)
+          }
         val answerOptionList =
           this.questionnaireViewItem.enabledAnswerOptions
             .map {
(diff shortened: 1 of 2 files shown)


MJ1998 (changes requested):
This is incorrect as per rules in [Questionnaire](https://build.fhir.org/questionnaire.html):-
`+ Rule: If one or more answerOption is present, initial cannot be present. Use answerOption.initialSelected instead`

However we don't even need this. You can override the hyphen string constant in your application and the dropdown will read that value instead of "-" defined in the sdc library.

What you may do is change the name of the string constant from `hyphen` to maybe `default_drop_down_value` ?

jingtang10 (changes requested):
i agree with jajoo - instead of using initial answers, just provide a way for the user to override the string resource in the xml file. what you might want to do is to give that string a better name. and in the same pr please update the documentation in the docs/ folder or sdc library.

datacapture/src/main/java/com/google/android/fhir/datacapture/views/factories/DropDownViewHolderFactory.kt
```diff
@@ -80,8 +86,8 @@ internal object DropDownViewHolderFactory :
         answerOptionList.add(
           0,
           DropDownAnswerOption(
-            context.getString(R.string.hyphen),
-            context.getString(R.string.hyphen),
+            initialAnswerString,
+            initialAnswerString,
```
jingtang10, line 90: one thing you do want to be careful is that if there's a clash between the user defined default value and one of the answer options from the questionnaire. in this case you might want to have a mechanism to give them distinct names -- this is a very very edge case though and i'm happy for you to just add a todo in the code.

jingtang10 (changes requested):

PR #2534, opened 2024-05-06
Disable top-level navigation tabs on new website

vorburger:
A simple 'full' menu on the left seems much clearer, actually.

This just changes the "look and feel" (content remains the same).

It also removes one entry which was a duplicate.

Follow-up to #2513 for #2232.

@MJ1998 LGTY?


MJ1998 (approved):
Lgtm

PR #2535, opened 2024-05-06
Add FhirEngine interface method 'withTransaction'

### LINKED ISSUE #2531: Expose an interface to db.withtransaction through FhirEngine
Multiple db operations involving create/update/delete of resources cannot be handled atomically such that when updating and inserting extracted resources due to the action of saving a questionnaire, errors that may occur in between lead to inconsistencies and in some cases duplication when a user actions subsequent retries to save the questionnaire…

Expose an interface to db.withtransaction through FhirEngine that would allow multiple operations to be grouped and committed as  a transaction

Grouping using a Bundle resource of type ‘transaction’ that would map the different request types to a db crud method. A concern would be on whether it would respect the order of the operations

Yes

LZRS:

Fixes #2531 

**Description**
Support performing FhirEngine actions as a single atomic transaction

**Alternative(s) considered**

**Type**

**Checklist**


engine/src/main/java/com/google/android/fhir/FhirEngine.kt
```diff
@@ -212,6 +212,12 @@ interface FhirEngine {
    *   back and no record is purged.
    */
   suspend fun purge(type: ResourceType, ids: Set<String>, forcePurge: Boolean = false)
+
+  /**
+   * Adds support for performing actions on `FhirEngine` as a single atomic transaction where the
+   * entire set of changes succeed or fail as a single entity
+   */
+  suspend fun withTransaction(block: suspend FhirEngine.() -> Unit)
```
aditya-07, line 220: It would be better to confine the lambda with receiver to have access to CRUD operations only. In the current scenario, app would be able to call all the `FhirEngine` api on the lambda.

aditya-07: The other way could be to have a single api take a list of Pair<Resource, CRUD-OP-TYPE> and perform particular db operations sequentially based on the CRUD-OP-TYPE.

aditya-07 (commented):
Left some suggestions for the api.

engine/src/test/java/com/google/android/fhir/impl/FhirEngineImplTest.kt
```diff
@@ -783,6 +784,87 @@ class FhirEngineImplTest {
...
+      }
+
+    fhirEngine.create(patient01)
+    assertThat(fhirEngine.get<Patient>(patient01ID).gender)
+      .isEqualTo(Enumerations.AdministrativeGender.FEMALE)
+    fhirEngine.withTransaction {
+      fhirEngine.create(patient02, patient03)
+      patient01.gender = Enumerations.AdministrativeGender.FEMALE
+      fhirEngine.update(patient01)
+      fhirEngine.delete<Patient>(patient01ID)
```
MJ1998, line 815: Can you make this more logical ? In the sense that it reflects why withTransaction is needed. Maybe something like what you are trying to do but were failing because of absence of withTransaction.

MJ1998 (commented):
LGTM. @jingtang10

MJ1998:
Note: Sync APIs are deprecated now and once removed we probably dont need the CRUDEngine interface.

engine/src/main/java/com/google/android/fhir/impl/FhirEngineImpl.kt
```diff
@@ -97,6 +97,10 @@ internal class FhirEngineImpl(private val database: Database, private val contex
     }
   }
 
+  override suspend fun withTransaction(block: suspend FhirEngine.() -> Unit) {
+    database.withTransaction { block.invoke(this@FhirEngineImpl) }
```
jingtang10, line 101: does this work?

```suggestion
    database.withTransaction { this.block() }
```

engine/src/test/java/com/google/android/fhir/impl/FhirEngineImplTest.kt
```diff
@@ -783,6 +788,132 @@ class FhirEngineImplTest {
     assertThat(services.database.getLocalChangesCount()).isEqualTo(0)
   }
 
+  @Test
+  fun `withTransaction saves all changes successfully in order`() = runTest {
```
jingtang10, line 792: Can you simplify this test case?

I think you should keep this test case to use at most 2 resources. For example a patient and an observation. this is because when you add more resources to reference each other it's not really testing anything new.

Even just create a patient and then an observation that references the patient in my opinion is enough.

The point of the transation API isn't that things happen in order.... things will be executed in order without transation. The point is really that a transaction is atomic and will be reverted as a whole. So this test case should just be simply testing that the transation api runs multiple operations. the next test case is more important.

engine/src/test/java/com/google/android/fhir/impl/FhirEngineImplTest.kt
```diff
@@ -783,6 +788,132 @@ class FhirEngineImplTest {
...
+    assertThat(
+        fhirEngine.get<Observation>(patient01AppointmentEncounterObservationID).encounter.reference,
+      )
+      .isEqualTo("Encounter/$patient01AppointmentEncounterID")
+    assertThat(fhirEngine.get<Appointment>(patient01AppointmentID).status)
+      .isEqualTo(Appointment.AppointmentStatus.FULFILLED)
+  }
+
+  @Test
+  fun `withTransaction reverts all changes when an error occurs`() = runTest {
```
jingtang10, line 850: this test case can be done using 2 resources.

before transation:
create a patient

in a transation:
create an encoutner for the patient, update the encouter to reference a non-existent patient.

check that the encounter is also rolled back.

jingtang10 (approved):
Thanks @LZRS!

engine/src/main/java/com/google/android/fhir/impl/FhirEngineImpl.kt
LZRS: Yeah, works

engine/src/test/java/com/google/android/fhir/impl/FhirEngineImplTest.kt
LZRS: Makes sense...I've updated the changes

engine/src/test/java/com/google/android/fhir/impl/FhirEngineImplTest.kt
LZRS: Okay, I've updated the changes

engine/src/test/java/com/google/android/fhir/impl/FhirEngineImplTest.kt
```diff
@@ -783,6 +787,71 @@ class FhirEngineImplTest {
...
+          Encounter().apply {
+            id = "enc-01"
+            status = Encounter.EncounterStatus.FINISHED
+            class_ = Coding()
+            subject = Reference(patient01)
+          }
+
+        this.create(patientEncounter)
+
+        // Update encounter to reference non-existent subject to force ResourceNotFoundException
```
jingtang10, line 843:
```suggestion
        // An exception will rollback the entire block
```

PR #2537, opened 2024-05-08
Fix cursor moving back and missing characters

### LINKED ISSUE #1914: Cursor skipping back while start typing a value in numeric field
**Describe the bug**
Cursor is skipping back while start typing on the numeric field if the initial value is set to 0

**Component**
SDC library

**To Reproduce**
1. Go to any questionnaire.
2. Click on any field that has a support number only.
3. Type a 0 first.
4. Then type any number like 100 after 0.
5. You'll notice that the cursor is back to the second last character instead of the end.

**Expected behavior**
Cursor should stick to the end position of the field.

**Screenshots**

https://user-images.githubusercontent.com/62104757/225643988-45aa0c54-56b1-403a-874d-2beec2593324.mp4

 - Device: emulator
 - Android version: 11
 - Build number: RSR1.201013.001

MJ1998:

Fixes #1914 

Cause of the issue: Sometimes processing of input is slower than the typing speed. This results into inconsistency between the answer stored in the QuestionnaireViewItem and the current text in the EditText. The UI is updated everytime the QuestionnaireViewItem is bound. So [here](https://github.com/google/android-fhir/blob/63151b5f1fdb6c1ad843d1421ea354b84fb51378/datacapture/src/main/java/com/google/android/fhir/datacapture/views/factories/EditTextStringViewHolderDelegate.kt#L68) we see the inconsistency where we use EditText#setText which does 2 things: Sets the EditText content to the answer in QuestionnaireViewItem (leading to missing character) AND also sets the cursor to beginning of the input.

Solution:-
1. Cursor does not go to the beginning: EditText#append moves the cursor to the last position.
2. Characters are not missed (skipped): We separate the validation update from input-text-box update. And we update the input text box only when the view is not in focus. This happens only when EITHER the RecyclerView loads a new question into the view OR there is a read-only question. This results into not updating the EditText programmatically for each user input.

~The EditText temporarily contains an extra character. The TextWatcher gets removed ([here](https://github.com/google/android-fhir/blob/63151b5f1fdb6c1ad843d1421ea354b84fb51378/datacapture/src/main/java/com/google/android/fhir/datacapture/views/factories/EditTextViewHolderFactory.kt#L104)) before it can process this extra character, leaving the saved text state (in QuestionnaireViewItem) inconsistent with the actual EditText content. This inconsistency triggers a UI update using EditText#setText ([here](https://github.com/google/android-fhir/blob/63151b5f1fdb6c1ad843d1421ea354b84fb51378/datacapture/src/main/java/com/google/android/fhir/datacapture/views/factories/EditTextStringViewHolderDelegate.kt#L68)), which unintentionally resets the cursor position to the beginning.~

~Read comments in rest of the PR for more details.~

**Description**

**Alternative(s) considered**

**Type**

**Checklist**


MJ1998:
This, however, has a side effect (which existed earlier as well but was overlooked because of cursor resetting) of missing some characters when typing too fast.
Reason: EditText's extra character is skipped when using the above solution.

Solving this will require more effort. I can think of a solution which will involve detecting "user has stopped typing". Something like this - [stackoverflow](https://stackoverflow.com/a/52944488).

Let me know if you want to fix this as well. Will raise a separate issue for this.

Alternative: Use RxBindings or do some backpressure solutioning to handle all character inputs from user. But this will need a more complex handling - like "not updating the EditText until upstream processing (handleInput) is complete".

pld:
@jingtang10 what are your thoughts on this?

MJ1998:
I made an attempt at solving the "side effect" as well. Seems good to me but Idk if this behaviour - "detect user has stopped typing" can have any other side effects. 

Only difficulty is to test this as racing conditions are always a pain to test.
How I tested it - Ran my finger over "QWERTY" tabs on keyboard back and forth really fast. Before the last commit I could see some missed characters. After the last commit, I didn't see any missed character in quite some attempts trying this test.

Can you try testing this PR ?
@pld

AngelaKabari:
Thanks for all your work on this @MJ1998. @FikriMilano will test the PR tomorrow and we'll get back to you.

MJ1998:
Just FYI an elaboration on the how I am testing this:-
I slide my finger and keep it pressed over "QWERTY" keys back and forth. So the input I generate is "qwertytrewqwertytrewq...". 

You can observe some characters being missed before the last commit if you test like this. @FikriMilano 

In actual case the upstream function can take more time to finish (for example, resolving fhir query)  - which makes characters missing clearly visible.

MJ1998:
Regarding your question on RxBinding, I think its a much bigger effort as we dont use simple TextView or EditText views, rather its a EditTextInputLayout which offers more features like validation, header views along with it. 

So moving from EditTextInputLayout to RxBinding views will be difficult. 
@FikriMilano

AngelaKabari:
@MJ1998 this fix you shared is working great. Thank you for prioritising this and getting it resolved so quickly!

AngelaKabari:
@MJ1998 upon further testing, the issue of the missing characters is significantly affecting the data entry.
https://github.com/google/android-fhir/assets/8417097/b9775530-db3a-4cc6-a9ae-352ccf68e2af

Can we please go back to the drawing board on this?

MJ1998:
Is this issue happening on this PR?

AngelaKabari:
@MJ1998 yes it is.

MJ1998:
Can you try increasing the DELAY TIME to 1 second or 2 second and see if this is happening ?
Not the best solution but I am trying to figure out what is causing the delay in the upstream function (that validates the input string and sends it to the QuestionnaireViewModel for updating the RecyclerView) if there is no dependent items also. 
To increase delay time search for this line: `textInputEditText.afterTextChangedDelayed(500)`

MJ1998:
Can you try this solution in  @f-odhiambo device and confirm if it works ?
I am not sure because that if condition is what we are anyway doing in updateUI.

MJ1998:
Context: User input is sent back to QuestionnaireViewModel for validation, updating dependent items and then building the RecyclerView list -> which then comes back to the EditViewHolderFactory. 
The problem is happening when user input speed exceeds the speed at which this loop finishes.
So here's what we are implementing: 
We will detect new user input and when loop comes back to EditViewHolderFactor we will cut off the loop to prevent updating UI when THERE IS new user input present!
@jingtang10 @FikriMilano

pld:
yay, awesome, thank you @MJ1998 🎉 !

datacapture/src/main/java/com/google/android/fhir/datacapture/views/factories/EditTextViewHolderFactory.kt
```diff
@@ -99,25 +97,45 @@ abstract class QuestionnaireItemEditTextViewHolderDelegate(private val rawInputT
...
 
-  private fun displayValidationResult(validationResult: ValidationResult) {
-    textInputLayout.error =
-      getValidationErrorMessage(textInputLayout.context, questionnaireViewItem, validationResult)
+      // TextWatcher is set only once for each question item in scenario 1
+      textWatcher =
+        textInputEditText.doAfterTextChanged { editable: Editable? ->
+          context.lifecycleScope.launch { handleInput(editable!!, questionnaireViewItem) }
+        }
+    }
```
aditya-07, line 138: Can we add unit test for this. You should be able to use `view.requestFocus` and `view.clearFoucs` to simulate the use cases.

aditya-07 (changes requested):
The approach taken to fix the issue looks fine to me and the changes look good as well.
Just a comment to add Unit test to make sure that the code keeps working as expected.

jingtang10 (approved):

aditya-07 (approved):
LGTM! Thanks @MJ1998 !

PR #2553, opened 2024-05-27
Improve search speed 

### LINKED ISSUE #2547: Improve FhirEngine Search performance for sorting
**Describe the bug**
When loading around 7k patients ordered by Patient.Name, the db query takes around 25s
Without sorting the query takes around 35ms

This is when using 
```
fhirEngine.search<Patient> {
        filter(Patient.ACTIVE, { value = of(true) })
        sort(Patient.NAME, Order.ASCENDING)
        count = PaginationConstant.DEFAULT_PAGE_SIZE
        from = currentPage * PaginationConstant.DEFAULT_PAGE_SIZE
      }
```

that generates SQL query
```
SELECT a.resourceUuid, a.serializedResource
FROM ResourceEntity a
         LEFT JOIN StringIndexEntity b
                   ON a.resourceType = b.resourceType AND a.resourceUuid = b.resourceUuid AND b.index_name = 'name'
WHERE a.resourceType = 'Patient'
  AND a.resourceUuid IN (SELECT resourceUuid
                         FROM TokenIndexEntity
                         WHERE resourceType = 'Patient'
                           AND index_name = 'active'
                           AND index_value = 'true')
ORDER BY b.index_value ASC
LIMIT 20 OFFSET 0
```

Tested out that removing the `ORDER BY b.index_value ASC` reduces the latency of the query to milliseconds

**Expected behavior**
The usecase of loading resources with sort should be more performant

**Additional context**
Here's a link to the resources db containing around 7k patients, that can be used for testing
https://drive.google.com/file/d/1TgPw36BFI8BijsUPc1JkQE3l3RXANYzT/view?usp=drive_link

MJ1998:

Fixes #2547 
As [commented](https://github.com/google/android-fhir/issues/2547#issuecomment-2123426078) in the above issue, this change reduces time from ~25 seconds to few milliseconds (7k patients).

**Description**
Remove extra JOIN clause (`a.resourceType = b.resourceType`). This clause is not needed as we already filter the resourceType in WHERE clause. 

Sample DSL search:-
```
fhirEngine.search<Patient> {
        sort(Patient.NAME, Order.ASCENDING)
        count = PaginationConstant.DEFAULT_PAGE_SIZE
        from = currentPage * PaginationConstant.DEFAULT_PAGE_SIZE
      }
```
Query before this change:-
```
SELECT a.resourceUuid, a.serializedResource
FROM ResourceEntity a
         LEFT JOIN StringIndexEntity b
                   ON a.resourceType = b.resourceType AND a.resourceUuid = b.resourceUuid AND b.index_name = 'name'
WHERE a.resourceType = 'Patient'
ORDER BY b.index_value ASC
LIMIT 20 OFFSET 0
```

Query after this change:-
```
SELECT a.resourceUuid, a.serializedResource
FROM ResourceEntity a
         LEFT JOIN StringIndexEntity b
                   ON a.resourceUuid = b.resourceUuid AND b.index_name = 'name'
WHERE a.resourceType = 'Patient'
ORDER BY b.index_value ASC
LIMIT 20 OFFSET 0
```
Notice that `a.resourceType = b.resourceType` clause is not present after `ON`.

**Alternative(s) considered**
None

**Type**

**Checklist**

### DIFF
--- a/engine/src/main/java/com/google/android/fhir/search/MoreSearch.kt
+++ b/engine/src/main/java/com/google/android/fhir/search/MoreSearch.kt
@@ -248,7 +248,7 @@ private fun Search.getSortOrder(
           //  spotless:off
       """
       LEFT JOIN ${sortTableName.tableName} $tableAlias
-      ON $otherTable.resourceType = $tableAlias.resourceType AND $otherTable.resourceUuid = $tableAlias.resourceUuid AND $tableAlias.index_name = ?
+      ON $otherTable.resourceUuid = $tableAlias.resourceUuid AND $tableAlias.index_name = ?
       """
         //  spotless:on
         }
--- a/engine/src/test/java/com/google/android/fhir/search/SearchTest.kt
+++ b/engine/src/test/java/com/google/android/fhir/search/SearchTest.kt
@@ -1717,7 +1717,7 @@ class SearchTest {
         SELECT a.resourceUuid, a.serializedResource
         FROM ResourceEntity a
         LEFT JOIN StringIndexEntity b
-        ON a.resourceType = b.resourceType AND a.resourceUuid = b.resourceUuid AND b.index_name = ?
+        ON a.resourceUuid = b.resourceUuid AND b.index_name = ?
         WHERE a.resourceType = ?
         ORDER BY b.index_value ASC
                 """
@@ -1737,7 +1737,7 @@ class SearchTest {
         SELECT a.resourceUuid, a.serializedResource
         FROM ResourceEntity a
         LEFT JOIN StringIndexEntity b
-        ON a.resourceType = b.resourceType AND a.resourceUuid = b.resourceUuid AND b.index_name = ?
+        ON a.resourceUuid = b.resourceUuid AND b.index_name = ?
         WHERE a.resourceType = ?
         ORDER BY b.index_value DESC
                 """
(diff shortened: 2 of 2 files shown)


ndegwamartin (approved):

jingtang10:
great pr thanks @MJ1998!

one thing i want to clarify, do you know if this column in the string index table (and similarly in other index tables) https://github.com/google/android-fhir/blob/5efb955619cc1f0ca54e0ce4bee9b7c69360d67b/engine/src/main/java/com/google/android/fhir/db/impl/entities/StringIndexEntity.kt#L49C26-L50C34 can be removed?

jingtang10 (approved):

chandrashekar-s:
Sorry for commenting late here, based on the solution I see that we are going with the assumption that `Resource Id` is unique across all resource types. But, as per the [FHIR specification](https://www.hl7.org/fhir/resource.html#id) `The logical id is unique within the space of all resources of the same type on the same server`.  Just raising it here so that we are aware of this.

jingtang10:
> Sorry for commenting late here, based on the solution I see that we are going with the assumption that `Resource Id` is unique across all resource types. But, as per the [FHIR specification](https://www.hl7.org/fhir/resource.html#id) `The logical id is unique within the space of all resources of the same type on the same server`.  Just raising it here so that we are aware of this.

In our query we already use a uuid to join the two tables so the resource type join is superfluous.

The uuid is not the resource logical id. It's an id we create for each resource.

chandrashekar-s:
> > Sorry for commenting late here, based on the solution I see that we are going with the assumption that `Resource Id` is unique across all resource types. But, as per the [FHIR specification](https://www.hl7.org/fhir/resource.html#id) `The logical id is unique within the space of all resources of the same type on the same server`.  Just raising it here so that we are aware of this.
> 
> In our query we already use a uuid to join the two tables so the resource type join is superfluous.
> 
> The uuid is not the resource logical id. It's an id we create for each resource.

Oh I see. I had a look at the `ResourceEntity` table and the uuid field is unique and there is also a `ResourceId` column which I asume is the logical id. Thanks for clarifying.

PR #2565, opened 2024-06-08
Balance SQLite expression tree for logical operators (AND/OR) to lower the depth

### LINKED ISSUE #2561: SQLite crashes with 'Expression tree is too large (maximum depth 1000)'
**Describe the bug**
Filtering data from the db using the FhirEngine#search interface 
```
val  patientListIdsReferenceParamFilterCriteria = .... 
fhirEngine.search<CarePlan>{
    filter(CarePlan.SUBJECT, *patientListIds.toTypedArray())
}
```
with a long list of more than 500 patient ids, SQLite crashes with this error

```
android.database.sqlite.SQLiteException: Expression tree is too large (maximum depth 1000) (code 1 SQLITE_ERROR): , while compiling: SELECT a.resourceUuid, a.serializedResource
                                                                                                    FROM ResourceEntity a
                                                                                                    WHERE a.resourceType = ?
                                                                                                    AND a.resourceUuid IN (
                                                                                                    SELECT resourceUuid FROM ReferenceIndexEntity
2024-06-03 18:55:15.354 30656-31493 WM-WorkerWrapper        org...rtregister.fhircore.mwcoreDev  E  WHERE resourceType = ? AND index_name = ? AND (index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR 
2024-06-03 18:55:15.355 30656-31493 WM-WorkerWrapper        org...rtregister.fhircore.mwcoreDev  E  index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_val
2024-06-03 18:55:15.356 30656-31493 WM-WorkerWrapper        org...rtregister.fhircore.mwcoreDev  E  ue = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ? OR index_value = ?)
                                                                                                    )
```

**Component**
Engine library

**Additional context**

Yes

LZRS:

Fixes #2561 

**Description**
Recursively bifurcates the conditional params expressions to prevent occurences of SQLite expression tree exceeding depth of 1000, as suggested in this [comment](https://github.com/google/android-fhir/pull/2565#issuecomment-2243596001)

**Alternative(s) considered**
Chunking large expression list to limit 50 within parantheses to avoid crashing with `Expression tree is too large (maximum depth 1000)`, as described [here](https://stackoverflow.com/a/17032196)

**Type**
Enhancement Feature

**Checklist**


engine/src/main/java/com/google/android/fhir/search/filter/FilterCriterion.kt
```diff
@@ -86,15 +86,18 @@ internal sealed class FilterCriteria(
...
-    this.joinToString(
-      separator = " ${operation.logicalOperator} ",
-      prefix = if (size > 1) "(" else "",
-      postfix = if (size > 1) ")" else "",
-    ) {
-      if (it.params.size > 1) {
-        "(${it.condition})"
-      } else {
-        it.condition
+    this.chunked(50) { conditionParams ->
```
pld, line 89: @MJ1998 do have thoughts on what the chunk size should be?

engine/src/main/java/com/google/android/fhir/search/filter/FilterCriterion.kt
```diff
@@ -86,15 +86,28 @@ internal sealed class FilterCriteria(
...
+      .joinToString(separator = " ${operation.logicalOperator} ")
+
+  companion object {
+    /**
+     * Represents the number of [ConditionParam]s that can be wrapped within a bracket
+     *
+     * This is to prevent SQLite expression tree exceeding max depth of 1000 See
+     * https://www.sqlite.org/limits.html for Maximum Depth Of An Expression Tree
+     */
+    const val CONDITION_PARAMS_CHUNK_SIZE = 50
```
pld, line 111: another question on this value, don't we want this to be as high as possible? I'd think fewer chunks are easier to process, and fewer checks definitely means fewer iterations. Why not set this to the max of 1000 as a default?

jingtang10:
judging purely by the error message, my hypothesis is that the expression tree depth is O(n) for nested OR operators because the expression tree is constructed naively by parsing the OR operators sequentially. For example, if you nest 7 or statements you'd get:

```
  o
 / \
x   o
   / \
  x   o
     / \
    x   o
       / \
      x   o
         / \
        x   o
           / \
          x   x
```

where each `o` stands for one of the 6 OR operators and the 7 `x`s are the 7 values. This has depth 7.

But what you really want is actually this:

```
        o
      /   \
    o       o
   / \     / \
  o   o   o   o
 / \ / \ / \ / \
 x x x x x x x x
```

where the tree is more "balanced" and this has depth 4.

If my hypothesis of what causes the problem is correct, instead of trying to break the OR statements into chunks (and having to come up with a value), all you actually have to do is keep the tree balanced by splitting the top level OR statment at the middle of the list of params.

Does this make sense?

jingtang10:
there's no guarantee what i said is true @LZRS - i've done no testing or verification and depending on sqlite's implementation what i said could be complete garbage... so pls test and see if this is true :) (for example you could try to see if you'll hit the depth limit a bit later to bifurcate the tree instead of chunking the parameters)

LZRS:
> there's no guarantee what i said is true @LZRS - i've done no testing or verification and depending on sqlite's implementation what i said could be complete garbage... so pls test and see if this is true :) (for example you could try to see if you'll hit the depth limit a bit later to bifurcate the tree instead of chunking the parameters)

@jingtang10 I tested this out for 1000, 2000 and upto 5000 parameters, and it worked perfectly. I also went ahead a drafted an implementation in the PR

engine/src/main/java/com/google/android/fhir/search/filter/FilterCriterion.kt
```diff
@@ -85,16 +85,29 @@ internal sealed class FilterCriteria(
...
-      postfix = if (size > 1) ")" else "",
-    ) {
-      if (it.params.size > 1) {
-        "(${it.condition})"
-      } else {
-        it.condition
-      }
+  private fun List<ConditionParam<*>>.toQueryString(operation: Operation): String {
+    if (this.size <= 1) {
+      return map {
```
jingtang10, line 90: why use `map` if the size is <= 1 anyway? i think you can just do the lamba inside the `map` function directly since you know there's only one (or zero) item.

engine/src/main/java/com/google/android/fhir/search/filter/FilterCriterion.kt
```diff
@@ -85,16 +85,29 @@ internal sealed class FilterCriteria(
...
+        }
+        .firstOrNull()
+        ?: ""
     }
+
+    val mid = this.size / 2
+    val left = this.subList(0, mid).toQueryString(operation)
+    val right = this.subList(mid, this.size).toQueryString(operation)
+
+    return listOf(left, right)
```
jingtang10, line 105: can you just join left and right with the separator without going through all this? in line 89 you have guarantee that the list has 2 or more items. that means neither left nor right can have size 0.

engine/src/test/java/com/google/android/fhir/search/SearchTest.kt
```diff
@@ -2752,6 +2754,43 @@ class SearchTest {
       .inOrder()
   }
 
+  @Test
+  fun `search CarePlan filter with large list of patient reference`() {
```
jingtang10, line 2758: i think you do want to also test base cases... for example an empty list, and a list with 1, 2, (maybe add a 3 just to check when the list is not an even number and power of 2), 4 and 8 items.

by that point we can be confident the algorithm works as intended.

we can have a huge one like this - but it's actually not as useful in my view.

jingtang10 (changes requested):

engine/src/test/java/com/google/android/fhir/search/SearchTest.kt
aditya-07, line 2758: We should add this test in DatabaseImplTest to make sure that it always compiles and runs on the actual database.

aditya-07 (changes requested):
Can we add a DatabaseImplTest with

engine/src/test/java/com/google/android/fhir/search/SearchTest.kt
LZRS: The base case for 1 has been covered in other test cases in the class, I guess I could add for others

engine/src/test/java/com/google/android/fhir/search/SearchTest.kt
```diff
@@ -2752,6 +2754,43 @@ class SearchTest {
       .inOrder()
   }
 
+  @Test
```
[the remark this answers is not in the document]
LZRS: Not really sure if it's easily possible, but checking it out...

engine/src/test/java/com/google/android/fhir/search/SearchTest.kt
jingtang10: Well - I know that the base case for 1 is covered via other test cases. BUT it's much better to write unit tests for base cases than complex cases.

I would say that tests cases for 0, 1, 2, 3, 4 together is much better than a single test case for 10.

This is because you really want to isolate any cases and make sure tests are easy to debug, easy to fix. If the test case for 10 starts to fail, you'll spend a long time trying to debug. But if the test case fails for either of 1, 2, 3, or 4, you can almost immediately pin down the cause for failure.

engine/src/test/java/com/google/android/fhir/search/SearchTest.kt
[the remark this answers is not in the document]
jingtang10: I'm not really sure that this is a requirement for our API. This is an implementation detail in the db and our APIs don't really promise to the app how many patients the appp can join... i think maybe we should improve our documentation here but i am ok with not having this as a test. just my opinion.

engine/src/androidTest/java/com/google/android/fhir/db/impl/DatabaseImplTest.kt
```diff
@@ -5169,6 +5170,29 @@ class DatabaseImplTest {
       assertThat(localChangeResourceReferences.size).isEqualTo(locallyCreatedPatients.size)
     }
 
+  @Test
+  fun searchTasksForManyPatientsReturnCorrectly() = runBlocking {
+    val patients = (0..5001).map { Patient().apply { id = "task-patient-index-$it" } }
```
jingtang10, line 5175: There are 5002 patients and 5001 tasks. Is this test case deliberately checking this? I feel like if you really want to test the filter clause, it's better to query tasks linked to patient 0-4999 so at least you're filtering out 1 patient to know that the filter clause works.

engine/src/androidTest/java/com/google/android/fhir/db/impl/DatabaseImplTest.kt
```diff
@@ -5169,6 +5170,29 @@ class DatabaseImplTest {
...
+    val tasks =
+      (0..5000).map {
+        Task().apply {
+          id = "patient-$it-task"
+          `for` = Reference(patients[it])
+        }
+      }
+    database.insert(*tasks.toTypedArray())
+
+    val patientsSearchIdList =
```
jingtang10, line 5186: you can also limit this list to 2000 or 3000.

engine/src/main/java/com/google/android/fhir/search/filter/FilterCriterion.kt
```diff
@@ -86,28 +86,19 @@ internal sealed class FilterCriteria(
...
-        .firstOrNull()
-        ?: ""
-    }
-
-    val mid = this.size / 2
-    val left = this.subList(0, mid).toQueryString(operation)
-    val right = this.subList(mid, this.size).toQueryString(operation)
+    when {
+      this.isEmpty() ->
+        return "true" // In case empty, return true to match other conditions in the where clause
```
jingtang10, line 91: first of all, we should never get to this point i think.. something is wrong if we are here...

but let's say we are here, i think we should return "true" if the oeprator is and but "false" if the operator is or... for this to be a no-op.

this is because if the operator is or, and you return a true here, then the whole thing will evaluate to true... and that doesn't seem right to me.

engine/src/test/java/com/google/android/fhir/impl/FhirEngineImplTest.kt
```diff
@@ -723,6 +726,47 @@ class FhirEngineImplTest {
...
+        Patient,
+        TokenParamFilterCriterion.() -> Unit,
+      > {
+        { value = of(it.logicalId) }
+      }
+    val result2 =
+      fhirEngine.search<Patient> {
+        filter(TokenClientParam("_id"), *filterValues.toTypedArray())
+        println(getQuery().query)
+        println(getQuery().args)
```
jingtang10, line 757: remove

engine/src/test/java/com/google/android/fhir/impl/FhirEngineImplTest.kt
```diff
@@ -723,6 +726,47 @@ class FhirEngineImplTest {
...
+    val patient5 = Patient().apply { id = "patient-899" }
+    val patient6 = Patient().apply { id = "patient-883376" }
+    fhirEngine.create(patient1, patient2, patient3, patient4, patient5, patient6)
+
+    val result1 =
+      fhirEngine.search<Patient> {
+        filter(TokenClientParam("_id"), { value = of(patient3.logicalId) })
+      }
+    assertThat(result1).isNotEmpty()
+    assertThat(result1.size).isEqualTo(1)
```
jingtang10, line 744: not sure this block is useful.

engine/src/test/java/com/google/android/fhir/search/SearchTest.kt
jingtang10, line 1: good test cases in this file that really test what is being done in this pr - thanks lazarus!

engine/src/main/java/com/google/android/fhir/search/filter/FilterCriterion.kt
jingtang10: actually - can you change the code calling this function on line 61 in this file. so that when the list is empty, you don't call this function.

and then you can safely remove this case.

and then, you need to update the test case in searchtest.kt file on line 2807 to remove the `AND true`.

jingtang10 (changes requested):

engine/src/androidTest/java/com/google/android/fhir/db/impl/DatabaseImplTest.kt
LZRS: > you can also limit this list to 2000 or 3000.

I couldn't set limit the test to either 3000 or 2000, ran to another sqlite error on android 
```
android.database.sqlite.SQLiteException: too many SQL variables (code 1)
```
It's related to https://issuetracker.google.com/issues/73634057

For now, I just used a limit of 990

jingtang10: but 5000 worked? this is odd...

can you please do some more testing to see what's goign on here?

this is not blocking the pr though.

jingtang10 (approved):

engine/src/androidTest/java/com/google/android/fhir/db/impl/DatabaseImplTest.kt
```diff
@@ -5169,6 +5170,32 @@ class DatabaseImplTest {
...
+      patients.map<Patient, ReferenceParamFilterCriterion.() -> Unit> {
+        { value = "Patient/${it.logicalId}" }
+      }
+    val searchQuery =
+      Search(ResourceType.Task)
+        .apply { filter(Task.SUBJECT, *patientsSearchIdList.toTypedArray()) }
```
jingtang10, line 5192: this filter is checking that tasks are referencing patients 0-990 (inclusive) so that's 991 patients.

but you only have tasks for patients 0-989 (990 in total).

so the filter criteria is not really being tests because the filter criteria is broader than what you have in the database.

if you want to test it, make the filter more strict. for example, create N patients, and in the filter list N-1 IDs. and check that you only get N-1 results.

jingtang10 (changes requested):

engine/src/androidTest/java/com/google/android/fhir/db/impl/DatabaseImplTest.kt
LZRS: Initially the way they'd passed for the 5000, I had generated the bifurcated queries and ran them in separate sqlite db browser. That might be the difference

jingtang10 (approved):

aditya-07 (approved):

Issue #2077, opened 2023-07-18
Need to Support Minimum API level 25


mahmud6390:
**Describe the Issue**
Many of Health Worker of BRAC are using API level 25 devices.
**Would you like to work on the issue?**
Yes, can we change the minimum API level 25 from 26?

omarismail94:
The reason the Workflow library must be minSDK 26 is due to a run time dependency that requires it:
https://github.com/google/android-fhir/blob/master/workflow/build.gradle.kts#L123-L126

jingtang10:
Looking at James's comment https://github.com/hapifhir/hapi-fhir/pull/3977#issuecomment-1309466453 we might have a chance to actually drop the API level with new HAPI versions.

We should at least give it a go.

fredhersch:
@mahmud6390 For the initial application you are working on you should be able to use just the FHIREngine and Data Capture Libraries. These only require API 24.

Can you confirm this is a viable approach whilst we investigate whether the Workflow Library can be dropped to API 25 - based on the the comment from @jingtang10 

Thanks.

mahmud6390:
@fredhersch  Okey, We are downgread  the API version and let you share the status with other dependency logic.

fredhersch:
> @fredhersch we have tested to run fhircore app without workflow library. but it's showing many dependency error after build. If it's mandatory to use minimum sdk version 26 for FHIR app, we have to go to client with this message to update their devices.

Thanks @mahmud6390 If you are using FHIRCore, can you explore with the Ona team to discuss whether it's possible to build a version without Workflow Libraries? 

As per the thread above, @jingtang10 and @omarismail94 are looking into whether the depedency issues - which is what is driving API 26 requirement - can be resolved

pld:
@mahmud6390 the main reason [opensrp](https://github.com/opensrp/fhircore/) depends on the workflow lib is for $evaluate-measure, which is required for in-app reporting. We'd be open to a PR that wraps all of in-app reporting, including the inclusion of the workflow lib, in a feature-flag that we could control at the flavor level.

But, a future reason for including the workflow lib will be using $apply. We are planning to shift from our $lite (from before $apply existed in the Android FHIR SDK) to $apply, but do not have a confirmed timeline for that. When that happens we would have a much harder, probably non-removable, dependency on the workflow lib.

PR #2575, opened 2024-06-18
Update gradle to 8.8 and AGP to 8.5.0

jingtang10:

**Description**
Update gradle to 8.8 and AGP to 8.5.0

**Alternative(s) considered**
NA

**Type**
Builds

**Checklist**


gradle/wrapper/gradle-wrapper.properties
```diff
@@ -1,6 +1,6 @@
 distributionBase=GRADLE_USER_HOME
 distributionPath=wrapper/dists
-distributionUrl=https\://services.gradle.org/distributions/gradle-8.0-bin.zip
+distributionUrl=https\://services.gradle.org/distributions/gradle-8.6-bin.zip
```
aditya-07, line 3: Can we use the latest gradle version available?

aditya-07 (commented):
Check to see if additional changes are required.
Also, is there a reason if we are updating to these specific versions?

gradle/wrapper/gradle-wrapper.properties
jingtang10: done

buildSrc/build.gradle.kts
```diff
@@ -11,7 +11,7 @@ repositories {
 dependencies {
   implementation("com.diffplug.spotless:spotless-plugin-gradle:6.22.0")
 
-  implementation("com.android.tools.build:gradle:8.1.4")
+  implementation("com.android.tools.build:gradle:8.4.1")
```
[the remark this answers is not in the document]
jingtang10: done

jingtang10:
addressed comments
thanks aditya for reviewing

MJ1998 (commented):
Have you verified compatibility of other dependencies' with this agp? 
I guess our test suite should be enough ?

Update: Some tests failed which is a good sign!

jingtang10:
After taking a look at the failure:

```
Lint found 2 errors, 38 warnings. First failure:
[1185](https://github.com/google/android-fhir/actions/runs/9566698928/job/26372611822?pr=2575#step:5:1186)
[1186](https://github.com/google/android-fhir/actions/runs/9566698928/job/26372611822?pr=2575#step:5:1187)/home/runner/work/android-fhir/android-fhir/datacapture/src/main/java/com/google/android/fhir/datacapture/extensions/MoreLocalDates.kt:128: Error: Implicit cast from IsoChronology to Chronology requires API level 26 (current min is 24) [NewApi]
[1187](https://github.com/google/android-fhir/actions/runs/9566698928/job/26372611822?pr=2575#step:5:1188)    IsoChronology.INSTANCE,
[1188](https://github.com/google/android-fhir/actions/runs/9566698928/job/26372611822?pr=2575#step:5:1189)    ~~~~~~~~~~~~~~~~~~~~~~
```

and trying to fix it without updating min sdk version, I realised we do need the ISOChronology class in order to get the localized date time pattern... without using this class, it's possible to generate the date time formatter object that is localized, but we won't be able to get the actually localized pattern e.g. yyyy-mm-dd (by calling the api `getLocalizedDateTimePattern`)... but we do need this pattern itself as the hint text...

now an alternative is to not use that as hint text, but instead use a formatted date e.g. 2019-09-12... which might be ok but it will just diverge a bit from the behaviour we have when the entry format string is provided by the questionnaire.

so i think the easiest thing to do is actually update the minsdk.. which i think wasn't an issue only because of the build tool wasn't picking it up since the api was indeed introduced in api level 26... so this way we're only making what was already the case explicit... (device on 24 would have had problem with this already).

MJ1998:
We dropped the minSdk (for all libraries) to 24 due to this issue #2077.

pld:
Just to confirm, we're fine raising the min level to 26

jingtang10:
discussed today at the developers call and no objection so far to this api level change.

if we don't hear anything else we'll include this change in the next release

aditya-07 (approved):
LGTM.

PR #2598, opened 2024-07-03
Migrate the rest of androidx, kotlin, junit, and truth libraries to version catalog

jingtang10:

**Description**
Migrate the rest of androidx, kotlin, junit, and truth libraries to version catalog

**Alternative(s) considered**
NA

**Type**
Builds

**Checklist**


gradle/libs.versions.toml
```diff
@@ -3,25 +3,40 @@
...
+junit = "4.13.2"
+kotlin-stdlib = "1.9.22"
+kotlin-test = "1.9.22"
+kotlinx-coroutines = "1.7.2"
+truth = "1.1.5"
 
 [libraries]
 androidx-activity = { module = "androidx.activity:activity", version.ref = "androidx-acivity" }
 androidx-appcompat = { module = "androidx.appcompat:appcompat", version.ref = "androidx-appcompat" }
+androidx-arch-core-testing = { module = "androidx.arch.core:core-testing", version.ref = "androidx-arch-core" }
```
MJ1998, line 33: I think you are naming this to the library (and similarly some other libraries as well) name which makes sense. However since the order of libraries here is lexicographic, it ungroups this library from the group of libraries that fell under AndroidXTest before. 
Should we do something about it ? Grouping helps to locate libraries faster.

jingtang10: thanks jajoo - i think dependencies from the same group should be together.

this library is under `android.arch.core` and is not under `androidx.test` even though the artifact id is `core-testing`

is there an example of libraries that should be together but aren't in this file?

jingtang10: in the past, we did have a "testing libraries" grouping for libraries used as a test dependency in the gradle build files. i think that actually wouldn't make as much sense as it does now. with the version catalog now we're grouping libraries purely by their group name and artifact name which is definitive and we don't have to artifically separate libraries into different groups ourselves.

MJ1998: Makes sense.
Although I notice a bit of inconsistency in naming the library aliases. Can we stick to one format ?
Example for a library "group.name.ref" we always give an alias of "group-name".
What do you think ?

jingtang10: I think that's what I've tried to adhere to. With two notable exceptions: 1) in the case of redundant words I try to simplify them. For example we have androidx.fragment:fragment-testing which is simplified to androidx-fragment-testing 2) in the case of long group names quite often it is irrelevant to include the entire organisational namespace e.g. jetbrains. So I try to not include them. But androidx is useful to keep since it indicates it's a jetpack library and some of the libraries have quite generic names. For example you don't just want testing you want androidx-testing.

MJ1998 (approved):
LGTM

jingtang10:
thanks jajoo!

PR #2634, opened 2024-07-22
Rename install API to index for file in knowledge manager

jingtang10:

**Description**
Rename install API to index for file in knowledge manager

**Note** In this PR we also change the test dependency in the workflow module to the current version of knowledge manager instead of the published version. This makes sure we're testing against the latest Knowledge Manager APIs.

**Alternative(s) considered**
NA

**Type**
Code health

**Checklist**


knowledge/src/main/java/com/google/android/fhir/knowledge/KnowledgeManager.kt
```diff
@@ -49,6 +49,9 @@ import timber.log.Timber
  * - database: indexing knowledge artifacts stored in the local file system,
  * - file manager: managing files containing the knowledge artifacts, and
  * - NPM downloader: downloading from an NPM package server the knowledge artifacts.
```
ndegwamartin, line 51: Not introduced by the PR but could we rename these to correspond to the actual parameters and properties of the Knowledge Manager class?  _i.e. knowledgeDatabase: KnowledgeDatabase, npmFileManager: NpmFileManager_ e.t.c?

knowledge/src/main/java/com/google/android/fhir/knowledge/KnowledgeManager.kt
```diff
@@ -93,9 +96,18 @@ internal constructor(
...
+          0L,
+          fhirNpmPackage.canonical ?: "",
+          fhirNpmPackage.name,
+          fhirNpmPackage.version,
+          rootDirectory,
+        ),
+      )
     rootDirectory.listFiles()?.sorted()?.forEach { file ->
       try {
         val resource = jsonParser.parseResource(FileInputStream(file))
```
ndegwamartin, line 113: We can refactor the `KnowledgeManager.import(` method to internally user the newly created `KnowledgeManager.index(` method, it will remove the duplicate code blocks

knowledge/src/main/java/com/google/android/fhir/knowledge/KnowledgeManager.kt
```diff
@@ -116,9 +128,43 @@ internal constructor(
...
+    val resource =
+      withContext(Dispatchers.IO) {
+        try {
+          FileInputStream(file).use(jsonParser::parseResource)
+        } catch (exception: Exception) {
+          Timber.d(exception, "Unable to import file: $file. Parsing to FhirResource failed.")
+        }
+      }
+    when (resource) {
+      is Resource -> {
```
ndegwamartin, line 158: Do we want to instead check whether this is a `MetadataResource` instead? https://hl7.org/fhir/R5/metadataresource.html

Also note, the the `KnowledgeManager.indexResourceFile(` method called on the next line actually tries to cast to a `MetadataResource` with no fallback.

knowledge/src/main/java/com/google/android/fhir/knowledge/KnowledgeManager.kt
```diff
@@ -220,8 +229,28 @@ internal constructor(
     }
   }
 
-  private fun loadResource(resourceEntity: ResourceMetadataEntity): IBaseResource {
-    return jsonParser.parseResource(FileInputStream(resourceEntity.resourceFile))
+  /**  */
+  private suspend fun indexResourceFile(igId: Long?, resource: Resource, file: File): Long {
+    val metadataResource = resource as? MetadataResource
```
ndegwamartin, line 234: This safe cast can result to the `metadataResource` variable being null.  We should probably check for that since the subsequent processing may lead to the entry of a corrupt `ResourceMetadataEntity` into the Knowledge manager db.

knowledge/src/main/java/com/google/android/fhir/knowledge/KnowledgeManager.kt
```diff
@@ -93,9 +96,18 @@ internal constructor(
    * Checks if the [fhirNpmPackage] is present in DB. If necessary, populates the database with the
    * metadata of FHIR Resource from the provided [rootDirectory].
```
ndegwamartin, line 97: I find this documentation a bit confusing, the function doesn't seem to perform any checks (at least not directly)

knowledge/src/main/java/com/google/android/fhir/knowledge/KnowledgeManager.kt
ndegwamartin: On this I think we may have instances where we want to persist non MetadataResource types? @dubdabasoduba can comment on this.

Not sure if it is restricted by the FHIR Spec but the answer can impact the casting here.

knowledge/src/main/java/com/google/android/fhir/knowledge/KnowledgeManager.kt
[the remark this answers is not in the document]
jingtang10: added a test

jingtang10:
> @jingtang10 for you consideration, please check out [my comment on the Github ticket here](https://github.com/google/android-fhir/issues/2490#issuecomment-2245884031). If the recommendations there make sense then this might be an appropriate PR to piggy back the enhancement on.

thanks martin - that will be handled in a subsequent pr - let's keep the scope of this pr smaller.

knowledge/src/main/java/com/google/android/fhir/knowledge/KnowledgeManager.kt
```diff
@@ -46,9 +43,20 @@ import timber.log.Timber
  * individually as JSON files or from FHIR NPM packages.
  *
  * Coordinates the management of knowledge artifacts by using the three following components:
- * - database: indexing knowledge artifacts stored in the local file system,
- * - file manager: managing files containing the knowledge artifacts, and
- * - NPM downloader: downloading from an NPM package server the knowledge artifacts.
+ * - knowledgeDatabase: indexing knowledge artifacts stored in the local file system,
+ * - npmFileManager: managing files containing the knowledge artifacts, and
+ * - npmPackageDownloader: downloading from an NPM package server the knowledge artifacts.
```
aditya-07, line 48:
```suggestion
 * - npmPackageDownloader: downloading the knowledge artifacts from an NPM package server .
```
a bit simpler to read?

knowledge/src/main/java/com/google/android/fhir/knowledge/KnowledgeManager.kt
```diff
@@ -90,35 +100,72 @@ internal constructor(
...
+   *
+   * This operation does not make a copy of the knowledge artifact, nor does it checksum the content
+   * of the file. Therefore, it cannot be guaranteed that subsequent retrievals of the knowledge
+   * artifact will produce the same result. Applications using this function must be aware of the
+   * risk of the content of the file being modified or corrupt, potentially resulting in incorrect
+   * or inaccurate result of decision support or measure evaluation.
+   *
+   * Use this API for knowledge artifacts in immutable files (e.g. in the app's `assets` folder).
+   *
+   * resources already indexed and resources already imported by ig?
```
aditya-07, line 154: Should this line be removed?
```suggestion
```

knowledge/src/main/java/com/google/android/fhir/knowledge/KnowledgeManager.kt
```diff
@@ -220,8 +230,37 @@ internal constructor(
...
+        null
+      }
+    }
+
+  /**
+   * Parses and returns the content of a file containing a FHIR metadata resource in JSON, or null
+   * if the file does not contain a FHIR metadata resource.
+   */
+  private suspend fun readMetadataResourceOrNull(file: File) =
+    readResourceOrNull(file) as? MetadataResource
```
aditya-07, line 252: It will throw a `ClassCastException` if File contains a non MetadataResource FHIR resource. It might be better to check for  `MetadataResource` before casting and returning null otherwise.

jingtang10: I've done some testing and I'm pretty sure the `as?` would handle this case gracefully by returning `null`. am i missing something?

aditya-07: Yup, you are right. Didn't realize that `as?` returns [null on failure ](https://kotlinlang.org/docs/typecasts.html#safe-nullable-cast-operator).

workflow/build.gradle.kts
```diff
@@ -114,8 +114,18 @@ dependencies {
...
+
+  configurations.all {
+    if (name.contains("test", ignoreCase = true)) {
+      resolutionStrategy.dependencySubstitution {
+        // To test the workflow library against the latest Knowledge Manager APIs, substitute the
+        // dependency on the released Knowledge Manager library with the current build.
+        substitute(module(Dependencies.androidFhirKnowledge)).using(project(":knowledge"))
+      }
+    }
+  }
```
aditya-07, line 128: Approving for now. But we should explore if there is a cleaner way to do this.

aditya-07 (approved):
LGTM. 👍

PR #2644, opened 2024-07-27
Refactor navigation view UI state

FikriMilano:

Fixes #[issue number]

**Description**

**Alternative(s) considered**

**Type**
Code health

**Checklist**


datacapture/src/main/java/com/google/android/fhir/datacapture/views/NavigationViewHolder.kt
```diff
@@ -48,9 +48,7 @@ class NavigationViewHolder(itemView: View) : RecyclerView.ViewHolder(itemView) {
       is QuestionnaireNavigationViewUIState.Enabled -> {
         visibility = View.VISIBLE
         isEnabled = true
-        if (navigationViewState.labelText.isNullOrBlank().not() && this is Button) {
-          text = navigationViewState.labelText
-        }
```
jingtang10, line 53: what is actually the use of the labelText field? looking through our code we don't ever set it to be different from the corresponding button's text in the xml layout files? so why don't we just remove this field altogether?

datacapture/src/main/res/values/strings.xml
```diff
@@ -10,7 +10,7 @@
     <string name="review_answers_text">Review answers</string>
     <string name="button_review">Review</string>
     <string name="submit_questionnaire">Submit</string>
-    <string name="cancel_questionnaire">@android:string/cancel</string>
+    <string name="cancel_questionnaire">Cancel</string>
```
jingtang10, line 13: why this change?

jingtang10: the reason to use `@android:string/cancel` is that it's already internationalised

datacapture/src/main/java/com/google/android/fhir/datacapture/views/NavigationViewHolder.kt
jingtang10: i see - we have all these button text set in the xml layout files but also rewrite them in code :/

i think i'd like either 1) set all text in code or 2) set most text in xml, but only override in code in the case of submit button.

what do you thinik? i slightly prefer 2....

jingtang10: why not just hide the button if it shouldn't be displayed? hiding the text to have a button that shows no text looks odd and doesn't really fulfill the requirement (the user can still click it right?)

jingtang10: shouldn't you just introduce another QuestionnaireNavigationViewUIState in this case and call it loading?

PR #2657, opened 2024-08-16
Add missing response items to repeated group child items

### LINKED ISSUE #2655: Matching of the correct pair of qItem and qrItem fails for repeated groups
**Describe the bug**

This newly introduced [requirement](https://github.com/google/android-fhir/blob/01cc80a8dc696c90a2b7f1ee6d78068276ffd17f/datacapture/src/main/java/com/google/android/fhir/datacapture/extensions/MoreQuestionnaires.kt#L202) through https://github.com/google/android-fhir/issues/2618, fails and throws a `java.lang.IllegalArgumentException: Failed requirement` when loading a Questionnaire that has repeated groups, and populating it with values from the latest QuestionnaireResponse.
 
The exception is thrown because when matching qItem and qrItem, the number of questionnaire items isn't equal to the number of questionnaire response items. 

For repeated groups this observed in the following scenarios: 

a. **If a Questionnaire has a repeated group with one set of repeated items/fields**:
If a field didn't have an answer because it was optional or it was hidden by an `enableWhen` expression then **it is  excluded from the Questionnaire response resulting in the QR with less items**

b. **If a Questionnaire has a repeated group with several repeated items/fields**: 
The repeated group items are put in the answerItems in the QR and when iterating and pairing the qItems will have a size of the number of fields in the group while it appears in the iteration, **the repeated group qrItems aren't being correctly retrieved** 
 

**Component**
 SDC library

**To Reproduce**
See additional context in comments below

**Expected behavior**

**Screenshots**

**Additional context**

FikriMilano:

Fixes #2655

**Description**
If you have a questionnaire item inside a repeated group, but the related response item is missing, then you call `addMissingResponseItems`. The missing response item is still missing.

An exception will show if you try to call `forEachItemPair` function where you need to have the exact same size between questionnaire items and response items.

This PR fixes that.

**Alternative(s) considered**
N/A

**Type**
Bug fix

**Checklist**

### DIFF
--- a/datacapture/src/main/java/com/google/android/fhir/datacapture/QuestionnaireViewModel.kt
+++ b/datacapture/src/main/java/com/google/android/fhir/datacapture/QuestionnaireViewModel.kt
@@ -419,7 +419,8 @@ internal class QuestionnaireViewModel(application: Application, state: SavedStat
    * Adds empty [QuestionnaireResponseItemComponent]s to `responseItems` so that each
    * [QuestionnaireItemComponent] in `questionnaireItems` has at least one corresponding
    * [QuestionnaireResponseItemComponent]. This is because user-provided [QuestionnaireResponse]
-   * might not contain answers to unanswered or disabled questions. This function should only be used for unpacked questionnaire.
+   * might not contain answers to unanswered or disabled questions. This function should only be
+   * used for unpacked questionnaire.
    */
   @VisibleForTesting
   internal fun addMissingResponseItems(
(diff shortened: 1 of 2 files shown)


datacapture/src/main/java/com/google/android/fhir/datacapture/QuestionnaireViewModel.kt
```diff
@@ -422,7 +422,8 @@ internal class QuestionnaireViewModel(application: Application, state: SavedStat
    * might not contain answers to unanswered or disabled questions. Note : this only applies to
    * [QuestionnaireItemComponent]s nested under a group.
    */
-  private fun addMissingResponseItems(
+  @VisibleForTesting
+  internal fun addMissingResponseItems(
```
jingtang10, line 426: can you 1) clarify if this function is editing questionnaire responses before or after we pack repeated groups, if it's editing an already packed questionnaire, then there will only be at most one questionnaire response item component for each quesetionnaire item component.

2) is this still valid? `Note : this only applies to [QuestionnaireItemComponent]s nested under a group.`

jingtang10 (changes requested):

PR #2669, opened 2024-09-07
Implement Parallelized map and optimize Database search API

### LINKED ISSUE #2668: Optimize the Database Search API
Did some profiling on [this DatabaseImpl method here](https://github.com/google/android-fhir/blob/master/engine/src/main/java/com/google/android/fhir/db/impl/DatabaseImpl.kt#L204-L212) that has the below code:

```kotlin
  override suspend fun <R : Resource> search(
    query: SearchQuery,
  ): List<ResourceWithUUID<R>> {
    return db.withTransaction {
      resourceDao.getResources(SimpleSQLiteQuery(query.query, query.args.toTypedArray())).map {
        ResourceWithUUID(it.uuid, iParser.parseResource(it.serializedResource) as R)
      }
    }
  }
```
The mapping block takes a lot of time since each serialized resource has to be deserialized sequentially for every element in the list returned from the database. The time taken is linear and the more the db results returned the longer the API takes to return a result.

 It should be possible to optimize this block by introducing a parallelized implementation e.g. Using _coroutines_ and _async_ within each iteration and then collecting the results and get a significant improvement. 

Instead of letting the _async_ launch with the current dispatcher(probably IO at that point) we could instead switch to the Default dispatcher since it is a computationally expensive task. This will however require us to create a new JSONParser object per iteration since it is not thread safe.

**Additional context**

Yeah

ndegwamartin:

Fixes #2668 

**Description**
Optimizes the DatabaseImpl search API's _serialized FHIR Resource to HAPI FHIR Structure_ mapping block by introducing a parallelized implementation that uses async coroutines for each mapping iteration.

**Alternative(s) considered**

**Type**
Enhancement

**Checklist**

### DIFF
--- a/engine/src/main/java/com/google/android/fhir/Util.kt
+++ b/engine/src/main/java/com/google/android/fhir/Util.kt
@@ -26,9 +26,6 @@ import java.time.ZoneId
 import java.time.format.DateTimeFormatter
 import java.util.Date
 import java.util.Locale
-import kotlinx.coroutines.async
-import kotlinx.coroutines.awaitAll
-import kotlinx.coroutines.coroutineScope
 import org.hl7.fhir.r4.model.OperationOutcome
 import org.hl7.fhir.r4.model.Resource
 import org.hl7.fhir.r4.model.ResourceType
@@ -85,8 +82,3 @@ internal const val ucumUrl = "http://unitsofmeasure.org"
 
 internal fun percentOf(value: Number, total: Number) =
   if (total == 0) 0.0 else value.toDouble() / total.toDouble()
-
-/** Implementation of a parallelized map */
-suspend fun <A, B> Iterable<A>.pmap(f: suspend (A) -> B): List<B> = coroutineScope {
-  map { async { f(it) } }.awaitAll()
-}
--- a/engine/src/main/java/com/google/android/fhir/Util.kt
+++ b/engine/src/main/java/com/google/android/fhir/Util.kt
@@ -26,6 +26,10 @@ import java.time.ZoneId
 import java.time.format.DateTimeFormatter
 import java.util.Date
 import java.util.Locale
+import kotlinx.coroutines.Dispatchers
+import kotlinx.coroutines.async
+import kotlinx.coroutines.awaitAll
+import kotlinx.coroutines.coroutineScope
 import org.hl7.fhir.r4.model.OperationOutcome
 import org.hl7.fhir.r4.model.Resource
 import org.hl7.fhir.r4.model.ResourceType
@@ -67,6 +71,11 @@ internal fun Resource.isUploadSuccess(): Boolean {
   val outcome: OperationOutcome = this as OperationOutcome
   return outcome.issue.isNotEmpty() &&
     outcome.issue.all { it.severity.equals(OperationOutcome.IssueSeverity.INFORMATION) }
+}
+
+/** Implementation of a parallelized map for CPU intensive tasks */
+suspend fun <A, B> Iterable<A>.pmapCPU(f: suspend (A) -> B): List<B> = coroutineScope {
+  map { async(Dispatchers.Default) { f(it) } }.awaitAll()
 }
 
 internal class OffsetDateTimeTypeAdapter : TypeAdapter<OffsetDateTime>() {
--- a/engine/src/main/java/com/google/android/fhir/Util.kt
+++ b/engine/src/main/java/com/google/android/fhir/Util.kt
@@ -26,7 +26,7 @@ import java.time.ZoneId
 import java.time.format.DateTimeFormatter
 import java.util.Date
 import java.util.Locale
-import kotlinx.coroutines.Dispatchers
+import kotlinx.coroutines.CoroutineDispatcher
 import kotlinx.coroutines.async
 import kotlinx.coroutines.awaitAll
 import kotlinx.coroutines.coroutineScope
@@ -73,10 +73,11 @@ internal fun Resource.isUploadSuccess(): Boolean {
     outcome.issue.all { it.severity.equals(OperationOutcome.IssueSeverity.INFORMATION) }
 }
 
-/** Implementation of a parallelized map for CPU intensive tasks */
-suspend fun <A, B> Iterable<A>.pmapCPU(f: suspend (A) -> B): List<B> = coroutineScope {
-  map { async(Dispatchers.Default) { f(it) } }.awaitAll()
-}
+/** Implementation of a parallelized map */
+suspend fun <A, B> Iterable<A>.pmap(dispatcher: CoroutineDispatcher, f: suspend (A) -> B): List<B> =
+  coroutineScope {
+    map { async(dispatcher) { f(it) } }.awaitAll()
+  }
 
 internal class OffsetDateTimeTypeAdapter : TypeAdapter<OffsetDateTime>() {
   override fun write(out: JsonWriter, value: OffsetDateTime) {
--- a/engine/src/main/java/com/google/android/fhir/Util.kt
+++ b/engine/src/main/java/com/google/android/fhir/Util.kt
@@ -26,6 +26,9 @@ import java.time.ZoneId
 import java.time.format.DateTimeFormatter
 import java.util.Date
 import java.util.Locale
+import kotlinx.coroutines.async
+import kotlinx.coroutines.awaitAll
+import kotlinx.coroutines.coroutineScope
 import org.hl7.fhir.r4.model.OperationOutcome
 import org.hl7.fhir.r4.model.Resource
 import org.hl7.fhir.r4.model.ResourceType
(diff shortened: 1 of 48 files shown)


jingtang10 (commented):
great work thanks @ndegwamartin!

ndegwamartin:
**Device:** Physical, Samsung Galaxy Active Tab 2
**Mode :** Benchmarking with Kotlin system Timing's `measureTimeMillis`
**Scope:** [Database search API method search](https://github.com/google/android-fhir/blob/6ad0a6751fea3e53d9e2be327ca205cca1112af8/engine/src/main/java/com/google/android/fhir/db/impl/DatabaseImpl.kt#L226) 

**Optimization:** None

**Run 1**
| Resource Type | Total Records | Timetaken(seconds)| DB Query(seconds) |
| -------- | ------- | ----------------------------|-----------------|
| Group | ~1K | 8 | ~0.2|
| Task | ~17K | ~24 | ~1.8|
| Patient | ~11K | ~456 | ~1.3|

**Run 2**
| Resource Type | Total Records | Timetaken(seconds)| DB Query(seconds) |
| -------- | ------- | ----------------------------|-----------------|
| Group | ~1K | ~2 | ~0.1 |
| Task | ~17K | ~22| ~1.7|
| Patient | ~11K | ~472 | ~1.2|

----------------------------------------------------------------------------
 
**Optimization:** Using _async_ with parent context (usually `Dispatchers.IO`)

**Run 1**
| Resource Type | Total Records | Timetaken(seconds)| DB Query(seconds) |
| -------- | ------- | ----------------------------|-----------------|
| Group | ~1K | 4.8 | ~0.2 |
| Task | ~17K | ~24 | ~1.7 |
| Patient | ~11K | ~450 | ~1.3 |

**Run 2**
| Resource Type | Total Records | Timetaken(seconds)| DB Query(seconds) |
| -------- | ------- | ----------------------------|-----------------|
| Group | ~1K | ~2 | ~0.1 |
| Task | ~17K | ~24| ~1.7 |
| Patient | ~11K | ~455 | ~1.3|

----------------------------------------------------------------------------

**Optimization:** Using _async_ with `Dispatchers.Default` . 
(_Note - Threads safety of the FHIR JsonParser is achieved through creating a new instance for each loop_) 

**Run 1**
| Resource Type | Total Records | Timetaken(seconds)| DB Query(seconds) |
| -------- | ------- | ----------------------------|-----------------|
| Group | ~1K | ~5 | ~0.2 |
| Task | ~17K | ~5.4 | ~1.8 |
| Patient | ~11K | ~208 | ~1.4 |

**Run 2**
| Resource Type | Total Records | Timetaken(seconds)| DB Query(seconds) |
| -------- | ------- | ----------------------------|-----------------|
| Group | ~1K | ~0.5 | ~0.1 |
| Task | ~17K | ~5 | ~1.7 |
| Patient | ~11K | ~204 | ~1.3|

**Note** - The tests were carried out in a QA test environment. In the real world Patients would be more than Groups (i.e. Patients = ~10 x No. of Groups ) and Tasks would be even more than Patients (i.e Tasks = ~30 x No. of Patients)

ndegwamartin:
**Full specs of the device:**
- Samsung Galaxy Tab Active2
- Android 9 (28)
- 3GB Memory

engine/src/main/java/com/google/android/fhir/db/impl/DatabaseImpl.kt
```diff
@@ -460,6 +470,11 @@ internal class DatabaseImpl(
   }
 }
 
+/** Implementation of a parallelized map */
+suspend fun <A, B> Iterable<A>.pmap(f: suspend (A) -> B): List<B> = coroutineScope {
```
MJ1998, line 474: can you rename pmap to something which recommends to pass functions doing CPU intensive work.
May be "pmapCPU" ?

ndegwamartin: Yeah makes total sense because of the Dispatcher constraint

ndegwamartin: I had restricted it for use in the DB search API class but with the rename I could potentially move it out to the generic Utils class for reuse elsewhere.

FikriMilano (approved):
Looks good!

MJ1998 (changes requested):
What happens when you don't parallelize the parsing but just run the parsing in the Default Dispatcher ?

A new function:-
```
fun IParser.parseResourceInCPU(resourceString: String) = 
    withContext(Dispatchers.Default) { parseResource(resourceString) }
```

and replace all parseResource with this function. 

I am surprised that parallelizing (second optimization) did not improve the search api performance. If parallelizing is not improving then do we really need it ?

ndegwamartin:
**Device:** Physical, Samsung Galaxy Active Tab 2
**Mode :** Benchmarking with Kotlin system Timing's `measureTimeMillis`
**Scope:** [Database search API method search - (Deserialization/mapping)](https://github.com/google/android-fhir/blob/6ad0a6751fea3e53d9e2be327ca205cca1112af8/engine/src/main/java/com/google/android/fhir/db/impl/DatabaseImpl.kt#L226) 

**Optimization:** Using IParser.parseResourceInCPU() with context (usually `Dispatchers.Default`)

**Run 1**
| Resource Type | Total Records | Timetaken(seconds)|
| -------- | ------- | ----------------------------|
| Group | ~1K | ~5 |
| Task | ~17K | ~29 |
| Patient | ~11K | ~457|

**Run 2**
| Resource Type | Total Records | Timetaken(seconds)|
| -------- | ------- | ----------------------------|
| Group | ~1K | ~2 |
| Task | ~17K | ~28 |
| Patient | ~11K | ~461 |


cc @MJ1998

ndegwamartin:
**Device:** Physical, Samsung Galaxy Active Tab 2
**Mode :** Benchmarking with Kotlin system Timing's `measureTimeMillis`
**Scope:** [Database search API method search - (Deserialization/mapping)](https://github.com/google/android-fhir/blob/6ad0a6751fea3e53d9e2be327ca205cca1112af8/engine/src/main/java/com/google/android/fhir/db/impl/DatabaseImpl.kt#L226) 

**Optimization:** Using _IParser.parseResourceInCPU()_ with context `Dispatchers.Default`

**Run 1**
| Resource Type | Total Records | Timetaken(seconds)|
| -------- | ------- | ----------------------------|
| Group | ~1K | ~5 |
| Task | ~17K | ~29 |
| Patient | ~11K | ~457|

**Run 2**
| Resource Type | Total Records | Timetaken(seconds)|
| -------- | ------- | ----------------------------|
| Group | ~1K | ~2 |
| Task | ~17K | ~28 |
| Patient | ~11K | ~461 |


cc @MJ1998

MJ1998:
So its only when we do both - parallelize and use Default dispatcher - we see performance improvement, correct ?
@ndegwamartin

ndegwamartin:
> So its only when we do both - parallelize and use Default dispatcher - we see performance improvement, correct ? @ndegwamartin

Correct.

MJ1998:
Okay so I think using Dispatchers.IO for CPU-intensive tasks might be less efficient due to thread type mismatch.

engine/src/main/java/com/google/android/fhir/Util.kt
```diff
@@ -69,6 +73,11 @@ internal fun Resource.isUploadSuccess(): Boolean {
     outcome.issue.all { it.severity.equals(OperationOutcome.IssueSeverity.INFORMATION) }
 }
 
+/** Implementation of a parallelized map for CPU intensive tasks */
+suspend fun <A, B> Iterable<A>.pmapCPU(f: suspend (A) -> B): List<B> = coroutineScope {
```
ellykits, line 77: Can we consider renaming this to `pmap` and providing the `Dispatcher` as a function parameter? default to `Dispatchers.Default`

jingtang10 (approved):
thanks @ndegwamartin for all the work! great results!

engine/src/main/java/com/google/android/fhir/Util.kt
ndegwamartin: I think instead of having a default we should have the user to explicitly specify the _Dispatcher_ so that they are forced to align it with their exact use case. Refactoring as such.

engine/src/test/java/com/google/android/fhir/UtilTest.kt
```diff
@@ -111,6 +115,24 @@ class UtilTest : TestCase() {
...
+      squaredNumberList =
+        numberList.pmap(Dispatchers.Default) {
+          delay(1000L)
+          it * 2
+        }
+    }
+    assertTrue(squaredNumberList.isNotEmpty())
+    assertThat(squaredNumberList.first()).isEqualTo(4)
+    assertThat(squaredNumberList.last()).isEqualTo(6)
+    assertTrue(timeTaken < 2000)
```
jingtang10, line 133: don't do this. is there a non flakey way to test parallelism? if not i'm happy for you to not include a test.

jingtang10 (approved):

engine/src/test/java/com/google/android/fhir/UtilTest.kt
ndegwamartin: I'm okay with removing it, however unless the system it runs on is actually unable to multithread then it would never fail?

jingtang10: please let me know once this is resolved.

ndegwamartin: Done!

MJ1998 (approved):
LGTM.

PR #2672, opened 2024-09-12
Add Launch and Submission Timestamps to QR

### LINKED ISSUE #2372: Ability to Capture Questionnaire Launch and Submit Timestamps
We would like to track how long it takes agents in the field to fill out questionnaires. This is for process auditing purposes and to be able to optimise data collection activities and processes.

- Capture the timestamp of when a questionnaire is launched.
- Capture the timestamp of when a questionnaire is submitted.
- The timestamps can then be loaded to the questionnaire response for processing post-submit.

**Additional context**

hamza-vd:

Fixes #2372 

**Description**

**Alternative(s) considered**

**Type**

**Checklist**

### DIFF
--- a/datacapture/src/main/java/com/google/android/fhir/datacapture/extensions/MoreQuestionnaireResponses.kt
+++ b/datacapture/src/main/java/com/google/android/fhir/datacapture/extensions/MoreQuestionnaireResponses.kt
@@ -16,8 +16,13 @@
 
 package com.google.android.fhir.datacapture.extensions
 
+import org.hl7.fhir.r4.model.DateTimeType
+import org.hl7.fhir.r4.model.Extension
 import org.hl7.fhir.r4.model.Questionnaire
 import org.hl7.fhir.r4.model.QuestionnaireResponse
+
+internal const val EXTENSION_LAUNCH_TIMESTAMP: String =
+  "http://github.com/google-android/questionnaire-launch-timestamp"
 
 /** Pre-order list of all questionnaire response items in the questionnaire. */
 val QuestionnaireResponse.allItems: List<QuestionnaireResponse.QuestionnaireResponseItemComponent>
@@ -154,3 +159,21 @@ private fun unpackRepeatedGroups(
     listOf(questionnaireResponseItem)
   }
 }
+
+/** Adds a launch timestamp as an extension to the Questionnaire Response */
+internal var QuestionnaireResponse.launchTimestamp: DateTimeType?
+  get() {
+    val extension = this.extension.firstOrNull { it.url == EXTENSION_LAUNCH_TIMESTAMP }
+    return extension?.value as? DateTimeType
+  }
+  set(value) {
+    val noLaunchTimeStampExists = this.extension.none { it.url == EXTENSION_LAUNCH_TIMESTAMP }
+    if (noLaunchTimeStampExists) {
+      this.extension.add(
+        Extension(
+          EXTENSION_LAUNCH_TIMESTAMP,
+          value,
+        ),
+      )
+    }
+  }
--- a/datacapture/src/test/java/com/google/android/fhir/datacapture/QuestionnaireViewModelParameterizedTest.kt
+++ b/datacapture/src/test/java/com/google/android/fhir/datacapture/QuestionnaireViewModelParameterizedTest.kt
@@ -183,8 +183,23 @@ class QuestionnaireViewModelParameterizedTest(
     val printer: IParser = FhirContext.forR4().newJsonParser()
 
     fun <T : IBaseResource> assertResourceEquals(actual: T, expected: T) {
-      assertThat(printer.encodeResourceToString(actual))
-        .isEqualTo(printer.encodeResourceToString(expected))
+      if (actual is QuestionnaireResponse && expected is QuestionnaireResponse) {
+        val actualResponse = (actual as QuestionnaireResponse)
+        val expectedResponse =
+          (expected as QuestionnaireResponse).apply {
+            extension.add(
+              actualResponse.extension.firstOrNull { extension ->
+                extension.url == "http://github.com/google-android/questionnaire-launch-timestamp"
+              },
+            )
+            authored = actualResponse.authored
+          }
+        assertThat(printer.encodeResourceToString(actualResponse))
+          .isEqualTo(printer.encodeResourceToString(expectedResponse))
+      } else {
+        assertThat(printer.encodeResourceToString(actual))
+          .isEqualTo(printer.encodeResourceToString(expected))
+      }
     }
 
     @JvmStatic
--- a/datacapture/src/test/java/com/google/android/fhir/datacapture/QuestionnaireViewModelTest.kt
+++ b/datacapture/src/test/java/com/google/android/fhir/datacapture/QuestionnaireViewModelTest.kt
@@ -651,7 +651,7 @@ class QuestionnaireViewModelTest {
 
     val viewModel = createQuestionnaireViewModel(questionnaire, questionnaireResponse)
 
-    runTest { assertResourceEquals(questionnaireResponse, viewModel.getQuestionnaireResponse()) }
+    runTest { assertResourceEquals(viewModel.getQuestionnaireResponse(), questionnaireResponse) }
   }
 
   @Test
@@ -7514,8 +7514,23 @@ class QuestionnaireViewModelTest {
     val printer: IParser = FhirContext.forR4().newJsonParser()
 
     fun <T : IBaseResource> assertResourceEquals(actual: T, expected: T) {
-      assertThat(printer.encodeResourceToString(actual))
-        .isEqualTo(printer.encodeResourceToString(expected))
+      if (actual is QuestionnaireResponse && expected is QuestionnaireResponse) {
+        val actualResponse = (actual as QuestionnaireResponse)
+        val expectedResponse =
+          (expected as QuestionnaireResponse).apply {
+            extension.add(
+              actualResponse.extension.firstOrNull { extension ->
+                extension.url == "http://github.com/google-android/questionnaire-launch-timestamp"
+              },
+            )
+            authored = actualResponse.authored
+          }
+        assertThat(printer.encodeResourceToString(actualResponse))
+          .isEqualTo(printer.encodeResourceToString(expectedResponse))
+      } else {
+        assertThat(printer.encodeResourceToString(actual))
+          .isEqualTo(printer.encodeResourceToString(expected))
+      }
     }
   }
 }
--- a/datacapture/src/test/java/com/google/android/fhir/datacapture/extensions/MoreQuestionnaireResponsesTest.kt
+++ b/datacapture/src/test/java/com/google/android/fhir/datacapture/extensions/MoreQuestionnaireResponsesTest.kt
@@ -20,6 +20,7 @@ import ca.uhn.fhir.context.FhirContext
 import ca.uhn.fhir.context.FhirVersionEnum
 import com.google.common.truth.Truth.assertThat
 import org.hl7.fhir.r4.model.BooleanType
+import org.hl7.fhir.r4.model.DateTimeType
 import org.hl7.fhir.r4.model.Questionnaire
 import org.hl7.fhir.r4.model.Questionnaire.QuestionnaireItemComponent
 import org.hl7.fhir.r4.model.QuestionnaireResponse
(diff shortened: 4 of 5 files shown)


datacapture/src/main/java/com/google/android/fhir/datacapture/extensions/MoreQuestionnaireResponses.kt
```diff
@@ -154,3 +159,21 @@ private fun unpackRepeatedGroups(
     listOf(questionnaireResponseItem)
   }
 }
+
+/** Adds a launch timestamp as an extension to the Questionnaire Response */
```
jingtang10, line 163: can you clarify in the doc that we only add the launch timestamp if there isn't one at the moment?

jingtang10: or should we make this extension repeated so that each time you edit the questionnaire response it is recorded.

useful for audit purposes?

datacapture/src/test/java/com/google/android/fhir/datacapture/extensions/MoreQuestionnaireResponsesTest.kt
```diff
@@ -695,4 +696,36 @@ class MoreQuestionnaireResponsesTest {
...
+
+  @Test
+  fun `should add launchTimestamp`() {
+    val questionnaireResponse = QuestionnaireResponse()
+    val dateTimeType = DateTimeType("2024-07-05T00:00:00Z")
+    questionnaireResponse.launchTimestamp = dateTimeType
+
+    val launchTimestamp = questionnaireResponse.launchTimestamp
+    assertThat(launchTimestamp).isNotNull()
+    assertThat(dateTimeType).isEqualTo(launchTimestamp)
```
jingtang10, line 708: these 3 lines can be collapsed into 1. no need to check if it's not null if you're going to compare it anyway

datacapture/src/main/java/com/google/android/fhir/datacapture/extensions/MoreQuestionnaireResponses.kt
jingtang10: OR if we really want to only keep one (to be a bit more consistent with the authored fied) we can rename this extension to `lastLaunched`?

datacapture/src/test/java/com/google/android/fhir/datacapture/extensions/MoreQuestionnaireResponsesTest.kt
```diff
@@ -695,4 +696,36 @@ class MoreQuestionnaireResponsesTest {
...
+  }
+
+  @Test
+  fun `launchTimestamp should be null when not added`() {
+    val questionnaireResponse = QuestionnaireResponse()
+
+    val launchTimestamp = questionnaireResponse.launchTimestamp
+    assertThat(launchTimestamp).isNull()
```
jingtang10, line 716: same here

datacapture/src/test/java/com/google/android/fhir/datacapture/extensions/MoreQuestionnaireResponsesTest.kt
```diff
@@ -695,4 +696,36 @@ class MoreQuestionnaireResponsesTest {
...
+  fun `launchTimestamp should not update if already exists`() {
+    val questionnaireResponse = QuestionnaireResponse()
+    val oldDateTimeType = DateTimeType("2024-07-01T00:00:00Z")
+    val newDateTimeType = DateTimeType("2024-07-05T00:00:00Z")
+    questionnaireResponse.launchTimestamp = oldDateTimeType
+    questionnaireResponse.launchTimestamp = newDateTimeType
+
+    val launchTimestamp = questionnaireResponse.launchTimestamp
+    assertThat(launchTimestamp).isNotNull()
+    assertThat(oldDateTimeType).isEqualTo(launchTimestamp)
```
jingtang10, line 729: same here

datacapture/src/main/java/com/google/android/fhir/datacapture/extensions/MoreQuestionnaireResponses.kt
hamza-vd: In my opinion, the `lastLaunched` will make more sense. cc: @qiarie

datacapture/src/test/java/com/google/android/fhir/datacapture/extensions/MoreQuestionnaireResponsesTest.kt
hamza-vd: Updated

datacapture/src/test/java/com/google/android/fhir/datacapture/extensions/MoreQuestionnaireResponsesTest.kt
hamza-vd: Updated

datacapture/src/test/java/com/google/android/fhir/datacapture/extensions/MoreQuestionnaireResponsesTest.kt
hamza-vd: Updated

datacapture/src/main/java/com/google/android/fhir/datacapture/extensions/MoreQuestionnaireResponses.kt
```diff
@@ -154,3 +159,26 @@ private fun unpackRepeatedGroups(
...
+  get() {
+    val extension = this.extension.firstOrNull { it.url == EXTENSION_LAST_LAUNCHED_TIMESTAMP }
+    return extension?.value as? DateTimeType
+  }
+  set(value) {
+    this.extension
+      .find { it.url == EXTENSION_LAST_LAUNCHED_TIMESTAMP }
+      ?.let {
+        // Replace the existing extension with a new one having the updated value
+        this.extension[this.extension.indexOf(it)] =
```
jingtang10, line 177: can we just overwrite the value here?

the whole operation of this `this.extension[this.extension.indexOf(it)]` seems to negate the benefit of using scope functions... looks a bit more clumsy in my opinion.

datacapture/src/test/java/com/google/android/fhir/datacapture/QuestionnaireViewModelParameterizedTest.kt
```diff
@@ -183,8 +184,23 @@ class QuestionnaireViewModelParameterizedTest(
     val printer: IParser = FhirContext.forR4().newJsonParser()
 
     fun <T : IBaseResource> assertResourceEquals(actual: T, expected: T) {
```
jingtang10, line 186: as @ndegwamartin pointed out, might as well split this assert function into two to make it clearer and use the right one in the relevant tests...

i think at the moment this code might hide errors and make debugging difficult in the future.

datacapture/src/test/java/com/google/android/fhir/datacapture/QuestionnaireViewModelTest.kt
```diff
@@ -7514,8 +7515,23 @@ class QuestionnaireViewModelTest {
     val printer: IParser = FhirContext.forR4().newJsonParser()
 
     fun <T : IBaseResource> assertResourceEquals(actual: T, expected: T) {
-      assertThat(printer.encodeResourceToString(actual))
-        .isEqualTo(printer.encodeResourceToString(expected))
+      if (actual is QuestionnaireResponse && expected is QuestionnaireResponse) {
```
jingtang10, line 7518: same here.

jingtang10 (changes requested):
thanks for creating this new pr @hamza-vd

a few small comments

datacapture/src/main/java/com/google/android/fhir/datacapture/extensions/MoreQuestionnaireResponses.kt
hamza-vd: @jingtang10 this is because we need to keep the previous extensions intact and only update the url for a specific extension. The url in extension is a immutable property so I had to assign a new one.

datacapture/src/test/java/com/google/android/fhir/datacapture/QuestionnaireViewModelParameterizedTest.kt
hamza-vd: @ndegwamartin can you elaborate further on how can we split the assertion here?

datacapture/src/main/java/com/google/android/fhir/datacapture/extensions/MoreQuestionnaireResponses.kt
jingtang10: i mean can you just overwrite th value without changeing the url? the url is already the same (as it has been checked in line 174).

so do somethign like this.extension.setValue

datacapture/src/test/java/com/google/android/fhir/datacapture/QuestionnaireViewModelParameterizedTest.kt
jingtang10: create a new function to assert two questionnaire responses in which you fiddle with the timestamps.

this assertion is used by a lot of tests. seems at least inefficient to do this operation for all resource types.

jingtang10 (changes requested):

datacapture/src/main/java/com/google/android/fhir/datacapture/extensions/MoreQuestionnaireResponses.kt
hamza-vd: Updated

datacapture/src/test/java/com/google/android/fhir/datacapture/QuestionnaireViewModelParameterizedTest.kt
hamza-vd: @jingtang10 @ndegwamartin  cc: @qiarie 

The updates I made to this method were not to test my own code, but to ensure the failing tests would pass. You can find the tests for the implementation [here](https://github.com/google/android-fhir/blob/98cb8662a67221abad7e80a2f73d79d6e11faaa8/datacapture/src/test/java/com/google/android/fhir/datacapture/extensions/MoreQuestionnaireResponsesTest.kt#L701).

After adding timestamps to the QuestionnaireResponses, many tests that asserted response equality started failing due to mismatching timestamps. Although the difference was only in milliseconds, it still caused the checks to fail. This left me with two options:

1. Update all the failing tests to use the same timestamp values for both launch and submission.
2. Modify the assertion method to programmatically adjust the expected and actual response timestamps.

I chose the second approach as it was more time-efficient, and we already have separate tests for timestamp functionality. I hope this makes sense.

jingtang10: The problem is that the function is now doing things that is quite specific to questionnaire responses and someone in the future is going to call the `assertResouceEquals` function and assume that it's just testing two resoures are equal but it's not... From a code health's perspective this is an anti pattern... we should make sure that the function does what the name suggests that it does.

Please create a function called `assertQuestionnaireResponseEqualsIgnoringLaunchTimestamp` which takes two `QuestionnaireResponse`s. In the test cases that you are seeing that fail - they should use this new funciton instead of the old function `assertResourceEquals`.

I think this change shouldn't be too big - it should be easy to find and replace usage of this function with questionnaire responses and use the new funciton.

jingtang10 (changes requested):

datacapture/src/test/java/com/google/android/fhir/datacapture/QuestionnaireViewModelParameterizedTest.kt
jingtang10: also make sure you don't mutate the resources - make a copy and then remove the timestamp.

jingtang10 (changes requested):

jingtang10 (approved):
thanks @hamza-vd!

datacapture/src/main/java/com/google/android/fhir/datacapture/extensions/MoreQuestionnaireResponses.kt
jingtang10: is this done?

hamza-vd: Yes it was updated

PR #2687, opened 2024-10-07
Configuration changes to accept UploadStrategy

### LINKED ISSUE #2686: Support uploading via full resource

**Additional context**

MJ1998:

Partially Fixes #2686 

**Description**
This improves configuration format and adds option to upload "updates to a resource" using PUT http method.

**Alternative(s) considered**

**Type**
Choose one: Feature | Code health

**Checklist**

### DIFF
--- a/demo/src/main/java/com/google/android/fhir/demo/data/DemoFhirSyncWorker.kt
+++ b/demo/src/main/java/com/google/android/fhir/demo/data/DemoFhirSyncWorker.kt
@@ -22,7 +22,7 @@ import com.google.android.fhir.demo.FhirApplication
 import com.google.android.fhir.sync.AcceptLocalConflictResolver
 import com.google.android.fhir.sync.DownloadWorkManager
 import com.google.android.fhir.sync.FhirSyncWorker
-import com.google.android.fhir.sync.upload.UpdatePayloadType
+import com.google.android.fhir.sync.upload.HttpUploadMethod
 import com.google.android.fhir.sync.upload.UploadStrategy
 
 class DemoFhirSyncWorker(appContext: Context, workerParams: WorkerParameters) :
@@ -35,7 +35,7 @@ class DemoFhirSyncWorker(appContext: Context, workerParams: WorkerParameters) :
   override fun getConflictResolver() = AcceptLocalConflictResolver
 
   override fun getUploadStrategy(): UploadStrategy =
-    UploadStrategy.AllChangesSquashedBundlePut(UpdatePayloadType.PATCH, 500)
+    UploadStrategy.forBundleRequest(HttpUploadMethod.PUT, HttpUploadMethod.PATCH, true, 500)
 
   override fun getFhirEngine() = FhirApplication.fhirEngine(applicationContext)
 }
--- a/demo/src/main/java/com/google/android/fhir/demo/data/DemoFhirSyncWorker.kt
+++ b/demo/src/main/java/com/google/android/fhir/demo/data/DemoFhirSyncWorker.kt
@@ -22,6 +22,7 @@ import com.google.android.fhir.demo.FhirApplication
 import com.google.android.fhir.sync.AcceptLocalConflictResolver
 import com.google.android.fhir.sync.DownloadWorkManager
 import com.google.android.fhir.sync.FhirSyncWorker
+import com.google.android.fhir.sync.upload.UpdatePayloadType
 import com.google.android.fhir.sync.upload.UploadStrategy
 
 class DemoFhirSyncWorker(appContext: Context, workerParams: WorkerParameters) :
@@ -33,7 +34,8 @@ class DemoFhirSyncWorker(appContext: Context, workerParams: WorkerParameters) :
 
   override fun getConflictResolver() = AcceptLocalConflictResolver
 
-  override fun getUploadStrategy(): UploadStrategy = UploadStrategy.AllChangesSquashedBundlePut
+  override fun getUploadStrategy(): UploadStrategy =
+    UploadStrategy.AllChangesSquashedBundlePut(UpdatePayloadType.PATCH, 500)
 
   override fun getFhirEngine() = FhirApplication.fhirEngine(applicationContext)
 }
--- a/demo/src/main/java/com/google/android/fhir/demo/data/DemoFhirSyncWorker.kt
+++ b/demo/src/main/java/com/google/android/fhir/demo/data/DemoFhirSyncWorker.kt
@@ -36,7 +36,7 @@ class DemoFhirSyncWorker(appContext: Context, workerParams: WorkerParameters) :
   override fun getConflictResolver() = AcceptLocalConflictResolver
 
   override fun getUploadStrategy(): UploadStrategy =
-    UploadStrategy.forBundleRequest(HttpCreateMethod.PUT, HttpUpdateMethod.PATCH, true, 500)
+    UploadStrategy.forBundleRequest(methodForCreate = HttpCreateMethod.PUT, methodForUpdate = HttpUpdateMethod.PATCH, squash = true, bundleSize = 500)
 
   override fun getFhirEngine() = FhirApplication.fhirEngine(applicationContext)
 }
--- a/demo/src/main/java/com/google/android/fhir/demo/data/DemoFhirSyncWorker.kt
+++ b/demo/src/main/java/com/google/android/fhir/demo/data/DemoFhirSyncWorker.kt
@@ -40,7 +40,7 @@ class DemoFhirSyncWorker(appContext: Context, workerParams: WorkerParameters) :
       methodForCreate = HttpCreateMethod.PUT,
       methodForUpdate = HttpUpdateMethod.PATCH,
       squash = true,
-      bundleSize = 500
+      bundleSize = 500,
     )
 
   override fun getFhirEngine() = FhirApplication.fhirEngine(applicationContext)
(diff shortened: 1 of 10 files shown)


engine/src/main/java/com/google/android/fhir/sync/upload/UploadStrategy.kt
jingtang10, line 1: can we just remove some of the strategies that we are not going to implement - especially given that we're using this approach where the verbs are being passed to the strategy.

jingtang10 (approved):

engine/src/main/java/com/google/android/fhir/sync/upload/UploadStrategy.kt
MJ1998: Would they be - SingleChange and SingleResource ones ?
Because we could encounter cycles and hence syncing in Bundle is the only choice. 
This would mean removing SingleResourcePost @santosh-pingle completed recently.

MJ1998: Following would be the valid strategies (assuming bundle is the only way and we are introducing PUT for updates):-

1. AllChangesSquashedPutForCreatePatchForUpdate
2. AllChangesSquashedPutForCreateAndUpdate
3. AllChangesSquashedPostForCreateAndPatchForUpdate
4. AllChangesSquashedPostForCreateAndPutForUpdate
5. AllChangesPutForCreatePatchForUpdate
6. AllChangesPutForCreateAndUpdate
7. AllChangesPostForCreateAndPatchForUpdate
8. AllChangesPostForCreateAndPutForUpdate

And like @aditya-07 said, 
Lets provide a generic data class (UploadStrategy) with 3 variables (each having 2 possibilities) :-
1. squash: [True, False]
2. methodForCreate: [PUT, POST]
3. methodForUpdate: [PATCH, PUT]

This makes it 8 options. 
And additionally we should add bundleSize to this class as bundles are the only way to go.

Is this what you meant, @aditya-07  ?

aditya-07: Yes, but we would want to know if the upload should be individual or Bundle.
In that case, we could add an additional param or have two distinct `UploadStrategy` one for each packaging style e.g. `IndividualUploadStrategy` and `BundleUploadStrategy` that may take appropriate params.

MJ1998: From discussion with Aditya offline, we came up with the current solution of having a factory to create UploadStrategy objects. 
Also, we still need to support uploading by single resource or single change as some servers might not support bundle uploading (they would probably turn off referential integrity when syncing single changes or single resources).

aditya-07 (changes requested):

demo/src/main/java/com/google/android/fhir/demo/data/DemoFhirSyncWorker.kt
```diff
@@ -33,7 +34,8 @@ class DemoFhirSyncWorker(appContext: Context, workerParams: WorkerParameters) :
 
   override fun getConflictResolver() = AcceptLocalConflictResolver
 
-  override fun getUploadStrategy(): UploadStrategy = UploadStrategy.AllChangesSquashedBundlePut
+  override fun getUploadStrategy(): UploadStrategy =
+    UploadStrategy.forBundleRequest(HttpUploadMethod.PUT, HttpUploadMethod.PATCH, true, 500)
```
jingtang10, line 38: use named function params here - hard to understand what these are otherwise.

engine/src/androidTest/java/com/google/android/fhir/db/impl/DatabaseImplTest.kt
```diff
@@ -559,7 +560,9 @@ class DatabaseImplTest {
     // Delete the patient created in setup as we only want to upload the patient in this test
     database.deleteUpdates(listOf(TEST_PATIENT_1))
     services.fhirEngine
-      .syncUpload(AllChangesSquashedBundlePut) { lcs, _ ->
+      .syncUpload(
+        UploadStrategy.forBundleRequest(HttpUploadMethod.PUT, HttpUploadMethod.PATCH, true, 500),
```
jingtang10, line 564: same here

engine/src/main/java/com/google/android/fhir/sync/upload/UploadStrategy.kt
```diff
@@ -50,94 +51,127 @@ import org.hl7.fhir.r4.model.codesystems.HttpVerb
  * fetching, patch generation, and upload request creation. Not all possible combinations of these
  * modes are valid or supported.
  */
-sealed class UploadStrategy
+open class UploadStrategy
```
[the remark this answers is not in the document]
MJ1998: Done

engine/src/main/java/com/google/android/fhir/sync/upload/UploadStrategy.kt
```diff
@@ -50,94 +51,127 @@ import org.hl7.fhir.r4.model.codesystems.HttpVerb
...
+     * @param methodForCreate The HTTP method to use for creating new resources (PUT or POST).
+     * @param methodForUpdate The HTTP method to use for updating existing resources (PUT or PATCH).
+     * @param squash Whether to combine multiple changes to the same resource into a single update.
+     *   Only `true` is supported currently.
+     * @param bundleSize The maximum number of resources to include in a single bundle.
+     * @return An [UploadStrategy] configured for bundle requests.
+     */
+    fun forBundleRequest(
+      methodForCreate: HttpUploadMethod,
+      methodForUpdate: HttpUploadMethod,
```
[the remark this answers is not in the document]
MJ1998: done.

engine/src/androidTest/java/com/google/android/fhir/db/impl/DatabaseImplTest.kt
MJ1998: done.

engine/src/main/java/com/google/android/fhir/sync/upload/UploadStrategy.kt
```diff
@@ -50,94 +51,127 @@ import org.hl7.fhir.r4.model.codesystems.HttpVerb
...
+     * @param methodForCreate The HTTP method to use for creating new resources (PUT or POST).
+     * @param methodForUpdate The HTTP method to use for updating existing resources (PUT or PATCH).
+     * @param squash Whether to squash multiple changes to the same resource into a single update.
+     *   If `true`, all changes for a resource are fetched and patches are generated per resource.
+     *   If `false`, only the earliest change is fetched and patches are generated per change.
+     * @return An [UploadStrategy] configured for individual requests.
+     */
+    fun forIndividualRequest(
+      methodForCreate: HttpUploadMethod,
+      methodForUpdate: HttpUploadMethod,
```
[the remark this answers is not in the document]
MJ1998: Done.

engine/src/main/java/com/google/android/fhir/sync/upload/UploadStrategy.kt
```diff
@@ -50,94 +51,130 @@ import org.hl7.fhir.r4.model.codesystems.HttpVerb
...
+     * @return An [UploadStrategy] configured for individual requests.
+     */
+    fun forIndividualRequest(
+      methodForCreate: HttpCreateMethod,
+      methodForUpdate: HttpUpdateMethod,
+      squash: Boolean,
+    ): UploadStrategy {
+      require(methodForUpdate != HttpUpdateMethod.PUT || squash) {
+        "Http method PUT not supported for UPDATE with squash set as false."
+      }
```
aditya-07, line 127: Since PUT is not supported, is this required.

engine/src/main/java/com/google/android/fhir/sync/upload/UploadStrategy.kt
```diff
@@ -50,94 +51,130 @@ import org.hl7.fhir.r4.model.codesystems.HttpVerb
...
+      if (methodForUpdate == HttpUpdateMethod.PUT) {
+        throw NotImplementedError("PUT for UPDATE not supported yet.")
+      }
```
aditya-07, line 130: Move this to first incase you wanna keep the following block for future.
 ``` 
require(methodForUpdate != HttpUpdateMethod.PUT || squash) {
        "Http method PUT not supported for UPDATE with squash set as false."
      }
```

aditya-07 (approved):
Nits, otherwise LGTM!

engine/src/main/java/com/google/android/fhir/sync/upload/UploadStrategy.kt
MJ1998: When we implement PUT for update I'll simply remove this if block in both the request types. 
The require statements will still be valid, correct ?

aditya-07: Correct. For now, I would like devs to get `NotImplementedError` first if they use PUT rather than the  `IllegalArgumentException` by `require`.

MJ1998: Makes sense!

MJ1998: done.

engine/src/main/java/com/google/android/fhir/sync/upload/UploadStrategy.kt
MJ1998: Still kept in code. Just reordered the error checking for dev convenience. Also its ok to keep it. When implementing PUT for updates we can remove the other if check and this would still hold true.

PR #2717, opened 2024-11-11
api support for sync job cancellation. 

### LINKED ISSUE #2695: Add API to cancel sync jobs
The sync jobs are identified by the fhir sync worker class at the moment. but maybe it would be a good idea to create an id that can be used to cancelling the job.

add a new id to identify jobs - so that we can have unique jobs even if they reuse the same fhir sync worker. and that id can be used to cancel jobs.

NA

**Additional context**
NA

@santosh-pingle

santosh-pingle:

Fixes #2695 

**Description**

**Alternative(s) considered**

**Type**
Choose one: Feature

**Checklist**

### DIFF
--- a/demo/src/main/java/com/google/android/fhir/demo/SyncFragment.kt
+++ b/demo/src/main/java/com/google/android/fhir/demo/SyncFragment.kt
@@ -30,6 +30,7 @@ import androidx.fragment.app.viewModels
 import androidx.navigation.fragment.NavHostFragment
 import com.google.android.fhir.demo.extensions.launchAndRepeatStarted
 import com.google.android.fhir.sync.CurrentSyncJobStatus
+import timber.log.Timber
 
 class SyncFragment : Fragment() {
   private val syncFragmentViewModel: SyncFragmentViewModel by viewModels()
@@ -49,9 +50,12 @@ class SyncFragment : Fragment() {
     view.findViewById<Button>(R.id.sync_now_button).setOnClickListener {
       syncFragmentViewModel.triggerOneTimeSync()
     }
+    view.findViewById<Button>(R.id.cancel_sync_button).setOnClickListener {
+      syncFragmentViewModel.cancelOneTimeSyncWork()
+    }
     observeLastSyncTime()
     launchAndRepeatStarted(
-      { syncFragmentViewModel.pollState.collect(::currentSyncJobStatus) },
+      { syncFragmentViewModel.pollState.collect(::updateSyncJobStatus) },
     )
   }
 
@@ -72,25 +76,42 @@ class SyncFragment : Fragment() {
     }
   }
 
-  private fun currentSyncJobStatus(currentSyncJobStatus: CurrentSyncJobStatus) {
-    requireView().findViewById<TextView>(R.id.current_status).text =
+  private fun updateSyncJobStatus(currentSyncJobStatus: CurrentSyncJobStatus) {
+    Timber.d("currentSyncJobStatus: $currentSyncJobStatus")
+    // Update status text
+    val statusTextView = requireView().findViewById<TextView>(R.id.current_status)
+    statusTextView.text =
       getString(R.string.current_status, currentSyncJobStatus::class.java.simpleName)
 
-    // Update progress indicator visibility and handle status-specific actions
+    // Get views once to avoid repeated lookups
     val syncIndicator = requireView().findViewById<ProgressBar>(R.id.sync_indicator)
+    val syncNowButton = requireView().findViewById<Button>(R.id.sync_now_button)
+    val cancelSyncButton = requireView().findViewById<Button>(R.id.cancel_sync_button)
+
+    // Update view states based on sync status
     when (currentSyncJobStatus) {
       is CurrentSyncJobStatus.Running -> {
         syncIndicator.visibility = View.VISIBLE
+        syncNowButton.visibility = View.GONE
+        cancelSyncButton.visibility = View.VISIBLE
       }
       is CurrentSyncJobStatus.Succeeded -> {
         syncIndicator.visibility = View.GONE
         syncFragmentViewModel.updateLastSyncTimestamp(currentSyncJobStatus.timestamp)
+        syncNowButton.visibility = View.VISIBLE
+        cancelSyncButton.visibility = View.GONE
       }
       is CurrentSyncJobStatus.Failed,
+      is CurrentSyncJobStatus.Cancelled, -> {
+        syncIndicator.visibility = View.GONE
+        syncNowButton.visibility = View.VISIBLE
+        cancelSyncButton.visibility = View.GONE
+      }
       is CurrentSyncJobStatus.Enqueued,
-      is CurrentSyncJobStatus.Cancelled,
       is CurrentSyncJobStatus.Blocked, -> {
         syncIndicator.visibility = View.GONE
+        syncNowButton.visibility = View.GONE
+        cancelSyncButton.visibility = View.VISIBLE
       }
     }
   }
--- a/engine/src/main/java/com/google/android/fhir/sync/Sync.kt
+++ b/engine/src/main/java/com/google/android/fhir/sync/Sync.kt
@@ -42,12 +42,14 @@ import com.google.android.fhir.sync.CurrentSyncJobStatus.Succeeded
 import com.google.gson.Gson
 import com.google.gson.GsonBuilder
 import java.time.OffsetDateTime
+import java.util.*
 import kotlinx.coroutines.ExperimentalCoroutinesApi
 import kotlinx.coroutines.flow.Flow
 import kotlinx.coroutines.flow.asFlow
 import kotlinx.coroutines.flow.combine
 import kotlinx.coroutines.flow.flatMapConcat
 import kotlinx.coroutines.flow.mapNotNull
+import timber.log.Timber
 
 object Sync {
   val gson: Gson =
@@ -67,7 +69,7 @@ object Sync {
   inline fun <reified W : FhirSyncWorker> oneTimeSync(
     context: Context,
     retryConfiguration: RetryConfiguration? = defaultRetryConfiguration,
-  ): Flow<CurrentSyncJobStatus> {
+  ): Pair<Flow<CurrentSyncJobStatus>, UUID> {
     val uniqueWorkName = "${W::class.java.name}-oneTimeSync"
     val flow = getWorkerInfo(context, uniqueWorkName)
     val oneTimeWorkRequest =
@@ -78,7 +80,8 @@ object Sync {
         ExistingWorkPolicy.KEEP,
         oneTimeWorkRequest,
       )
-    return combineSyncStateForOneTimeSync(context, uniqueWorkName, flow)
+    val workId = oneTimeWorkRequest.id
+    return Pair(combineSyncStateForOneTimeSync(context, uniqueWorkName, flow), workId)
   }
 
   /**
@@ -95,7 +98,7 @@ object Sync {
   inline fun <reified W : FhirSyncWorker> periodicSync(
     context: Context,
     periodicSyncConfiguration: PeriodicSyncConfiguration,
-  ): Flow<PeriodicSyncJobStatus> {
+  ): Pair<Flow<PeriodicSyncJobStatus>, UUID> {
     val uniqueWorkName = "${W::class.java.name}-periodicSync"
     val flow = getWorkerInfo(context, uniqueWorkName)
     val periodicWorkRequest =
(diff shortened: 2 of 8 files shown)


santosh-pingle:
https://github.com/user-attachments/assets/4b795055-2ec5-4f5e-9d1c-6fca3b29fecd


https://github.com/user-attachments/assets/26e67e9b-68ed-4101-a0f3-7630093edbf3

MJ1998 (changes requested):
One way to do this is to accept `ExistingWorkPolicy` in sync apis. Instead of `ExistingWorkPolicy.KEEP` if application provides `ExistingWorkPolicy.REPLACE`, workmanager library will cancel the current one and add new one with the same name.

Why not the above approach ?

jingtang10 (approved):

engine/src/main/java/com/google/android/fhir/sync/FhirDataStore.kt
```diff
@@ -100,6 +101,21 @@ internal class FhirDataStore(context: Context) {
...
+      preferences[stringPreferencesKey("$workIdKey-key")] = uuid.toString()
+    }
+  }
+
+  /** Fetches the stored UUID from DataStore. */
+  @PublishedApi
+  internal suspend fun fetchWorkId(workIdKey: String): UUID? {
+    val preferences = dataStore.data.first()
+    return preferences[stringPreferencesKey("$workIdKey-key")]?.let { UUID.fromString(it) }
+  }
```
MJ1998, line 117: Should there be some locking mechanism here? 
What if both apis are triggered together for the same workIdKey ?

santosh-pingle: Any specific use case which is tightly coupled with these api?

MJ1998: Yeah I mean application could try to sync / un-sync repeatedly which could lead to a race-condition. I think there should be a lock for each workIdKey. @jingtang10 WDYT ?

engine/src/main/java/com/google/android/fhir/sync/Sync.kt
```diff
@@ -79,6 +81,8 @@ object Sync {
         existingWorkPolicy,
         oneTimeWorkRequest,
       )
+    val workId = oneTimeWorkRequest.id
+    FhirEngineProvider.getFhirDataStore(context).storeWorkId(workId, uniqueWorkName)
```
MJ1998, line 85: Why not save against `uniqueWorkName` instead of `workId` ?

MJ1998, line 85: This will get stored even when the sync is running and existingWorkPolicy is set to KEEP (which will not create any new work requests) ?

MJ1998: Instead if we use uniqueWorkName this problem will not arise. 
We can check if the uniqueWorkName already exists in the FhirDataStore, then do nothing.

MJ1998 (changes requested):

engine/src/main/java/com/google/android/fhir/sync/Sync.kt
```diff
@@ -42,12 +42,14 @@ import com.google.android.fhir.sync.CurrentSyncJobStatus.Succeeded
 import com.google.gson.Gson
 import com.google.gson.GsonBuilder
 import java.time.OffsetDateTime
+import java.util.*
```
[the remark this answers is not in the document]
santosh-pingle: done.

engine/src/main/java/com/google/android/fhir/sync/FhirDataStore.kt
santosh-pingle: done.

engine/src/main/java/com/google/android/fhir/sync/Sync.kt
santosh-pingle: done.

santosh-pingle: done.

engine/src/main/java/com/google/android/fhir/sync/FhirDataStore.kt
```diff
@@ -100,6 +102,25 @@ internal class FhirDataStore(context: Context) {
     runBlocking { dataStore.edit { pref -> pref[lastSyncTimestampKey] = datetime.toString() } }
   }
 
+  private val mutex = Mutex()
```
MJ1998, line 105: This should be a mutex map for each key.  Something like following
```suggestion
  private val mutexMap: Map<String, Mutex>
```
Reason: keys (uniqueWorkName) can be different (example - "onetimesync" and "periodicsync") and we dont want to block either of them.

engine/src/main/java/com/google/android/fhir/sync/FhirDataStore.kt
```diff
@@ -100,6 +102,25 @@ internal class FhirDataStore(context: Context) {
...
+
+  /** Fetches the stored unique-work-name from DataStore. */
+  @PublishedApi
+  internal suspend fun fetchUniqueWorkName(key: String): String? {
+    mutex.withLock {
+      val preferences = dataStore.data.first()
+      return preferences[stringPreferencesKey("$key-key")]
+    }
+  }
+
```
MJ1998, line 123: You should have a `cancelUniqueWorkName` api here and use that in the Sync.kt file.

engine/src/main/java/com/google/android/fhir/sync/Sync.kt
```diff
@@ -261,83 +279,67 @@ object Sync {
...
+      WorkInfo.State.ENQUEUED -> {
+        Enqueued
+      }
+      WorkInfo.State.RUNNING -> {
+        return when (syncJobStatusFromWorkManager) {
+          is SyncJobStatus.Started,
+          is SyncJobStatus.InProgress, -> Running(syncJobStatusFromWorkManager)
+          is SyncJobStatus.Succeeded -> Succeeded(syncJobStatusFromWorkManager.timestamp)
+          is SyncJobStatus.Failed -> Failed(syncJobStatusFromWorkManager.timestamp)
+          null -> Running(SyncJobStatus.Started())
```
MJ1998, line 297: Can you leave a description in the PR about these changes ?

MJ1998 (changes requested):

engine/src/main/java/com/google/android/fhir/sync/FhirDataStore.kt
santosh-pingle: If I understand correctly, you are suggesting canceling the job on the WorkManager in the FhirDataStore?`WorkManager.getInstance(context).cancelUniqueWork(uniqueWorkNameValueFromDataStore).`

engine/src/main/java/com/google/android/fhir/sync/Sync.kt
santosh-pingle: done.

engine/src/main/java/com/google/android/fhir/sync/FhirDataStore.kt
santosh-pingle: I'm not sure why we need to cancel the job in the FhirDataStore?

engine/src/main/java/com/google/android/fhir/sync/FhirDataStore.kt
santosh-pingle: done.

engine/src/main/java/com/google/android/fhir/sync/FhirDataStore.kt
MJ1998: We need to remove the keys from datastore. You are right. You may cancel the work in Sync.kt only but you also need to remove the keys from the datastore. You can have an api to remove the keys from datastore

santosh-pingle: done.

engine/src/main/java/com/google/android/fhir/sync/FhirDataStore.kt
```diff
@@ -100,6 +102,41 @@ internal class FhirDataStore(context: Context) {
     runBlocking { dataStore.edit { pref -> pref[lastSyncTimestampKey] = datetime.toString() } }
   }
 
+  private val mutexMap = mutableMapOf<String, Mutex>()
+
+  private fun getMutexForKey(key: String): Mutex {
+    return mutexMap.getOrPut(key) { Mutex() }
+  }
```
MJ1998, line 109: Make this thread safe
```suggestion
  private val mutexMap = ConcurrentHashMap<String, Mutex>()

  private fun getOrCreateMutex(key: String): Mutex {
      // computeIfAbsent is thread-safe and only creates a Mutex if one doesn't exist.
      return mutexMap.computeIfAbsent(key) { Mutex() }
  }
```

demo/src/main/java/com/google/android/fhir/demo/SyncFragment.kt
```diff
@@ -72,25 +76,42 @@ class SyncFragment : Fragment() {
     }
   }
 
-  private fun currentSyncJobStatus(currentSyncJobStatus: CurrentSyncJobStatus) {
-    requireView().findViewById<TextView>(R.id.current_status).text =
+  private fun collectSyncJobStatus(currentSyncJobStatus: CurrentSyncJobStatus) {
```
MJ1998, line 79: Nit: Earlier name of the function was better
```suggestion
  private fun currentSyncJobStatus(currentSyncJobStatus: CurrentSyncJobStatus) {
```

engine/src/main/java/com/google/android/fhir/sync/Sync.kt
```diff
@@ -64,12 +59,12 @@ object Sync {
    * @param retryConfiguration configuration to guide the retry mechanism, or `null` to stop retry.
    * @return a [Flow] of [CurrentSyncJobStatus]
    */
-  inline fun <reified W : FhirSyncWorker> oneTimeSync(
+  suspend inline fun <reified W : FhirSyncWorker> oneTimeSync(
```
MJ1998, line 62: Why suspend ?

engine/src/main/java/com/google/android/fhir/sync/FhirDataStore.kt
```diff
@@ -100,6 +102,41 @@ internal class FhirDataStore(context: Context) {
...
+  @PublishedApi
+  internal suspend fun removeUniqueWorkName(key: String) {
+    getMutexForKey(key).withLock {
+      dataStore.edit { preferences ->
+        val value = preferences.remove(stringPreferencesKey("$key-key"))
+        Timber.d("Removed value: $value")
+      }
+    }
+
+    synchronized(mutexMap) { mutexMap.remove(key) }
```
MJ1998, line 128: This is not needed. The mutex keys will be garbage collected when not being used.

engine/src/main/java/com/google/android/fhir/sync/Sync.kt
santosh-pingle: unique-work-name api in the datastore are marked as suspend.

MJ1998: Oh yes. Why do the apis need to be suspend ?

santosh-pingle: datastore edit and read performs disk io operations, right?

engine/src/main/java/com/google/android/fhir/sync/FhirDataStore.kt
santosh-pingle: done.

demo/src/main/java/com/google/android/fhir/demo/SyncFragment.kt
santosh-pingle: done.

engine/src/main/java/com/google/android/fhir/sync/FhirDataStore.kt
santosh-pingle: done.

santosh-pingle: done.

engine/src/main/java/com/google/android/fhir/sync/Sync.kt
MJ1998: makes sense. resolving.

MJ1998 (approved):
LGTM. Thanks Santosh.

PR #2743, opened 2024-11-27
Overload FhirEngine.get() to Allow Fetching Multiple Resources by their IDs

qiarie:

Fixes #[issue number]

**Description**

**Alternative(s) considered**

**Type**

**Checklist**


engine/src/main/java/com/google/android/fhir/FhirEngine.kt
```diff
@@ -83,6 +83,17 @@ interface FhirEngine {
...
+  /**
+   * Loads multiple FHIR resources given [ResourceType] and logical IDs.
+   *
+   * @param type The type of the resource to load.
+   * @param ids The logical IDs of the resources.
+   * @return The list of requested FHIR resources.
+   * @throws ResourceNotFoundException if the resources are not found.
+   */
+  @Throws(ResourceNotFoundException::class)
+  suspend fun get(type: ResourceType, vararg ids: String): List<Resource>
```
jingtang10, line 95: update the exception to say that if any resources are not found, this will throw.

engine/src/main/java/com/google/android/fhir/FhirEngine.kt
jingtang10, line 1: please add some tests

engine/src/main/java/com/google/android/fhir/FhirEngine.kt
jingtang10, line 95: i don't actually know how this really works with the original api.. it seems to me it'd be ambiguous to the compiler if the first or the second api is called when you're getting only 1 resource... can you verify this?

qiarie: Does this mean if a request is made for 3 resources and only 2 are available it should throw an exception?

qiarie: The alternative would be to add a separate `getResources` API that accepts a resourceType and the list of IDs and returns a list of resources. This would cover the ambiguity.

qiarie: Is there a scenario where getting all resources in the list is mandatory and returning a subset may break functionality?

PR #2751, opened 2024-12-06
populate initial with initialExpression result collection

### LINKED ISSUE #2750: [SDC library] initialExpression with multiple values as result rendering incorrectly.
**Describe the bug**
When the result of initialExpression is a collection instead of a single value, The SDC only renders the first value instead of the whole collection. 
For example, while using it for a multiple choice question, it only populate the first element of resulting collection of the expression.

**Component**
SDC library

**To Reproduce**
- Apply this patch to main branch. [checkbox_not_populated_correctly_.patch](https://github.com/user-attachments/files/18033496/checkbox_not_populated_correctly_.patch)

- Open any patient & select edit patient questionnaire. 
- Observe the selected options.

**Expected behavior**
(Mother , sister & son should be selected).

**Screenshots**
Image

 - Pixel4a emulator

**Additional context**
Here it's only selecting the first value: [https://github.com/google/android-fhir/blob/f42e80427b34dbe9f14ee2f0b4460c38cee0ef[…]a/com/google/android/fhir/datacapture/mapping/ResourceMapper.kt](https://github.com/google/android-fhir/blob/f42e80427b34dbe9f14ee2f0b4460c38cee0ef27/datacapture/src/main/java/com/google/android/fhir/datacapture/mapping/ResourceMapper.kt#L259)

Yes, creating a PR for fix. 

cc: @jingtang10 , @icrc-jofrancisco

parthfloyd:

Fixes #2750 

**Description**
Updated `populateInitialValue()` function in `ResourceMapper.kt` of the SDC module to populate `questionnaireItem.initial` with values of collection instead of just the first value of the collection.

**Alternative(s) considered**
None | Looking forward to feedback

**Type**
Bug fix

**Checklist**


LZRS:
To support multiple initially selected options from the dropdown, not quite sure populating the `Questionnaire.item.initial` would work. 
From this [`Questionnaire.item.initial` invariant](https://build.fhir.org/questionnaire-definitions.html#Questionnaire.item.initial)

> que-11	[Rule](https://build.fhir.org/conformance-rules.html#rule)	If one or more answerOption is present, initial cannot be present. Use answerOption.initialSelected instead	answerOption.empty() or initial.empty()

I would instead suggest using [`Questionnaire.item.answerOption.initialSelected`](https://build.fhir.org/questionnaire-definitions.html#Questionnaire.item.answerOption.initialSelected)

parthfloyd:
@LZRS Thank you so much for your feedback! 
I've pushed code & now it's conforming with the  [invariant ](https://build.fhir.org/questionnaire-definitions.html#Questionnaire.item.initial)rules. Please feel free if any changes or optimizations are required in the code.

Invariant:
> que-8	Rule	Initial values can't be specified for groups or display items

> que-13	Rule	Can only have multiple initial values for repeating items

> que-11	Rule	If one or more answerOption is present, initial cannot be present. Use answerOption.initialSelected instead

cc: @jingtang10 , @icrc-jofrancisco

datacapture/src/main/java/com/google/android/fhir/datacapture/mapping/ResourceMapper.kt
```diff
@@ -248,22 +249,51 @@ object ResourceMapper {
...
-          mutableListOf(Questionnaire.QuestionnaireItemInitialComponent().setValue(value))
+        // Set initial value for the questionnaire item.
+        if (it.isEmpty()) return@let
+
+        val evaluatedExpressionResult = it.map { it.asExpectedType(questionnaireItem.type) }
+
+        if (questionnaireItem.answerOption.isNotEmpty()) {
+          questionnaireItem.answerOption.forEach { answerOption ->
+            answerOption.initialSelected =
+              evaluatedExpressionResult.any { answerOption.value.equalsDeep(it) }
```
parthfloyd, line 280: For questionnaireItems with answerOptions, it is checking using equalsDeep to match answerOptions value with the result of initialExpression. 

Please let me know if this should be updated,
I had to update two test cases as the initialExpression result was not matching exactly with the answerOptions (system was missing in answerOption)

Reference:

> If the expression returns elements of type 'code', when the user selects one or more candidates, the code will be matched against the allowed Codings (from answerValueSet, answerOptions, or answerExpresssion).

- https://build.fhir.org/ig/HL7/sdc/StructureDefinition-sdc-questionnaire-initialExpression-definitions.html

> If two codings have the same system, version and code then they have the same meaning. If the version information is missing, or the system, version or the code elements differ, then how the codes are related can only be determined by consulting the definitions of the system(s) and any mappings available.

- https://hl7.org/fhir/R4/datatypes.html#codesystem

LZRS (approved):
LGTM

datacapture/src/main/java/com/google/android/fhir/datacapture/mapping/ResourceMapper.kt
parthfloyd: As discussed in developer call:
- added test for selecting answerOption with value `Coding` without `Coding.system` (in initialExpression result & answerOption).

To sum it up:
(In cases where result of initialExpression & type of answer is Coding) it will only select answerOption if the Coding matches exactly.

datacapture/src/main/java/com/google/android/fhir/datacapture/mapping/ResourceMapper.kt
```diff
@@ -248,22 +249,51 @@ object ResourceMapper {
...
+          }
+        } else {
+          // Handle initial values based on whether the questionnaire item repeats
```
jingtang10, line 283: so in this `else` block, the `anwerOption` is empty in the questionnaire item? why do we have to do anything in this case?

jingtang10: pls also reflect these changes with new test cases.

jingtang10 (changes requested):

datacapture/src/main/java/com/google/android/fhir/datacapture/mapping/ResourceMapper.kt
parthfloyd: Thanks for the review @jingtang10 Best wishes for 2025!! I'll be adding test cases for it.

Explanation for code:
- So line 277-281 handles the case when questionnaireItem has an answerOption (then selects initialSelected) [que-11]
- line 284-291 handles the case when it doesn't have answerOption but it is a repeating item (only repeating item are allowed multiple initial values. [que-13]
- lastly 292-295 handles the default case when a regular questionnaireItem without answerOption has 1 initial value. 

Reference:

> Invariant:

> que-8 Rule Initial values can't be specified for groups or display items
> 
> que-13 Rule Can only have multiple initial values for repeating items
> 
> que-11 Rule If one or more answerOption is present, initial cannot be present. Use answerOption.initialSelected instead

datacapture/src/main/java/com/google/android/fhir/datacapture/mapping/ResourceMapper.kt
```diff
@@ -248,22 +249,51 @@ object ResourceMapper {
...
-        // initial value and initial expression.
-        val value = it.asExpectedType(questionnaireItem.type)
-        questionnaireItem.initial =
-          mutableListOf(Questionnaire.QuestionnaireItemInitialComponent().setValue(value))
+        // Set initial value for the questionnaire item.
+        if (it.isEmpty()) return@let
+
+        val evaluatedExpressionResult = it.map { it.asExpectedType(questionnaireItem.type) }
+
+        if (questionnaireItem.answerOption.isNotEmpty()) {
```
jingtang10, line 277: add a comment here to say for answer options, the initialSelected extension is used to highlight initial values.

but another thing to add here, is that if the initial expression evaluates to 1,2,3,4,5, but you only have 3 answer options 1,2,3... then 4 and 5 will be ignored here... we don't add them as additional values (nor does it make sense to do so). but it's just worth commenting that.

parthfloyd: Thanks @jingtang10  I've added the comment & test-cases! Please feel free to let me know if any changes are required!

datacapture/src/test/java/com/google/android/fhir/datacapture/mapping/ResourceMapperTest.kt
```diff
@@ -2973,6 +2979,414 @@ class ResourceMapperTest {
       )
   }
 
+  @Test
+  fun `populate() should fail with IllegalStateException when QuestionnaireItem of group or display item has a initial value`():
```
jingtang10, line 2983: can you split this into 2 smaller test cases one for group one for display item?

In each questionnaire just include a single questionnaire item.

In general, for tests, it's much better to keep tests small and focused. So when a test fails from the test name you'll immediately figure out what's the problem, instead of having to debug.

datacapture/src/test/java/com/google/android/fhir/datacapture/mapping/ResourceMapperTest.kt
```diff
@@ -2973,6 +2979,414 @@ class ResourceMapperTest {
...
+      val observationBundle =
+        Bundle().apply {
+          addEntry().resource = observation1
+          addEntry().resource = observation2
+        }
+
+      val questionnaireResponse =
+        ResourceMapper.populate(questionnaire, mapOf("observations" to observationBundle))
+
+      assertThat((questionnaireResponse.item[0].answer[0].value as Coding).code)
```
jingtang10, line 3228: extract questionnaireResponse.item[0].answer[0].value as a variable. or use kotlin's `with`:

```
with(questionnaireResponse.item[0].answer[0].value) {
  assert(this.code...)
  assert(this.system...)
}
```

datacapture/src/test/java/com/google/android/fhir/datacapture/mapping/ResourceMapperTest.kt
```diff
@@ -2973,6 +2979,414 @@ class ResourceMapperTest {
...
+            CodeableConcept().apply {
+              coding =
+                mutableListOf(
+                  Coding().apply {
+                    code = "correct-code-val"
+                    display = "correct-display-val"
+                  },
+                )
+            }
+        }
```
jingtang10, line 3218: what's the significance of having 2 observations in this test case? it seems that's not really being tested?

is the point you would collapse the 2 choices? would it be better to give the second observation a different coding and show that the first one is chosen?

datacapture/src/test/java/com/google/android/fhir/datacapture/mapping/ResourceMapperTest.kt
```diff
@@ -2973,6 +2979,414 @@ class ResourceMapperTest {
...
+        ResourceMapper.populate(questionnaire, mapOf("observations" to observationBundle))
+
+      assertThat((questionnaireResponse.item[0].answer[0].value as Coding).code)
+        .isEqualTo("correct-code-val")
+      assertThat((questionnaireResponse.item[0].answer[0].value as Coding).display)
+        .isEqualTo("correct-display-val")
+      assertThat((questionnaireResponse.item[0].answer[1].value as Coding).code)
+        .isEqualTo("correct2-code-val")
+      assertThat((questionnaireResponse.item[0].answer[1].value as Coding).display)
+        .isEqualTo("correct2-display-val")
```
jingtang10, line 3333: similarly extract local variables instead of repeating

datacapture/src/test/java/com/google/android/fhir/datacapture/mapping/ResourceMapperTest.kt
```diff
@@ -2973,6 +2979,414 @@ class ResourceMapperTest {
...
+    }
+
+  @Test
+  fun `populate() should select a single initial answer for non repeating question without answerOption`() =
```
jingtang10, line 3337: i think a missing test case here is that if there's no answer option but the expression evaluates to multiple values?

you should test in that case only the first value is used. right?

jingtang10 (commented):
Very good test cases! Thanks @parthfloyd!

A few comments below

datacapture/src/test/java/com/google/android/fhir/datacapture/mapping/ResourceMapperTest.kt
parthfloyd: Thanks @jingtang10 This was a ctrl-c/v error! Thanks for finding it! Fixed accordingly!

datacapture/src/test/java/com/google/android/fhir/datacapture/mapping/ResourceMapperTest.kt
parthfloyd: Yes! added test case for it!

MJ1998 (approved):
LGTM

MJ1998:
Please resolve the NIT and update your branch. Thanks for the PR @parthfloyd

parthfloyd:
@MJ1998 Thanks for the review! I've pushed changes accordingly. Looking forward to the merge & further development on populate!

jingtang10 (approved):

PR #2763, opened 2025-01-01
Include min and max values in the validation message in a number question if the values are in the extension

### LINKED ISSUE #2757: Number question validation message does not consider min max value given in the question
**Describe the bug**
The validation message for a number field displays "The answer should be a number between <INT_MIN> and <INT_MAX>" when input a non-numeric character. This should rather be the min and max value (if) given in the questionnaire item.

**Component**

**To Reproduce**

**Expected behavior**
The validation message should be, for example, "The answer should be a number between 30 and 200."

**Screenshots**

**Additional context**

chungwwei:

Fixes #2757

**Description**
Show  `number must be between <INT_MIN> and <INT_MAX>` when input is not an integer.
Use the max and min values in the extension if they are defined. 

**Alternative(s) considered**

**Type**
Choose one: Bug fix

**Checklist**

### DIFF
--- a/datacapture/src/main/java/com/google/android/fhir/datacapture/views/factories/EditTextIntegerViewHolderFactory.kt
+++ b/datacapture/src/main/java/com/google/android/fhir/datacapture/views/factories/EditTextIntegerViewHolderFactory.kt
@@ -24,6 +24,8 @@ import android.text.InputType
 import androidx.annotation.RequiresApi
 import com.google.android.fhir.datacapture.R
 import com.google.android.fhir.datacapture.extensions.getValidationErrorMessage
+import com.google.android.fhir.datacapture.validation.MAX_VALUE_EXTENSION_URL
+import com.google.android.fhir.datacapture.validation.MIN_VALUE_EXTENSION_URL
 import com.google.android.fhir.datacapture.views.QuestionnaireViewItem
 import com.google.android.material.textfield.TextInputEditText
 import com.google.android.material.textfield.TextInputLayout
@@ -38,6 +40,22 @@ internal object EditTextIntegerViewHolderFactory :
       QuestionnaireItemEditTextViewHolderDelegate(
         InputType.TYPE_CLASS_NUMBER or InputType.TYPE_NUMBER_FLAG_SIGNED,
       ) {
+
+      private val minValue: Int by lazy {
+        getExtensionValueOrDefault(MIN_VALUE_EXTENSION_URL, Int.MIN_VALUE)
+      }
+
+      private val maxValue: Int by lazy {
+        getExtensionValueOrDefault(MAX_VALUE_EXTENSION_URL, Int.MAX_VALUE)
+      }
+
+      private fun getExtensionValueOrDefault(url: String, defaultValue: Int): Int {
+        return questionnaireViewItem.questionnaireItem.extension
+          .find { it.url == url }
+          ?.let { (it.value as? IntegerType)?.value }
+          ?: defaultValue
+      }
+
       override suspend fun handleInput(
         editable: Editable,
         questionnaireViewItem: QuestionnaireViewItem,
@@ -95,8 +113,8 @@ internal object EditTextIntegerViewHolderFactory :
           textInputLayout.error =
             textInputLayout.context.getString(
               R.string.integer_format_validation_error_msg,
-              formatInteger(Int.MIN_VALUE),
-              formatInteger(Int.MAX_VALUE),
+              formatInteger(minValue),
+              formatInteger(maxValue),
             )
         }
       }


datacapture/src/main/java/com/google/android/fhir/datacapture/views/factories/EditTextIntegerViewHolderFactory.kt
```diff
@@ -38,6 +40,22 @@ internal object EditTextIntegerViewHolderFactory :
...
+        getExtensionValueOrDefault(MAX_VALUE_EXTENSION_URL, Int.MAX_VALUE)
+      }
+
+      private fun getExtensionValueOrDefault(url: String, defaultValue: Int): Int {
+        return questionnaireViewItem.questionnaireItem.extension
+          .find { it.url == url }
+          ?.let { (it.value as? IntegerType)?.value }
+          ?: defaultValue
+      }
+
```
FikriMilano, line 58: feels like we can move this code to `MoreQuestionnaireItemComponents.kt` and probably incorporate the new code to some existing ones

FikriMilano (changes requested):
Thanks for the PR!

datacapture/src/main/java/com/google/android/fhir/datacapture/views/factories/EditTextIntegerViewHolderFactory.kt
chungwwei: 😅 I updated the code to use the extension values directly

https://github.com/google/android-fhir/blob/f343e26cc4d8625aeb7f328338bfb8b69c0947d3/datacapture/src/main/java/com/google/android/fhir/datacapture/extensions/MoreQuestionnaireItemComponents.kt#L336-L348

```
val minValue = (questionnaireViewItem.minAnswerValue as? IntegerType)?.value ?: Int.MIN_VALUE
val maxValue = (questionnaireViewItem.maxAnswerValue as? IntegerType)?.value ?: Int.MAX_VALUE
```

datacapture/src/main/java/com/google/android/fhir/datacapture/views/factories/EditTextIntegerViewHolderFactory.kt
```diff
@@ -90,13 +91,19 @@ internal object EditTextIntegerViewHolderFactory :
             questionnaireViewItem,
             questionnaireViewItem.validationResult,
           )
+
+        val minValue =
+          (questionnaireViewItem.minAnswerValue as? IntegerType)?.value ?: Int.MIN_VALUE
+        val maxValue =
+          (questionnaireViewItem.maxAnswerValue as? IntegerType)?.value ?: Int.MAX_VALUE
+
```
FikriMilano, line 99: what happens when the minValue is bigger than the maxValue? 
how about we throw an exception when that happens?

jingtang10 (approved):
Thanks for adding this. I'm ok with this PR.

Fikri's point is a good one - we do throw exceptions in other widgets when min value > max value. But I'm a bit concerned about the UI crasing... I wonder if we actually should remove that exception from the date and slider widgets.

But I'm open-minded in this case. I'll approve this PR - but pls resolve the exception question with fikri - and we can then merge this pr.

Thanks again for this - good work!

datacapture/src/main/java/com/google/android/fhir/datacapture/views/factories/EditTextIntegerViewHolderFactory.kt
chungwwei: I added the exception. UI will crash when minValue > maxValue, like the slider and date widgets.

FikriMilano: Okay, but feel free to speak your thoughts on  this, we could for example: show an error message to the user, instead of outright crashing the app. We're open to change.

What do you think?

datacapture/src/main/java/com/google/android/fhir/datacapture/views/factories/EditTextIntegerViewHolderFactory.kt
```diff
@@ -38,6 +38,18 @@ internal object EditTextIntegerViewHolderFactory :
...
+        super.bind(questionnaireViewItem)
+
+        val minValue = (questionnaireViewItem.minAnswerValue as? IntegerType)?.value
+        val maxValue = (questionnaireViewItem.maxAnswerValue as? IntegerType)?.value
+
+        if (minValue != null && maxValue != null && minValue > maxValue) {
+          throw IllegalArgumentException("minValue cannot be greater than maxValue")
+        }
+      }
+
```
FikriMilano, line 52: I feel overriding the `bind` function is not necessary, `updateValidationTextUI` function is already called in the parent's `bind` (you can check it), so throwing the error in `updateValidationTextUI` should suffice.

What do you think?

FikriMilano (changes requested):

jingtang10:
if you have the max and min values on a different question type, e.g. slider? or date / date time?

if you put this logic in questionnaire view model this would be done for all widgets - instead of specific ui widgets.

datacapture/src/main/java/com/google/android/fhir/datacapture/views/factories/EditTextIntegerViewHolderFactory.kt
chungwwei: Ig it's rare for a scenario like that to happen. Adheres to the rules, but with nonsensical values.

Alternatively, I view the minValue > maxValue a meaningless constraint on the UI, so it make sense to treat it as though the constraint doesn’t exist? We can set `minAnswerValue` and `maxAnswerValue` to `null` if when that happens.

Looks like the web app https://lhcforms.nlm.nih.gov/lhcforms ignored that scenario

datacapture/src/main/java/com/google/android/fhir/datacapture/views/factories/EditTextIntegerViewHolderFactory.kt
chungwwei: I had it that way initially,  but i thought placing the constraint check didn’t fit the intent of the function name `updateValidationTextUI`, so I moved it to `bind`. Like jingtang10 commented, it would be better to have it in the view model to remove redundancy

chungwwei:
hey @jingtang10, thanks for the pointer! I made some changes to the code based on your last comment. could you please take a look and let me know what you think

PR #2764, opened 2025-01-03
Add custom callback for autocomplete

### LINKED ISSUE #2738: Enhance the ExternalAnswerValueSetResolver to take a search string
The [ExternalAnswerValueSetResolver](https://github.com/google/android-fhir/blob/507e89cc8b75c7b6667c8bff47b914c3ac94a770/datacapture/src/main/java/com/google/android/fhir/datacapture/DataCaptureConfig.kt#L82) takes a URI of the value set and resolves that to a list of codings. The implementation is provided by client application (it can load it from disk, retrieve it from a terminology server, or from hardcoded values) and this is done ONLY ONCE when the question is rendered.

However, in reality, often the valueset is too large. So a more smart version of this operation is needed if we want to integrate with a terminology server. For example, the user might type "dia" and we want to send "dia" as well as the valueset URL to the terminology server and retrieve a smaller subset of codes in the valueset to make the user experience better. 

We could either create a new interface, or a new API in ExternalAnswerValueSetResolver that takes not only the url but also a search string.

We should also provide a default implementation that client application can use to connect to any terminology server.

NA

**Additional context**
@fredhersch @dilys-l @johngrimes

@dilys-l ?

dilys-l:

Fixes #2738 

**Description**
Adds the ability for the client to pass a custom callback to AutoComplete for dynamic AutoComplete answer resolution

**Alternative(s) considered**
NA

**Type**
Feature

**Checklist**

### DIFF
--- a/catalog/src/main/java/com/google/android/fhir/catalog/CatalogApplication.kt
+++ b/catalog/src/main/java/com/google/android/fhir/catalog/CatalogApplication.kt
@@ -21,7 +21,10 @@ import ca.uhn.fhir.context.FhirContext
 import com.google.android.fhir.FhirEngine
 import com.google.android.fhir.FhirEngineConfiguration
 import com.google.android.fhir.FhirEngineProvider
+import com.google.android.fhir.datacapture.CustomCallback.AutoCompleteCallback
+import com.google.android.fhir.datacapture.CustomCallbackType
 import com.google.android.fhir.datacapture.DataCaptureConfig
+import com.google.android.fhir.datacapture.views.factories.AutoCompleteViewAnswerOption
 import com.google.android.fhir.search.search
 import kotlinx.coroutines.CoroutineScope
 import kotlinx.coroutines.Dispatchers
@@ -44,6 +47,15 @@ class CatalogApplication : Application(), DataCaptureConfig.Provider {
         xFhirQueryResolver = { fhirEngine.search(it).map { it.resource } },
         questionnaireItemViewHolderFactoryMatchersProviderFactory =
           ContribQuestionnaireItemViewHolderFactoryMatchersProviderFactory,
+        callbacks = mapOf(Pair(CustomCallbackType.AUTO_COMPLETE, AutoCompleteCallback(
+          callback = { query ->
+            run {
+              listOf(AutoCompleteViewAnswerOption("a", "Type 2 Diabetes Mellitus"),
+                AutoCompleteViewAnswerOption("b", "Test")
+                )
+            }
+          }
+        )))
       )
 
     CoroutineScope(Dispatchers.IO).launch {
--- a/datacapture/src/main/java/com/google/android/fhir/datacapture/DataCaptureConfig.kt
+++ b/datacapture/src/main/java/com/google/android/fhir/datacapture/DataCaptureConfig.kt
@@ -1,5 +1,5 @@
 /*
- * Copyright 2022-2024 Google LLC
+ * Copyright 2022-2025 Google LLC
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -56,6 +56,12 @@ data class DataCaptureConfig(
   var questionnaireItemViewHolderFactoryMatchersProviderFactory:
     QuestionnaireItemViewHolderFactoryMatchersProviderFactory? =
     null,
+
+  /**
+   * A [CustomCallback] may be set by the client to override the behaviour of an existing component
+   * in the sdc. Currently only supports [CustomCallbackType.AUTO_COMPLETE].
+   */
+  var callbacks: Map<CustomCallbackType, CustomCallback>? = null,
 ) {
 
   /**
--- a/datacapture/src/main/java/com/google/android/fhir/datacapture/QuestionnaireEditAdapter.kt
+++ b/datacapture/src/main/java/com/google/android/fhir/datacapture/QuestionnaireEditAdapter.kt
@@ -1,5 +1,5 @@
 /*
- * Copyright 2022-2024 Google LLC
+ * Copyright 2022-2025 Google LLC
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -29,6 +29,7 @@ import com.google.android.fhir.datacapture.extensions.shouldUseDialog
 import com.google.android.fhir.datacapture.views.NavigationViewHolder
 import com.google.android.fhir.datacapture.views.QuestionnaireViewItem
 import com.google.android.fhir.datacapture.views.factories.AttachmentViewHolderFactory
+import com.google.android.fhir.datacapture.views.factories.AutoCompleteViewAnswerOption
 import com.google.android.fhir.datacapture.views.factories.AutoCompleteViewHolderFactory
 import com.google.android.fhir.datacapture.views.factories.BooleanChoiceViewHolderFactory
 import com.google.android.fhir.datacapture.views.factories.CheckBoxGroupViewHolderFactory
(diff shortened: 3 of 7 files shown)


catalog/src/main/java/com/google/android/fhir/catalog/CatalogApplication.kt
```diff
@@ -44,6 +47,15 @@ class CatalogApplication : Application(), DataCaptureConfig.Provider {
...
           ContribQuestionnaireItemViewHolderFactoryMatchersProviderFactory,
+        callbacks = mapOf(Pair(CustomCallbackType.AUTO_COMPLETE, AutoCompleteCallback(
+          callback = { query ->
+            run {
+              listOf(AutoCompleteViewAnswerOption("a", "Type 2 Diabetes Mellitus"),
+                AutoCompleteViewAnswerOption("b", "Test")
+                )
+            }
+          }
+        )))
```
FikriMilano, line 58: other than the query, I think we also need the value set resolver url to be returned in the API. 

the url will be provided by the questionnaire item, what it does that it can point to a terminology server to retrieve the option lists, then the query will filter that lists so we can have less options to process

FikriMilano (changes requested):

catalog/src/main/java/com/google/android/fhir/catalog/CatalogApplication.kt
jingtang10: i think what fikri meant is tha tthe url should be passed as another parameter besides the query string itself. otherwise how do we distinguish different valueset searches?

datacapture/src/main/java/com/google/android/fhir/datacapture/QuestionnaireEditAdapter.kt
```diff
@@ -391,3 +392,12 @@ internal object DiffCallbacks {
...
     }
 }
+
+sealed class CustomCallback {
+  data class AutoCompleteCallback(val callback: (String) -> List<AutoCompleteViewAnswerOption>) :
+    CustomCallback()
+}
+
+enum class CustomCallbackType {
+  AUTO_COMPLETE,
```
jingtang10, line 402: what are the other callbacks types?

i think the definition of the callback should probably be the same for different widgets (e.g. auto complete, or dropdown), but the ui should behave differently.

jingtang10 (changes requested):
Thanks @dilys-l! A couple of questions below.

catalog/src/main/java/com/google/android/fhir/catalog/CatalogApplication.kt
dilys-l: Yeah good point. I will make this change.

datacapture/src/main/java/com/google/android/fhir/datacapture/QuestionnaireEditAdapter.kt
dilys-l: My initial thinking was to make it a bit more open ended, but i agree that standardising the definition would be better. I'll make this change

catalog/src/main/java/com/google/android/fhir/catalog/CatalogApplication.kt
```diff
@@ -44,6 +47,15 @@ class CatalogApplication : Application(), DataCaptureConfig.Provider {
         xFhirQueryResolver = { fhirEngine.search(it).map { it.resource } },
         questionnaireItemViewHolderFactoryMatchersProviderFactory =
           ContribQuestionnaireItemViewHolderFactoryMatchersProviderFactory,
+        callbacks = mapOf(Pair(CustomCallbackType.AUTO_COMPLETE, AutoCompleteCallback(
+          callback = { query ->
```
FikriMilano, line 51: this code is throwing an error when I tried to build the project, could you have a look?
and since this callback is being used to resolve searched options from value sets, should we rename it to something else?

FikriMilano (changes requested):

catalog/src/main/java/com/google/android/fhir/catalog/CatalogApplication.kt
dilys-l: Sorry, i forgot to include the update for this file. Is there a preferred naming convention for something like this?
Perhaps a ValueSetLookupResolver or similar? 
I'll update the code once we agree on a new name.

FikriMilano: that's a good name actually

datacapture/src/main/java/com/google/android/fhir/datacapture/DataCaptureConfig.kt
```diff
@@ -56,6 +56,12 @@ data class DataCaptureConfig(
   var questionnaireItemViewHolderFactoryMatchersProviderFactory:
     QuestionnaireItemViewHolderFactoryMatchersProviderFactory? =
     null,
+
+  /**
+   * A [CustomCallback] may be set by the client to override the behaviour of an existing component
+   * in the sdc.
+   */
+  var callback: CustomCallback<*>? = null,
```
FikriMilano, line 64: since we are going to use this callback only for valueSetLookupResolver, should we change the type to `(String, String) -> List<Coding>`?

datacapture/src/main/java/com/google/android/fhir/datacapture/QuestionnaireEditAdapter.kt
```diff
@@ -391,3 +391,5 @@ internal object DiffCallbacks {
       }
     }
 }
+
+typealias CustomCallback<T> = (String, String) -> List<T>
```
jingtang10, line 395: are we expecting any other types than coding?

datacapture/src/main/java/com/google/android/fhir/datacapture/DataCaptureConfig.kt
jingtang10, line 64: if you just add an option url to ExternalAnswerValueSetResolver below would it achieve the same thing?

datacapture/src/main/java/com/google/android/fhir/datacapture/views/factories/AutoCompleteViewHolderFactory.kt
```diff
@@ -242,8 +272,69 @@ internal object AutoCompleteViewHolderFactory :
...
+    return object : Filter() {
+      override fun performFiltering(constraint: CharSequence?): FilterResults {
+        val query = (constraint?.toString() ?: "").trim()
+        val filteredResults: List<AutoCompleteViewAnswerOption> =
+          if (callback != null && answerValueSet != null && objects.isEmpty()) {
+            (callback as? CustomCallback<AutoCompleteViewAnswerOption>)?.invoke(
+              query,
+              answerValueSet,
+            )
+              ?: emptyList()
```
FikriMilano, line 323: could you move this implementation to the QuestionnaireViewModel? resolving the options in the UI code seems a bit wrong
https://github.com/google/android-fhir/blob/2a7e91f7e118569de896a033b1583cafc6cc1555/datacapture/src/main/java/com/google/android/fhir/datacapture/QuestionnaireViewModel.kt#L926-L930

FikriMilano (changes requested):

datacapture/src/main/java/com/google/android/fhir/datacapture/DataCaptureConfig.kt
dilys-l: the way that answerOptionValues are handled in AutoCompleteViewHolderFactory would need to change. i could probably change it so that it's handled in the adapter either way, but then the resolver would be triggered every time the query changes which might not be ideal

datacapture/src/main/java/com/google/android/fhir/datacapture/QuestionnaireEditAdapter.kt
dilys-l: probably not, i can change this

PR #2766, opened 2025-01-14
Fix validation message for number fields to consider min and max values

harshitIIITD:
Update validation messages for number fields to consider min and max values from the questionnaire item.

* Modify `EditTextDecimalViewHolderFactory.kt` to include min and max values in the validation message for decimal fields.
* Modify `EditTextIntegerViewHolderFactory.kt` to include min and max values in the validation message for integer fields.

### DIFF
--- a/datacapture/src/main/java/com/google/android/fhir/datacapture/views/factories/EditTextDecimalViewHolderFactory.kt
+++ b/datacapture/src/main/java/com/google/android/fhir/datacapture/views/factories/EditTextDecimalViewHolderFactory.kt
@@ -76,8 +76,14 @@ internal object EditTextDecimalViewHolderFactory :
           )
         // Update error message if draft answer present
         if (questionnaireViewItem.draftAnswer != null) {
+          val minValue = questionnaireViewItem.minAnswerValue?.toString() ?: DecimalType(Int.MIN_VALUE).toString()
+          val maxValue = questionnaireViewItem.maxAnswerValue?.toString() ?: DecimalType(Int.MAX_VALUE).toString()
           textInputLayout.error =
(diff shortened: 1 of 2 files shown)


datacapture/src/main/java/com/google/android/fhir/datacapture/views/factories/EditTextDecimalViewHolderFactory.kt
```diff
@@ -76,8 +76,14 @@ internal object EditTextDecimalViewHolderFactory :
           )
         // Update error message if draft answer present
         if (questionnaireViewItem.draftAnswer != null) {
+          val minValue = questionnaireViewItem.minAnswerValue?.toString() ?: DecimalType(Int.MIN_VALUE).toString()
+          val maxValue = questionnaireViewItem.maxAnswerValue?.toString() ?: DecimalType(Int.MAX_VALUE).toString()
```
FikriMilano, line 80: If we check fhir spec about decimal. It was mentioned that the limit is 18 digit + 1 decimal point (https://build.fhir.org/datatypes.html#decimal)

So, I think we shouldn't use the max or min value of integer, as it only accepts 10 digits at most (https://build.fhir.org/datatypes.html#integer)

Also, if we check the parent class of DecimalType, it's actually BigDecimal. In Java BigDecimal doesn't really have a limit, it's constrained only by the amount of available memory.

But it's good to follow the limit set by fhir spec.

FikriMilano (changes requested):

jingtang10 (changes requested):
thanks for the work!

please see the comment i left in the other pr: https://github.com/google/android-fhir/pull/2763#issuecomment-2604297117

PR #2913, opened 2025-11-19
Add missing versions in compose dependencies

jingtang10:
Also remove unused dependencies and reorganize dependencies.


ellykits:
> Maybe we should consider using this approach to handle compose dependencies if it is any better [Compose BOM](https://developer.android.com/develop/ui/compose/bom#kts)

I've confirmed that this is already being used. It is safe to ignore the previous comment.

